PR #9, opened 2015-05-13
Fix overflow problem with converting SchemaRDD with more than 128 cols

nfergu:
This fixes an apparent bug when converting SchemaRDDs with more than 128 columns to DataFrames.

The code in H2OSchemaUtils seems to convert column indexes to Bytes (from Ints), which overflows if there are more than 128 columns. This eventually resulted in an ArrayIndexOutOfBoundsException when getting the column value.

I haven't tested this extensively -- I just changed everything I could see in H2OSchemaUtils that looked like an index from a Byte to Int and verified that it fixed the error, so it certainly needs somebody who knows the code better to check it.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -361,7 +361,7 @@ object H2OContext extends Logging {
   }
 
   private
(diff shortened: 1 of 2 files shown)


mmalohlava:
Hey Neil, thanks for fix! I was too optimistic to use Byte for number of Schema attributes.

Going to merge your changes into master directly.

Thx!

PR #15, opened 2015-06-20
Adapt H2OContext for usage in Spark 1.4.0+

hvanhovell:
A small PR to enable the use of Sparkling Water with the Spark 1.4.0. The PR will work with Spark 1.3 though.

The key 'problem' is that a separation has been made between internal and external data types as of Spark 1.4. Using the current version of Sparkling Water will lead to a number of ClassCastExceptions when using Strings (it will try to cast them to a UTF8String which is the internal String implementation).


mmalohlava:
Hi Herman, sorry for delay in merging - i hope i will merge it during next week.
It looks good for me, but need to perform tests on the top of our infrastructure.
Thanks for contribution! It is great!

PR #17, opened 2015-07-14
Frame metadata

petro-rudenko:
Computing summary statistics in spark is a bit painful. There are methods dataframe.describe and dataframe.stat, that do some summary statistics only for numeric column. To calculate cardinality, missing value count, etc. need to implement something [like this](http://blog.cloudera.com/blog/2015/07/how-to-do-data-quality-checks-using-apache-spark-dataframes/). If we transfering h2o frame to spark dataframe we have this statistics for free and we can put into spark's metadata, to reuse later in ml pipeline (e.g. to do feature scaling to not recompute mean and std, based on cardinality estimation tract features as numeric or categorical, etc.).

TODO:
- [ ] If the column is string and categorical (e.g. only 2 categories like "YES", "NO"), can't see how to treat it as enum without doing ordinal encoding on spark/h2o side.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OSchemaUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OSchemaUtils.scala
@@ -45,7 +45,7 @@ object H2OSchemaUtils {
         putLong("naCnt", vec.naCnt())
 
       if (vec.isEnum) {
-        metadata = metadata.putStringArray("domain", vec.domain()).
+        metadata = metadata.putStringArray("vals", vec.domain()).
           putLong("cardinality", vec.cardinality().toLong)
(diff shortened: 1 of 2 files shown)


mmalohlava:
Hi @petro-rudenko ! That's cool feature!

I was thinking to use spark metadata to easy transfers between Spark and H2O (for example, if we need to decide if a column should be marked as categorical or as string-column).

Regarding your TODO:
  - we treat all small domain columns as categorical. On the way from H2O to Spark I decided to transfer categorical always to String column. But i think in your case, you can save cardinality as metadata.

mmalohlava:
Hi @petro-rudenko , i rebase&squashed your PR. It is in the master.

Issue #19, opened 2015-10-20
NullPointerException in Docker run-example.sh


tcfuji:
Getting the following exception after running `bin/run-example.sh` within the bash container:
```
15/10/20 16:29:02 INFO TaskSetManager: Finished task 196.0 in stage 17.0 (TID 1040) in 565 ms on 0b4ad106ca69 (200/200)
15/10/20 16:29:02 INFO TaskSchedulerImpl: Removed TaskSet 17.0, whose tasks have all completed, from pool
15/10/20 16:29:02 INFO DAGScheduler: Stage 17 (runJob at H2OContext.scala:258) finished in 20.698 s
15/10/20 16:29:02 INFO DAGScheduler: Job 8 finished: runJob at H2OContext.scala:258, took 21.150594 s
Exception in thread "main" java.lang.NullPointerException
	at org.apache.spark.examples.h2o.AirlinesWithWeatherDemo2$.main(AirlinesWithWeatherDemo2.scala:83)
	at org.apache.spark.examples.h2o.AirlinesWithWeatherDemo2.main(AirlinesWithWeatherDemo2.scala)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.apache.spark.deploy.SparkSubmit$.launch(SparkSubmit.scala:358)
	at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:75)
	at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)

^C15/10/20 16:32:24 INFO ExecutorRunner: Killing process!
15/10/20 16:32:24 INFO ExecutorRunner: Killing process!
15/10/20 16:32:24 INFO ExecutorRunner: Killing process!
15/10/20 16:32:24 ERROR FileAppender: Error writing stream to file /opt/spark/work/app-20151020162558-0000/1/stderr
java.io.IOException: Stream closed
	at java.io.BufferedInputStream.getBufIfOpen(BufferedInputStream.java:162)
	at java.io.BufferedInputStream.read1(BufferedInputStream.java:272)
	at java.io.BufferedInputStream.read(BufferedInputStream.java:334)
[...]

PR #20, opened 2015-10-23
Update Dockerfile to Spark 1.5.1 and latest sparkling-water version

tcfuji:
#19


tcfuji:
Added instructions to line 6 to prevent build errors.

MadMan0708:
Thank you for this PR! We have actually created gradle task which creates docker file and determines latest version of sparkling-water automatically. It's gonna be in master and hopefully in a new sparkling-water release.

Kuba

PR #34, opened 2016-05-25
Fix a typo

Atry:


MadMan0708:
Hi @Atry , good catch! Thanks for reporting it! Since this is quite a small change, I incorporated it into a bigger change which will be in the next release.

PR #38, opened 2016-08-10
[SWPRIVATE-12] Config for SSL and utility class for key creation

mdymczyk:
---

<a name="Security"></a>

Both Spark and H2O support basic node authentication and data
encryption.

To enable it in Spark please check [their
documentation](http://spark.apache.org/docs/latest/security.html).

Security for H2O nodes can be enabled manually by generating all
necessary files, distributing them to all worker nodes as described in
[H2O's
documentation](https://github.com/h2oai/h2o-3/blob/master/h2o-docs/src/product/security.rst)
and passing the "spark.ext.h2o.internal_security_conf" to spark submit:

::
```
bin/sparkling-shell /
--conf "spark.ext.h2o.internal_security_conf=ssl.properties"
```

We also provide a utility method which will automatically generate all
files and enable security on all H2O nodes:

::

```
import org.apache.spark.network.Security
import org.apache.spark.h2o._
Security.enableSSL(sc) // generate properties file, keypairs and set appropriate H2O parameters
val hc = H2OContext.getOrCreate(sc) // start the H2O cloud
```

This method will generate all files and distributed them via YARN or
Spark methods to all worker nodes. This communication will be secure if
you configured YARN/Spark security.

---

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
@@ -81,8 +81,9 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
   // Options in form key=value
     Seq(
       ("-name", conf.cloudName.get),
-      ("-nthreads", if (conf.nthreads > 0) conf.nthreads else null))
-      .filter(x => x._2 != null)
+      ("-nthreads", if (conf.nthreads > 0) conf.nthreads else null),
+      ("-internal_security_conf", conf.sslConf.orNull)
+    ).filter(x => x._2 != null)
       .flatMap(x => Seq(x._1, x._2.toString)) ++ // Append single boolean options
       Seq(("-ga_opt_out", conf.disableGA))
         .filter(_._2).map(x => x._1)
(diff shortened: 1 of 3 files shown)


mdymczyk:
Needs a change of H2O version after SSL gets published.

mmalohlava:
Is it still in WIP?

mdymczyk:
@mmalohlava no, just needed a rebase from master, also I had to bump the H2O version used because the one used in master (7) is one too early (no SSL).

Up for review.

gradle.properties
```diff
@@ -6,7 +6,7 @@ h2oMajorVersion=3.10.0
 # Name of H2O major version
 h2oMajorName=turing
 # H2O Build version, defined here to be overriden by -P option
-h2oBuild=7
+h2oBuild=9
```
mmalohlava: Lets bump it first in master, then we can remove it here.

jakubhava: There are also some small additional changes because of the bump. It's good to have them separated I think

mmalohlava:
I just merged upgrade to the latest version of H2O into master. Can we update this branch?

mdymczyk:
@mmalohlava @jakubhava done

PR #39, opened 2016-08-11
[SW-170] Unification of the API for H2OFrame data source and asDataFrame transformation

mmalohlava:
The fix introduces `H2OFrameRelation` implementing `BaseRelation` and additional releation
to allow column filtering. The H2OFrameRelation is used instead of H2ORelation which was removed.


mmalohlava:
Problem with Travis - because of using a new sudo enabled Ubuntu images. See https://www.traviscistatus.com/incidents/11hp8bhkrkn7

mmalohlava:
Going to merge into master

PR #41, opened 2016-08-11
[SW-168] a flag to control transfer of metadata from H2OFrame to DataFrame

mmalohlava:
@mklechan observed slow down of `select count(*) from h2oFrame` query after `asDataFrame` call during handling larger datasets.
The slow down was caused by Spark querying metadata which we attached during transformation
of H2OFrame to Dataset.

The fix introduces an additional option for asDataFrame call to control transfer
of metadata from H2OFrame to DataFrame.


mmalohlava:
Again Travis CI build failure  - still Travis using wrong build image - see https://www.traviscistatus.com/incidents/11hp8bhkrkn7

PR #44, opened 2016-08-12
[SW-177] REPL simplification

MadMan0708:
This PR cleans REPL a little bit.

The initialization piece of code does not have to have session id as part of it's name. It is just used to test wheter the compiler has been successfully initialised. We don't therefore have to change the initialization code and thus can remove it. ( the interpreter test runs anyway )

Also the code in `setupClassNames` was simplified. This change also allows us to use the same code in this method among all spark versions for scala 2.10 and 2.11. 

Also remove unnecessary setting of classpath since it's always overriden in `createInterpreter` method.

### DIFF
--- a/repl/src/main/scala/org/apache/spark/repl/h2o/H2OIMain.scala
+++ b/repl/src/main/scala/org/apache/spark/repl/h2o/H2OIMain.scala
@@ -17,114 +17,71 @@
 
 package org.apache.spark.repl.h2o
 
-import org.apache.spark.util.MutableURLClassLoader
-import org.apache.spark.{SparkContext, SparkEnv, HttpServer}
-import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.repl.{Main, SparkIMain}
+import org.apache.spark.util.MutableURLClassLoader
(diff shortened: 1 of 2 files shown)


mmalohlava:
We need JIRA ticket for this guy.

mmalohlava:
My bad, there is jira ticket, sorry for bothering!

MadMan0708:
Yup, I have created JIRA and did push --force with updated comment. I forgot for the first time.

MadMan0708:
Has somebody any feedback or can we merge it ? ( The commit with REPL changes for  Spark 2.11 in `JH_SW_2.0` already contains these changes + changes necessary for Spark 2.0 )

mdymczyk:
LGTM (code wise, hard for me to comment on the actual REPL logic)

MadMan0708:
@mdymczyk Thanks for review! 

We are doing there a little bit of reflection magic in order to avoid copying of whole `SparkIMain` and `SparkILoop` classes.

MadMan0708:
OK, if nobody is against, I will merge this PR later today.

PR #48, opened 2016-08-13
[SW-158] Added Dataset to H2O spark interface.

vpatryshev:
Products (aka case classes) can contain multiple primitive or optional components. Optional are of type Option[T], where T is primitive.

Serialization writes options as undefined.
Deserialization transforms undefined to None and legit values x to Some(x) - when the field is optional. When the field is primitive, undefined values mean the whole row is undefined, and it is being skipped. Any other transformation errors also lead to skipping offending rows.


mdymczyk:
@vpatryshev could you rebase the current master onto this PR? Will make reviewing much easier (now it's showing a bunch of classes/files as added which are in master like `PipelineDemo`).

vpatryshev:
Just did.

Thanks,
-Vlad

mdymczyk:
Hm github is still complaining `This branch has conflicts that must be resolved
 Use the command line`.

MadMan0708:
Yup, it would be great if we could see only changes related to this change.

vpatryshev:
I can try to _merge_ master into this branch. This kind of guarantees, I guess, unlike rebase. Do you mind if I merge?

mdymczyk:
Hm after having a second look not sure a merge will help since its all now squashed into a single commit, not really sure what happened here (looks like some other commits and the DataSet commits got squashed together?).

But a merge before we fast forward it back into master is necessary anyway so you can go ahead I guess if @mmalohlava and @MadMan0708 have no objections.

We should come up with a dev strategy now that we're using PRs and the team is growin (rebase vs merge etc).

h2o-ops:
On 8/17/16 12:27 PM, Mateusz Dymczyk wrote:
>
> Hm after having a second look not sure a merge will help since its all now squashed into a single 
> commit, not really sure what happened here (looks like some other commits and the DataSet commits 
> got squashed together?).
>
> But a merge before we fast forward it back into master is necessary anyway so you can go ahead I 
> guess if @mmalohlava <https://github.com/mmalohlava> and @MadMan0708 
> <https://github.com/MadMan0708> have no objections.
>
Please, hold the merge - test.h2o.ai reports red and also CDH has some problems with Sparkling Water.
>
> We should come up with a dev strategy now that we're using PRs and the team is growin (rebase vs 
> merge etc).
>
In SW i would say rebase since we need to integrate changes from master into release branches and it 
is much more easier if git history is without merges.

michal

vpatryshev:
Okay, so my rebase attempts don't lead to success; an ideas?
Merge manually, as the script suggests?

mdymczyk:
@vpatryshev ok I see you were able to unsquash those commits, good! I took this branch, cherry picked your commits and squashed them into on, pushed it here https://github.com/h2oai/sparkling-water/commits/MD_vlad_dataset

Please check if nothing's missing and rebase the current master into that branch. You will get quite a bit of conflicts as @mmalohlava and @MadMan0708 pushed some refactoring commits.

After you're done you can push that branch into this PR.

Think this should work.

I know merge is a better option when the branch is public but as @mmalohlava mentioned we need a cleaner git history to handle merges to release branches.

vpatryshev:
Yesterday I rebased three times. Same thing, have a conflict, two
identically-looking chunks can't be merged automatically; I fix it
manually, add, continue rebase, push - does not work.

What's your suggestion? Do it four more times? I can write a script for
that.

Thanks,
-Vlad

mdymczyk:
@vpatryshev I meant please try rebasing master onto this branch https://github.com/h2oai/sparkling-water/commits/MD_vlad_dataset (i cleaned up the history and squashed your commits into one). You should get ~5 conflicts.

vpatryshev:
Ok, working on it now.
Found something to fix, meanwhile.

Thanks,
-Vlad

PR #49, opened 2016-08-15
[SW-178] Implement external backend

jakubhava:
This architectural change allows to connect to existing h2o cluster from sparkling water. This has a benefit that we are no longer affected by Spark killing it's executors thus we should have more stable solution in environment with lots of h2o/spark node.

This change depends on JH_SW_connection branch in h2o project. Before integrating this change, the change on h2o side has to be integrated first.


jakubhava:
This PR has been separated to 2 smaller PRs. Please see branches `JH_SW-178_external_cluster` and `JH_SW-178_pre_external_cluster`

PR #51, opened 2016-08-17
[SWPRIVATE-14] Spark's Gaussian Mixture model

mdymczyk:
Wrapping [Spark's Gaussian Mixture model clustering](http://spark.apache.org/docs/latest/mllib-clustering.html#gaussian-mixture) as H2O model so it can be used from Flow/on H2OFrames.


mdymczyk:
TODO:

- column standardization (after NA value handler is merged to master)
- NA value handler (waiting for a merge)
- progress bar (waiting for a merge)
- POJO generation
- unit tests I guess ;-)

mdymczyk:
POJO generation for now not supported as the scoring code used by Spark (which I want to mimic) is very complicated and uses a lot of algebraic libraries.

My initial idea is to make this POJO, contrary to our other POJOs, depend on some of those libs, how does that sound?

tomkraljevic:
The H2O core genmodel package itself should definitely not have any dependencies on Spark (or anything).

If you want to make a new package, in Sparkling Water or something that isn't core H2O, (say "SparkGenModel") that does depend on Spark, that seems conceptually fine to me.

mdymczyk:
@tomkraljevic I was thinking about just using some `import random.algebraic.lib` in my POJO and telling the users "ok for this particular POJO you need this lib on your classpath". Definitely don't want to add any deps to the genmodel jar.

jakubhava:
I'll do review as soon as I can, but just want to point this out when I have it in my mind:

We need to add Gaussian Mixture model and related classes to uber jar generated by `extendJar` task in `assembly-h2o` subproject so the users can use it also on the external backend.

mdymczyk:
@jakubhava ah ok I see, I should be able to just add the whole `ml` submodule as dependency, right?

jakubhava:
@mdymczyk yes, that's perfectly alright. I haven't done that before because I wasn't using zip64 and wanted to minimise the jar size. We need to run tests on this also in external cluster to be sure it doesn't need any extra classes from spark side to be available in the uber jar and add them as well if needed.

ml/src/main/java/org/apache/spark/ml/spark/models/gm/GaussianMixtureScorer.java
```diff
@@ -0,0 +1,97 @@
...
+
+
+    public static void generate(SBPrintStream bodySb,
+                                CodeGeneratorPipeline classCtx,
+                                CodeGeneratorPipeline fileCtx,
+                                double[] weights,
+                                MultivariateGaussian[] gaussians,
+                                double epsilon,
+                                boolean meanImputation) {
+        // TODO implement
```
jakubhava: Should this be implemented as part of this PR ?

PR #52, opened 2016-08-21
[SW-176] Adds a missing method from the CDH SparkListener trait.

mdymczyk:
**Why the fix:**

CDH added a new method to `SparkListener` (which was added in Spark2.0) https://github.com/cloudera/spark/blob/cdh5.7.1-release/core/src/main/scala/org/apache/spark/scheduler/SparkListener.scala#L245 We're getting an `AbstractMethodError` when running on CDH5.7.1 since the trait against which we're compiling (Spark 1.6) doesn't have `onExecutorAdded`.

Adding the `onExecutorAdded` has to be done in a separate trait/class, adding it on the anonymous class won't work due to how Scala compiles it. 

**How this was tested:**

Manually ran `./sparkling-shell --num-executors 2 --executor-memory 6g --driver-memory 6g --master yarn-client --conf spark.dynamicAllocation.enabled=false`:

```
import org.apache.spark.h2o._
val hc = H2OContext.getOrCreate(sc)
sc.parallelize(Array(1,1)).toDF.count()
```

on `mr-0xd6-precise1` with `SPARK_HOME=/opt/cloudera/parcels/CDH-5.7.1-1.cdh5.7.1.p0.11/lib/spark/` (previously, with SW1.6.5 would fail with the aforementioned error).

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
@@ -21,7 +21,7 @@ import org.apache.spark.Logging
 import org.apache.spark.h2o.backends.SparklingBackend
 import org.apache.spark.h2o.utils.NodeDesc
 import org.apache.spark.h2o.{H2OConf, H2OContext}
-import org.apache.spark.scheduler.{SparkListener, SparkListenerExecutorAdded}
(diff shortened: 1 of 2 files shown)


MadMan0708:
Looks good to me.

We had similar problems in REPL since CDH already put some commits aimed for spark 2.0 into their release for 1.6

PR #53, opened 2016-08-22
[SW-185] Post-refactorization and post-integration fixes

MadMan0708:
This PR fixed problem in H2ODataFrame where we are not allowed to use methods requiring H2O backend since in future external cluster mode, h2o is not running in a spark executor.

Tests are passing.

@mmalohlava Pls let me know if it looks good to you.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
@@ -18,7 +18,7 @@
 package org.apache.spark.h2o.converters
 
 
-import org.apache.spark.h2o.{H2OConf, H2OContext}
+import org.apache.spark.h2o.H2OConf
 import org.apache.spark.h2o.utils.ReflectionUtils
 import org.apache.spark.rdd.RDD
 import org.apache.spark.{Partition, SparkContext, TaskContext}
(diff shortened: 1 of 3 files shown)


MadMan0708:
There is actually more bugs, please wait before merging

MadMan0708:
More fixes, tested - both integration tests and tests are passing

- fr.vecs can't be called in compute method on external cluster
- method prepareExpectedTypes has to be called on driver node
- convertorContext has to be implemented right in the iterator, not in the trait - otherwise it uses default value set in trait which causes that internal convertor context is used all the time
- other small changes

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -76,14 +80,22 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
                   s"Constructor must take exactly ${colNames.length} args")
       })
 
-      override val expectedTypes: Option[Array[Byte]] = ConverterUtils.prepareExpectedTypes(isExternalBackend, types)
       override val keyName = frameKeyName
       override val partIndex = split.index
+      val expectedTypes: Option[Array[Byte]] = expectedTypesAll
```
mmalohlava: You can move it into trait

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -76,14 +80,22 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+
+      /* Converter context */
+      override val converterCtx: ReadConverterContext =
+      ConverterUtils.getReadConverterContext(isExternalBackend,
```
mmalohlava: You can move the creation into trait

PR #54, opened 2016-08-24
[SW-165] Fixes CSS for Scala cell to black

mdymczyk:


MadMan0708:
Cool! 

The best would be to put this change straight into h2o-flow - https://github.com/h2oai/h2o-flow/blob/master/src/scala-editor.styl

mmalohlava:
It is Sparkling Water specific style - flow dev iteration is too long since it depends on also releasing H2O (tiny change in style would trigger release of Flow followed by release of H2O - in our 3 week release cycle it is simply slow). 

Better to have in our project where we can iterate much more easily.

But at the end it should go eventually to flow

MadMan0708:
ok, make sense, never mind than

mmalohlava:
Lets look at style, improve it little bit, then propagate to Flow.

PR #56, opened 2016-08-24
[SW-71] Added setLogLevel on H2O internal config

mdymczyk:
The [Jira](https://0xdata.atlassian.net/browse/SW-71) mentioned adding them on `H2OContext` but setting those values on those instances will have no effect after `init()` so I added them to the config classes.

Thoughts?

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
@@ -53,6 +53,12 @@ trait InternalBackendConf extends SharedH2OConf {
         |  h2oClientLog         : ${h2oClientLogLevel}
         |  nthreads             : ${nthreads}
         |  drddMulFactor        : $drddMulFactor""".stripMargin
+
+
+  def setH2ONodeLogLevel(level: String): H2OConf = {
+    sparkConf.set(PROP_NODE_LOG_LEVEL._1, level)
+    self
+  }
 }
 
 object InternalBackendConf {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
```diff
@@ -53,6 +53,12 @@ trait InternalBackendConf extends SharedH2OConf {
         |  h2oClientLog         : ${h2oClientLogLevel}
         |  nthreads             : ${nthreads}
         |  drddMulFactor        : $drddMulFactor""".stripMargin
+
+
+  def setH2OLogLevel(level: String): H2OConf = {
```
MadMan0708: `setH2ONodeLogLevel` would be better here probably, just to make it more clear and consistent with the other method

MadMan0708:
It's great by me 👍

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
mdymczyk: Agreed, done.

mmalohlava:
Is it possible no modify log level in the client node at runtime? That would be really useful! :)

For H2O nodes (=executors), you can write a simple H2O `MRtask` which will be invoked on the all nodes and reconfigure the log subsystem. If you need to modify H2O, please, let me know.

mmalohlava:
You can also try to rebase on top of master, since i fixed the travis build.

mdymczyk:
@mmalohlava setting the level at runtime now, changing both the water.Log._level and the log4j `water.default` level should someone be using a custom config file and not our default `TRACE` level.

PR #57, opened 2016-08-24
[SW-128] Move flows into core so they are shown in flowui

mdymczyk:
Is this enough? The [Jira](https://0xdata.atlassian.net/browse/SW-128) mentioned something about integrating into github but they were already in github? 

The flows are now packaged with the core jar and show up in the FlowUI `Help > examples` together with core H2O examples when running SW.

### DIFF
--- a/examples/build.gradle
+++ b/examples/build.gradle
@@ -3,6 +3,24 @@ description = "Sparkling Water Examples"
 apply from: "$rootDir/gradle/utils.gradle"
 apply from: "$rootDir/gradle/scriptsTest.gradle"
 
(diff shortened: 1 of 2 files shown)


mmalohlava:
Is it possible to leave the flows in examples project. And just override resource directly in examples project?

mdymczyk:
How about this @mmalohlava ? This will work from a `jar` after assembling but won't work when run from intellij as I'm rewriting the path in `jar` task because FlowUI requires a `www/flow/packs/examples` path (alternative would be to rename the folder in examples to that path but we probably don't want that).

mmalohlava:
Looks good!

PR #58, opened 2016-08-24
Vp dataset take3

vpatryshev:
At last, the branch with all Dataset-enabling changes for sparkling water.


core/build.gradle
```diff
@@ -76,6 +76,7 @@ test {
 
     // Working dir will be root project
     workingDir = rootDir
+    testLogging.showStandardStreams = true
```
jakubhava: Do we need to show console output during running tests ?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -49,18 +50,21 @@ import scala.util.control.NoStackTrace
...
+  */
 
 /**
   * Create new H2OContext based on provided H2O configuration
   *
   * @param sparkContext Spark Context
   * @param conf H2O configuration
   */
-class H2OContext private (@transient val sparkContext: SparkContext, @transient conf: H2OConf) extends org.apache.spark.Logging
+class H2OContext private (@transient val sparkContext: SparkContext, @transient conf: H2OConf, sqlc: SQLContext) extends org.apache.spark.Logging
```
jakubhava: I was thinking whether we need another parameter for sql context and change the api. We could just do `SQLContext.getOrCreate(sc)` inside of `H2OContext`. This returns sql context already created by spark or new one if none exists.

What is please the intention for this ?

Thanks for explanation 👍

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
     * @return H2O Context
     */
   def getOrCreate(sc: SparkContext): H2OContext = {
-    getOrCreate(sc, new H2OConf(sc))
+    getOrCreate(sc, new H2OConf(sc))(SQLContext.getOrCreate(sc))
+  }
+
+  val defaultFieldNames = (i: Int) => "f" + i
```
jakubhava: The code from here down should be located in `org.apache.spark.converts` package in order to keep h2o context clean. There are classes which takes care of converting various types such as RDD containing primitive type, RDD containing product class, dataframe to h2o frames and back.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
...
+
+      // Add Vec headers per-Chunk, and finalize the H2O Frame
+      buildH2OFrame(kn, meta.vecTypes, res)
+    }
+
+    private
+    def perTypedDataPartition[A<:Product](keystr:String, vecTypes: Array[Byte])
+                                         ( context: TaskContext, it: Iterator[A] ): (Int,Long) = {
+      // An array of H2O NewChunks; A place to record all the data in this partition
+      val nchks = water.fvec.FrameUtils.createNewChunks(keystr, vecTypes, context.partitionId)
```
jakubhava: This should use `InternalReadConvertorContext`

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
...
+      it.foreach(prod => { // For all rows which are subtype of Product
+        for( i <- 0 until prod.productArity ) { // For all fields...
+        val fld = prod.productElement(i)
+          val chk = nchks(i)
+          val x = fld match {
+            case Some(n) => n
+            case _ => fld
+          }
+          x match {
+            case n: Number  => chk.addNum(n.doubleValue())
```
jakubhava: We should use chunks right away but again use `InternalWriteConvertorContext`. Please see examples in `convertors` package how it's done for example for rdd of primitive type. 

This is all needed because new upcoming feature for sparkling water adds another backend - h2o won't be running in each executor, but as a separate cluster and these convertors are here to unify the api.

core/src/main/scala/org/apache/spark/h2o/SupportedDataset.scala
```diff
@@ -0,0 +1,72 @@
...
+        prototype.withFields(params)
+      } catch {
+        case oops: Exception =>
+          oops.printStackTrace()
+          throw oops
+      }
+      res
+    }
+  }
+}
```
jakubhava: Should be probably again in `convertors` package

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
```diff
@@ -28,7 +28,8 @@ import scala.collection.mutable.ListBuffer
 import scala.reflect.runtime.universe._
 
 
-private[converters] trait ConverterUtils {
+//private[converters] - will uncomment when H2OContext moves into converters
```
jakubhava: Why should be H2OContext in converters package ? 

Convertors package contains all supported types from which we can transfer to `H2OFrame` and back. `H2OContext` is like orchestrator of communication inside sparkling-water, from my point of view it does not belong to `convertors` and `ConvertorUtils` should remain private to `converters`

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
```diff
@@ -110,6 +119,13 @@ object ConverterUtils {
     converterContext
   }
 
+  def getReadConverterContext(keyName: String,
```
jakubhava: Can you keep only original `getReadConverterContext` function ?
This function will probably get removed anyway after merging of https://github.com/h2oai/sparkling-water/pull/49. Can you please keep the original version and variables like `isExternal` ? Or I can put it back when doing my merge, but seems like extra work :)

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
```diff
@@ -23,7 +23,8 @@ import water.fvec.{Frame, FrameUtils}
 /**
  * Contains functions that are shared between all H2ORDD types (i.e., Scala, Java)
  */
-private[converters] trait H2ORDDLike[T <: Frame] {
```
jakubhava: Why removing the restriction to `convertors` ?

core/src/main/scala/org/apache/spark/h2o/utils/H2OTypeUtils.scala
```diff
@@ -23,20 +23,25 @@ import water.fvec.Vec
...
+  classOf[java.sql.Timestamp] -> Vec.T_TIME
+  ).withDefault(q => throw new IllegalArgumentException(s"Do not understand type $q"))
+
+  def dataTypeToVecType(t: Class[_]): Byte = ClassMap(t)
+
+  import scala.reflect.runtime.universe._
+  import ReflectionUtils._
+
+  def dataTypeToVecType(t: Type): Byte = {
+    ClassMap(typ(t))
```
jakubhava: I like this solution!. It's better to create the map once and then just query it then doing the match all over again.

Thanks !

core/src/main/scala/water/api/RDDs/RDDsHandler.scala
```diff
@@ -64,6 +64,8 @@ class RDDsHandler(val sc: SparkContext, val h2oContext: H2OContext) extends Hand
         case t if t.isInstanceOf[Long] => h2oContext.asH2OFrame(rdd.asInstanceOf[RDD[Long]], name)
         case t if t.isInstanceOf[java.sql.Timestamp] => h2oContext.asH2OFrame(rdd.asInstanceOf[RDD[java.sql.Timestamp]], name)
         case t if t.isInstanceOf[Product] => h2oContext.asH2OFrame(rdd.asInstanceOf[RDD[Product]], name)
+// !!! TODO(vlad): enable it
+//        case t if t.isInstanceOf[Product] => h2oContext.FromPureProduct(sc, rdd.asInstanceOf[RDD[Product]], name).withDefaultFieldNames()
```
jakubhava: Why is this not enabled at this moment ?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+    "Double"  -> ((source: ReadConverterContext) => (col: Int) => source.getDouble(col)),
+    "Float"   -> ((source: ReadConverterContext) => (col: Int) => source.getFloat(col)),
+    "Integer" -> ((source: ReadConverterContext) => (col: Int) => source.getInt(col)),
+    "Long"    -> ((source: ReadConverterContext) => (col: Int) => source.getLong(col)),
+    "Short"   -> ((source: ReadConverterContext) => (col: Int) => source.getShort(col)),
+    "String"  -> ((source: ReadConverterContext) => (col: Int) => source.getString(col)))
+
+  private def returnOption[X](op: ReadConverterContext => Int => X) = (source: ReadConverterContext) => (col: Int) => opt(op(source)(col))
+
+  type TypeName = String
```
jakubhava: Why introduce new TypeName for simple type like `String` ?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
-      ConverterUtils.getReadConverterContext(isExternalBackend,
-        keyName,
-        chksLocation,
-        expectedTypes,
-        partIndex)
+  private val columnTypeNames = ReflectionUtils.typeNames[A](colNamesInResult)
+
+  private val jc = implicitly[ClassManifest[A]].runtimeClass
+
+  private type Deserializer = ReadConverterContext => Int => Any
```
jakubhava: This is just really small thing, but I was thinking whether `Deserializer` is proper name. What we do with `ReadConverterContext` is that we either read data from spark or from external h2o cluster.

Maybe just `Reader` would be better? but not sure

Also can could you do here private `type Deserializer[X] = ReadConverterContext => Int => X` and later instead of  `returnOption[X](op: ReadConverterContext => Int => X)` do `returnOption[X](ip: Deserializer[X]`

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+
+  val extractors = extractorsMap compose columnTypeNames
+
+  private def opt[X](op: => Any): Option[X] = try {
+    Option(op.asInstanceOf[X])
+  } catch {
+    case ex: Exception => None
+  }
+  // maps data columns to product components
+  val columnMapping: Map[Int, Int] =
```
jakubhava: Do we need this `columnMapping` function ? It seems like this case is covered in the `multicolumnMapping` already. In that case I would also rename `multicolumnMapping`  to just `columnMapping`

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+
+
+  lazy val constructors: Seq[Constructor[_]] = {
+
+    val cs = jc.getConstructors
+    val found = cs.collect {
+      case c if c.getParameterTypes.length == colNamesInResult.length => c
     }
 
-    ConverterUtils.getIterator[A](isExternalBackend, iterator)
```
jakubhava: We will need this guy later when merging SW-178, but I can probably add it there :)

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+        throw new scala.IllegalArgumentException(s"Missing columns: ${bads mkString ","}")
+      }
+
+      mappings.toMap
+    }
+  }
+
+  class H2ORDDIterator(val keyName: String, val partIndex: Int) extends H2OChunkIterator[A] {
+
+    private def cell(i: Int) = {
```
jakubhava: Maybe instead of `cell` `convertPerColumn` ? But it's just idea

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+            data <- dataOpt
+            instance <- builder(data)
+            } yield instance
+
+            res.toList match {
+              case Nil => None
+              case unique :: Nil => Option(unique)
+              case one :: two :: more =>
+                throw new scala.IllegalArgumentException(
+                                         s"found more than une $jc constructor for given args - can't choose")
```
jakubhava: typo - more than `uno`

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+    "Boolean" -> ((source: ReadConverterContext) => (col: Int) => source.getBoolean(col)),
+    "Byte"    -> ((source: ReadConverterContext) => (col: Int) => source.getByte(col)),
+    "Double"  -> ((source: ReadConverterContext) => (col: Int) => source.getDouble(col)),
+    "Float"   -> ((source: ReadConverterContext) => (col: Int) => source.getFloat(col)),
+    "Integer" -> ((source: ReadConverterContext) => (col: Int) => source.getInt(col)),
+    "Long"    -> ((source: ReadConverterContext) => (col: Int) => source.getLong(col)),
+    "Short"   -> ((source: ReadConverterContext) => (col: Int) => source.getShort(col)),
+    "String"  -> ((source: ReadConverterContext) => (col: Int) => source.getString(col)))
+
+  private def returnOption[X](op: ReadConverterContext => Int => X) = (source: ReadConverterContext) => (col: Int) => opt(op(source)(col))
```
jakubhava: I would change `op` to `deserializer` in order to be consistent

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+            val dataOpt = extractRow
+
+            val res: Seq[A] = for {
+            builder <- builders
+            data <- dataOpt
+            instance <- builder(data)
+            } yield instance
+
+            res.toList match {
+              case Nil => None
```
jakubhava: Shouldn't we throw here exception saying that no suitable constructor was found ?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+      while (hd.isEmpty && super.hasNext) {
+        hd = readOne()
+        total += 1
+      }
+      hd.isDefined
+    }
+
+    def next(): A = {
+      if (hasNext) {
+        val a = hd.get
```
jakubhava: I'm thinking whether these checks are necessary. The iterator is serving us data and basically when  `hd.isDefined` return false, it does not mean that there are no more data in the iterator, but that there was something wrong with this particular row ( If I understand the new logic of H2ORDD right ) => so maybe it would make more sense to throw an exception and write `hasNext` and `next` without the magic of `hd`.

But maybe I'm missing something. What do you think @mmalohlava @mdymczyk ?

I don't like

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+    }
+
+    private var hd: Option[A] = None
+    private var total = 0
 
-      def next(): A = {
-        val data = new Array[Option[Any]](numCols)
-        // FIXME: this is not perfect since ncols does not need to match number of names
-        (0 until numCols).foreach{ idx =>
-          val value = if (converterCtx.isNA(idx)) None
```
jakubhava: I haven't found in the new code check whether the data at particular position is NA using `converterCtx.isNA(idx)`

Before we get the data, we need to always check whether the data is NA or not. This is also requirement for normal sparkling water ( if I'm not wrong ) and also for new backend solution where the communication protocol between spark executor and H2O always sends first information whether the data is NA or not and if not, then the actual data. If we don't check the data to be NA we would break this communication protocol.

jakubhava:
Thanks for writing all this :) !

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
mdymczyk: Can't we leave this one as it is and add a new one that would resemble the old one `this(sparkContext: SparkContext, conf: H2OConf)` that would call `this(sparkContext, conf, SQLContext.getOrCreate(sc))`? But only if we really need the old constructor.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -224,10 +233,10 @@ object H2OContext extends Logging{
   @transient private val instantiatedContext = new AtomicReference[H2OContext]()
 
   /**
-    * Tries to get existing H2O Context. If it has been created, returns Option containing this H2O Context, otherwise
-    * returns None
+    * Tries to get existing H2O Context. If it has been created, returns this H2O Context, otherwise
+    * returns creates a new one
```
mdymczyk: This doesn't seem right, it does return `None` if context wasn't instantiated.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Yes, if @vpatryshev needs to pass sql context as an argument, I like your proposal for backwards compatibility.

Only reason against using `SQLContext.getOrCreate` I can come up with is that someone would like to use separate sqlContext for sparkling-water within one spark run.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -224,10 +233,10 @@ object H2OContext extends Logging{
...
     *
-    * @return Option containing H2O Context or None
+    * @return H2OContext
```
mdymczyk: The old comment seems correct, why was it changed?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -238,9 +247,10 @@ object H2OContext extends Logging{
     * @param conf H2O configuration
     * @return H2O Context
     */
-  def getOrCreate(sc: SparkContext, conf: H2OConf): H2OContext = synchronized {
+  def getOrCreate(sc: SparkContext, conf: H2OConf)(implicit sqlContext: SQLContext): H2OContext = synchronized {
+    // TODO(vlad): figure out how thread-safe are these operations IRL
```
mdymczyk: The whole method is `synchronized` so it should be ok.

srisatish:
thanks for all the commentary! Keep it up!

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -238,9 +247,10 @@ object H2OContext extends Logging{
     * @param conf H2O configuration
     * @return H2O Context
     */
-  def getOrCreate(sc: SparkContext, conf: H2OConf): H2OContext = synchronized {
+  def getOrCreate(sc: SparkContext, conf: H2OConf)(implicit sqlContext: SQLContext): H2OContext = synchronized {
```
mdymczyk: This will break backward compatibility, can we leave the old signature and simply call this one from it like you're doing with `getOrCreate(SparkContext)`?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
mdymczyk: Agreed, let's keep `H2OContext` clean.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
...
+
+  //  private
+  //  def finalizeFrame[T](keyName: String,
+  //                       res: Array[Long],
+  //                       colTypes: Array[Byte],
+  //                       colDomains: Array[Array[String]] = null):Frame = {
+  //    val fr:Frame = DKV.get(keyName).get.asInstanceOf[Frame]
+  //    water.fvec.FrameUtils.finalizePartialFrame(fr, res, colDomains, colTypes)
+  //    fr
+  //  }
```
mdymczyk: Should this be removed or uncommented and used somewhere?

jakubhava: both `initFrame` and `finalizeFrame` are now in `ConverterUtils` and are called in `convert[T]` method also in `ConvertorUtils` 

I would say remove it here for sure, it should be only at one place. It's also related to the fact that this part of code in `H2OContext` doesn't use the new API.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
     * @return H2O Context
     */
   def getOrCreate(sc: SparkContext): H2OContext = {
-    getOrCreate(sc, new H2OConf(sc))
+    getOrCreate(sc, new H2OConf(sc))(SQLContext.getOrCreate(sc))
+  }
+
+  val defaultFieldNames = (i: Int) => "f" + i
+
+  def buildH2OFrame(kn: String, vecTypes: Array[Byte], res: Array[Long]): H2OFrame = {
```
jakubhava: This function seems like internal function which should not be visible on H2OContext, therefore I wouldn't put it in H2OContext and `ConvertorUtils` should stay private to convertors.

Maybe better than defining this function would be to use `water.fvec.Frame` constructor 
`public Frame( Key key, String names[], Vec vecs[] )` in places you need to create a frame.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -255,7 +265,142 @@ object H2OContext extends Logging{
...
+
+      val ftypes = first.productIterator map inferFieldType
+
+      // Collect H2O vector types for all input types
+      val vecTypes:Array[Byte] = ftypes map dataTypeToVecType toArray
+
+      MetaInfo(fnames, vecTypes)
+    }
+
+    def metaInfo(tuples: List[(String, Type)]): MetaInfo = {
```
mdymczyk: Lets call this `fields` or something more descriptive than `tuples`.

core/src/test/scala/org/apache/spark/h2o/H2OFrameToDataFrameTestSuite.scala
```diff
@@ -56,19 +56,19 @@ class H2OFrameToDataFrameTestSuite extends FunSuite with SharedSparkTestContext
...
-      & h2oFrame.vec(7).isNumeric & h2oFrame.vec(8).isNumeric)
-    val rdd = hc.asRDD[Prostate](h2oFrame)
-
-    assert (rdd.count == h2oFrame.numRows())
-    assert (rdd.take(5)(4).productArity == 9)
-    assert (rdd.take(8)(7).AGE.get == 61)
-
-    h2oFrame.delete()
-  }
+//  test("H2OFrame[T_NUM] to RDD[Prostate]") {
```
mdymczyk: Is this test failing? Or is it not necessary and should be removed?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+    }
+  }
+
+  class H2ORDDIterator(val keyName: String, val partIndex: Int) extends H2OChunkIterator[A] {
+
+    private def cell(i: Int) = {
+      val j = columnMapping(i)
+      val ex = extractors(i)(converterCtx)
+      val data = ex(j).asInstanceOf[Object]
+//      println(s"@$i/$j = $data")
```
mdymczyk: 🔫

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -58,63 +62,166 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
+      data
+    }
+
+    def extractRow: Option[Array[AnyRef]] = {
+      val rowOpt = opt {
+        val objects: IndexedSeq[Object] = columnTypeNames.indices map cell
+        val row = objects toArray
+
+//        println(s"${converterCtx.rowIdx} -> ${res mkString ":"}")
```
mdymczyk: 🔫

core/build.gradle
vpatryshev: Well, can comment it out.

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: I have no objections. Removing this parameter.

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Not sure if I understood you correctly. You mean, moving H2OContext class
to converts?

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Makes sense, but that's a different change. It has a nothing to do with adding Dataset. Let's add a TODO to the file, and do it next, ok?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: It's a different change. Can do it next after this one.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: No no, keep h2o context here please. But `perTypeDataPartition` and related functionalities should be in its own class in `converters` package, not in H2OContext

core/src/main/scala/org/apache/spark/h2o/SupportedDataset.scala
vpatryshev: This was easy to implement.

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
vpatryshev: I agree; but why not separate this step?

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
vpatryshev: The original version is just above this one.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
vpatryshev: That's while H2OContext is still in h2o. Once we move it over, the trait will be again package-private.

core/src/main/scala/water/api/RDDs/RDDsHandler.scala
vpatryshev: Will fix.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: The trick is called strong typing. String typing is not a good practice, it's easy to make a mess. It's popular, but it's not good.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: New types should be introduced. They specify exactly what we mean. Throwing in tuples and strings and arrays is not a good Scala practice.

Regarding the word `Reader`, it's a nice word, but it's way too popular in a general context. Are we dealing with a Reader monad? Most probably not.

As to parameterizing the serializer, it would be nice, and there's a lot of reason to do that, but I do not see how it is possible without using `HList`. The extractor map will be heterogeneous. Anyway, will think about it.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev:
```
  val columnMapping: Map[Int, Int] =
    if (columnTypeNames.size == 1) Map(0->0) else multicolumnMapping
```
It is not the same thing. If there's one column, nobody cares about column names (judging by the existing tests)

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: I would it it as part of this commit since you created this code as part of this change and also otherwise we would have too kinds of different APIs at the same time. Code for transforming is already not in H2OContext so proper integration with current api would be to put it into converters and unify it wit the existent api.

From my point of view making the code to play nicely with the rest of the code is also part of any change, even though the primary goal of that change is different.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: I don't see it in master branch.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: ok

core/src/main/scala/org/apache/spark/h2o/SupportedDataset.scala
jakubhava: I dont understand the comment to the original suggestion,  can you please make it more clear ?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: I think I still don't agree, please see the comment above. But let's also wait for comment from the other guys.

What do you think @mmalohlava @mdymczyk ?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Actually, here's the trick. Constructors are checked previously. Now we scan the data. If the data are nullable, and the product type does not accept an option, we skip that kind of data. There's more than one case when invalid data are being skipped, as I understand; this is one of them.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: This is the right implementation of an iterator. that may cache or skip the data. That's how they do it in Scala.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: That's how optional data are handled. When we pass a None to serializer, it's converted to null. When we read it, null is converted back to None.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: I guess this is not addressed to me.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Done

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: with pleasure; done.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: made it private

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: done

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
mmalohlava: What about implicit parameter? We are already using that in `asDataFrame` calls.
Do we need SQLContext directly inside H2OContext?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
mmalohlava: can we do `private[spark]` it is fine for me to limit visibility in this way.

core/src/main/scala/org/apache/spark/h2o/utils/H2OTypeUtils.scala
mmalohlava: Neat! 👍

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
jakubhava: yup, I'm good with restriction like that, but still don't see why moving `H2OContext` somewhere. What are your intentions @vpatryshev with this please ?

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
jakubhava: please see the comment above

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
mmalohlava: I am ok with that - it is private type - valid only in the context of the class - let's say domain specific type.

However, if it makes sense to share type definition cross code, then we should put it at the package level definition. Or any ideas?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
mmalohlava: Ups, if it is missing in the master, that's my fault - sorry.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: Cool, didn't know that. Do you pls have some materials where I could get more info about the problem in general ?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: Ok, but still, `NA` is "regular" type in H2O so we should handle it properly.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: Thanks for explanation. Would be great if this is part of the code in a form of comment for people like me :)

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
mmalohlava: hmm, it depends on `converterCtx` implementation - if it follows H2O semantics, then you need to check for NAs before reading data. If converterCtx check for NA and returns null, then it is fine.

jakubhava: We can hide the semantic into `convertorCtx`. This way it could be less error-prone. 

Right now we follow H2O semantics and have to check for NA explicitly. I can hide that into converter context, actually, I like hiding it better.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: oki, thanks for explanation.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Ok, that's easy, will do.

core/src/test/scala/org/apache/spark/h2o/H2OFrameToDataFrameTestSuite.scala
vpatryshev: Oh, thanks. Restoring it.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: I have moved it all to convertors. I use InternalReadConvertorContext, but well, it works with primitive types only. You want me to move over all that SupportedDataset to IRCC? Just asking for confirmation; it's doable.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: There's a pretty similar piece of code in ProductRDDConverter; I think both should be changed to use IWCC. Just checking to confirm; I don't mind including it.

core/src/main/scala/org/apache/spark/h2o/SupportedDataset.scala
vpatryshev: What I mean: the class is moved to convertors now. :) It was easy.

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
vpatryshev: Done, I guess. Split and moved.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Sure, if it's going to be shared.

BTW, I figured, there's no way we can parameterize it. It contains data for a constructor, and a constructor receives an array of `Object`s anyway...

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: done

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: There's a discussion e.g. here: http://stackoverflow.com/questions/19492542/understanding-what-type-keyword-does-in-scala
In general, I guess, it was being discussed at some of Scala Bay talks, forgot which one.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: &#x2714;

vpatryshev: ✓

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: ✓

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Converters check for NA anyway, no?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Sure!

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: 10x! Fixed.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
vpatryshev: Restored now.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: Nope, we have to do it explicitly so far. But as @mmalohlava mentioned, it would be good if they do.

I'll create branch based on your dataset branch @vpatryshev with this change and please feel free to integrate it with your branch before we merge this one.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: If I'm not wrong `ProductRDDConverter` already uses  `InternalWriteConvertorContext`.

You're right, both should use `InternalWriteConvertorContext`.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Yeah, we need every code that transforms H2OFrames <--> RDDs/DataFrames/Datasets to use `WriteConvertorContext` and `ReadConvertorContext`, also for creation of these contexts `CovertorUtils` should be used instead of `water.fvec.FrameUtils.createNewChunks` right away.

Basically the Read/Write contexts create abstractions of querying data right from chunk or from network from different h2o node. Chunk will also give you only primitive types if I'm not wrong.

If you could do it, that would be great.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: this line `ConverterUtils.getIterator[A](isExternalBackend, iterator)` is in the master, at the end of `H2ORDD` branch.

You can either keep this method call there, or I will have to add this call there as part of integration with SW-178.
For Internal backend ( this one ) it just forwards the iterator, for the external it has to do some buffering.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Hmm, conflicting signals I hear... I was convinced that anyway it's a
singleton, and can be retrieved from static.

Thanks,
-Vlad

On Thu, Aug 25, 2016 at 10:23 AM, Michal Malohlava <notifications@github.com

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDDLike.scala
vpatryshev: You mean, instead of private[converters]?

Thanks,
-Vlad

On Thu, Aug 25, 2016 at 10:27 AM, Michal Malohlava <notifications@github.com

vpatryshev: H2OContext was not moved. Its inner class was moved, as requested. The rest
is where it was.

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Let's discuss it today; this can be done in the next iteration, I guess. I
plan to remove some code duplication; this one can be incorporated, to
streamline the architecture.

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Done; thanks!

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: yeah, but if we postpone it, we will be breaking current internal API by having 2 versions, which I don't like + future integration with another backend will be harder.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: Yup, I read about in on stackoverflow, but still don't see the benefit of doing `type TypeName = String` instead of using just `String`

vpatryshev: The benefit is that it's much harder to mix two string parameters if they
have specific types.
Also, types may be turned into value classes, for better type safety (not
just token one).

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
vpatryshev: Do I mind? No, I don't. It's more a a production issue, I guess.

Michał?

Thanks,
-Vlad

jakubhava: Well, from my point of view we all should try to make consistent code, even on internal level, but if that's something you think we can afford not to do, let's merge. Though, this way our technical dept will be higher and higher.

vpatryshev: Ok, so your stuff will be merged, I'll have to rebase anyway, and will work
on this new thing.

Thanks,
-Vlad

PR #60, opened 2016-08-26
[SW-186] Hide checks whether incoming data is NA into convertorCtx

jakubhava:
This is needed for @vpatryshev dataset PR and also it's easier to read solution

I'll do corresponding change in https://github.com/h2oai/sparkling-water/pull/49, this should be merge to unblock Vlad and merging of Dataset branch

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
@@ -38,13 +38,13 @@ class InternalReadConverterContext(override val keyName: String, override val ch
 
   override def getLong(columnNum: Int): Long =  if(isNA(columnNum)){
     null.asInstanceOf[Long]
-  }else{
+  }else {
     chks(columnNum).at8(rowIdx)
   }
 
   override def getDouble(columnNum: Int): Double =  if(isNA(columnNum)){
     null.asInstanceOf[Double]
-  }else{
+  }else {
     chks(columnNum).atd(rowIdx)
   }
 
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
@@ -36,23 +36,35 @@ class InternalReadConverterContext(override val keyName: String, override val ch
 
   override def isNA(columnNum: Int): Boolean = chks(columnNum).isNA(rowIdx)
 
-  override def getLong(columnNum: Int): Long =  chks(columnNum).at8(rowIdx)
+  override def getLong(columnNum: Int): Long =  if(isNA(columnNum)){
+    null.asInstanceOf[Long]
+  }else{
+    chks(columnNum).at8(rowIdx)
+  }
 
-  override def getDouble(columnNum: Int): Double =  chks(columnNum).atd(rowIdx)
+  override def getDouble(columnNum: Int): Double =  if(isNA(columnNum)){
+    null.asInstanceOf[Double]
+  }else{
+    chks(columnNum).atd(rowIdx)
+  }
 
-  override def getString(columnNum: Int): String = {
-    if (chks(columnNum).vec().isCategorical) {
-      val str = chks(columnNum).vec().domain()(chks(columnNum).at8(rowIdx).asInstanceOf[Int])
-      str
-    } else if (chks(columnNum).vec().isString) {
-      chks(columnNum).atStr(valStr, rowIdx) // TODO improve this.
-      valStr.toString
-    } else if (chks(columnNum).vec().isUUID) {
-      val uuid = new UUID(chks(columnNum).at16h(rowIdx), chks(columnNum).at16l(rowIdx))
-      uuid.toString
-    } else{
-      assert(assertion = false, "Should never be here")
-      null
+  override def getString(columnNum: Int): String = if(isNA(columnNum)){
+    null.asInstanceOf[String]
+  }else {
+    {
+      if (chks(columnNum).vec().isCategorical) {
+        val str = chks(columnNum).vec().domain()(chks(columnNum).at8(rowIdx).asInstanceOf[Int])
+        str
+      } else if (chks(columnNum).vec().isString) {
+        chks(columnNum).atStr(valStr, rowIdx) // TODO improve this.
+        valStr.toString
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -49,11 +49,7 @@ trait ReadConverterContext {
   def getTimestamp(columnNum: Int): Long  = getLong(columnNum) * 1000L
   def getUTF8String(columnNum: Int): UTF8String = {
     val str = getString(columnNum)
-    if(str == null){
-      null
-    }else{
-      UTF8String.fromString(str)
-    }
+    if(str == null) null else UTF8String.fromString(str)
```
mdymczyk: this can be changed to `def getUTF8String(columnNum: Int): UTF8String = UTF8String.fromString(getString(columnNum))`

core/src/main/scala/org/apache/spark/h2o/converters/H2ODataFrame.scala
```diff
@@ -98,21 +98,17 @@ class H2ODataFrame[T <: water.fvec.Frame](@transient val frame: T,
...
+            case ByteType => mutableRow.setByte(idx, converterCtx.getByte(i))
+            case ShortType => mutableRow.setShort(idx, converterCtx.getShort(i))
+            case IntegerType => mutableRow.setInt(idx, converterCtx.getInt(i))
+            case LongType => mutableRow.setLong(idx, converterCtx.getLong(i))
+            case FloatType => mutableRow.setFloat(idx, converterCtx.getFloat(i))
+            case DoubleType => mutableRow.setDouble(idx, converterCtx.getDouble(i))
+            case BooleanType => mutableRow.setBoolean(idx, converterCtx.getBoolean(i))
+            case StringType => mutableRow.update(idx, converterCtx.getUTF8String(i))
+            case TimestampType => mutableRow.setLong(idx, converterCtx.getTimestamp(i))
+            case _ => ???
```
mdymczyk: I know this was here before but can we add a better exception here? `NotImplementedError` might be vague, something saying that `typ` is not supported wouldn't be better?

jakubhava: Yup, that make sense. I just wanted to do the smallest change for @vpatryshev  - we needed to hide calling of `isNA`.

I'm up to change it, but i would probably do it after @vpatryshev's dataset branch is merged.

Is it ok with you ?

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
jakubhava: thanks!

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -96,17 +96,15 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@transient val
...
-            case q if q == classOf[java.lang.Double]  => Some(converterCtx.getDouble(idx))
-            case q if q == classOf[java.lang.Float]   => Some(converterCtx.getFloat(idx))
-            case q if q == classOf[java.lang.Boolean] => Some(converterCtx.getBoolean(idx))
+          data(idx) = types(idx) match {
+            case q if q == classOf[Integer]           => Option(converterCtx.getInt(idx))
+            case q if q == classOf[java.lang.Long]    => Option(converterCtx.getLong(idx))
+            case q if q == classOf[java.lang.Double]  => Option(converterCtx.getDouble(idx))
+            case q if q == classOf[java.lang.Float]   => Option(converterCtx.getFloat(idx))
+            case q if q == classOf[java.lang.Boolean] => Option(converterCtx.getBoolean(idx))
             case q if q == classOf[String] => Option(converterCtx.getString(idx))
```
vpatryshev: This is cool... but maybe converterCtx should return an Option?

jakubhava: If it would return Option, then all these calls 

`mutableRow.setByte(idx, converterCtx.getByte(i))` would have to be like 
`mutableRow.setByte(idx, converterCtx.getByte(i).getOrElse(null))` because we need to propagate `null` here

In this case I would rather return like this and create Option out of it when needed, but it's probably just matter of style.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -36,23 +36,35 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
-      valStr.toString
-    } else if (chks(columnNum).vec().isUUID) {
-      val uuid = new UUID(chks(columnNum).at16h(rowIdx), chks(columnNum).at16l(rowIdx))
-      uuid.toString
-    } else{
-      assert(assertion = false, "Should never be here")
-      null
+  override def getLong(columnNum: Int): Long =  if(isNA(columnNum)){
+    null.asInstanceOf[Long]
+  }else{
```
mmalohlava: This is really, really minor comment (is not blocking merge): the syntax is not consistent :) `}else{` and `}else {`.

I can enable check style rule for that, if we want.

mmalohlava:
👍

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
vpatryshev: Good point... what is the byte equal to before it's being assigned?

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
jakubhava: thanks, it's good to be consistent.

Scala checks are good for me with these cases.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: What do you mean please ? Thanks for clarification

vpatryshev: I was asking, if we have a None, and want to set the byte to null, what was it before? But then, well, I see the habit of using mutables, so this question does not make sense. Have to override anyway.

PR #63, opened 2016-08-28
[FIX] Fix read converter implementation

vpatryshev:


vpatryshev:
There's a bug; fixing.

vpatryshev:
Bug fixed now.

core/src/test/scala/org/apache/spark/h2o/H2OFrameToRDDTestSuite.scala
```diff
@@ -231,6 +271,19 @@ class H2ORDDTest extends FunSuite with SharedSparkTestContext {
...
+  private def assertInvariantsWithNulls[T<:Product](rdd: RDD[T], df: H2OFrame, rowAssert: RowValueAssert): Unit = {
+    assertHolderProperties(df)
+    assert (rdd.count == df.numRows(), "Number of rows in H2OFrame and RDD should match")
+    // Check numbering
+
+    val vec = df.vec(0)
+    var row = 0
+    while(row < df.numRows()) {
+      rowAssert (row, vec)
+      row += 1
```
jakubhava: Maybe `rowIdx` would be better ?

core/src/test/scala/org/apache/spark/h2o/H2OFrameToRDDTestSuite.scala
```diff
@@ -51,6 +51,28 @@ class H2ORDDTest extends FunSuite with SharedSparkTestContext {
...
 
+  test("RDD[IntHolder] with nulls, does it hold nulls?") {
+
+    def almostDefined(i: Long) = Some(i.toInt) filter (_ % 31 != 0)
+
+    val rdd = sc.parallelize(1 to 1000, 100).map( v => IntHolder(almostDefined(v)))
+    val h2oFrame:H2OFrame = hc.asH2OFrame(rdd)
+
+    assertInvariantsWithNulls(rdd, h2oFrame, (row, vec) => {
+      val row1 = row + 1
```
jakubhava: Maybe rowIdx would be better ? When I read it for first time I thought `row` represents really a row

core/src/main/scala/org/apache/spark/h2o/converters/H2ODataFrame.scala
```diff
@@ -92,31 +94,25 @@ class H2ODataFrame[T <: water.fvec.Frame](@transient val frame: T,
         expectedTypes,
         partIndex)
 
+      /*a sequence of converters, per column*/
+      private val columnConverters = filteredTypes map converterCtx.get
+
+
+      private def readRow: InternalRow = {
+          val optionalData: Seq[Option[Any]] =
+            columnConverters.zipWithIndex map { case (c, i) => c(selectedColumnIndices(i)) }
```
jakubhava: Instead of i `idx`, instead of c `convertPerCol`, just to make it more readable ?

jakubhava:
I just put 2 minor comments above, otherwise I really like this change, thanks! :)

vpatryshev:
Thanks; implemented.

Tests pass.

Thanks,
-Vlad

core/src/main/scala/org/apache/spark/h2o/converters/H2ODataFrame.scala
```diff
@@ -92,31 +94,27 @@ class H2ODataFrame[T <: water.fvec.Frame](@transient val frame: T,
         expectedTypes,
         partIndex)
 
+      /*a sequence of converters, per column*/
+      private val columnConverters = filteredTypes map converterCtx.get
+
+
+      private def readRow: InternalRow = {
+          val optionalData: Seq[Option[Any]] =
```
mmalohlava: I would be interested in comparing performance of this code - will create JIRA for that

vpatryshev: Comparing to which version?

Thanks,
-Vlad

mmalohlava: Comparing the original implementation without convertors. I would like to know performance cost of using one more indirection. But right now, it is low priority...

PR #65, opened 2016-08-30
[MINOR] Client IP fetched from system properties.

mdymczyk:
Previously used `InetAddress.getLocalHost.getHostAddress` as client IP can return an IP address which is not supported by an H2O node since it internally goes through a nameserver and H2O supports only IPs from network interfaces.

This is a setting depending on where it runs so I think system property isn't a bad idea, thoughts?


jakubhava:
Yup, environmental property make sense in this case.

PR #67, opened 2016-08-31
[SW-158] Support for Spark DataSet

vpatryshev:
This stuff is working; I'll need to work on code deduplication, can be done later.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -28,7 +28,7 @@ import org.apache.spark.sql.{DataFrame, SQLContext}
 import water._
 
 import scala.collection.mutable
-import scala.language.implicitConversions
+import scala.language.{implicitConversions, postfixOps}
 import scala.reflect.ClassTag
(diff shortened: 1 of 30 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -61,6 +61,7 @@ class H2OContext private (@transient val sparkContext: SparkContext, @transient
   with Serializable with SparkDataFrameConverter with SupportedRDDConverter with H2OContextUtils{
   self =>
 
+  val sqlc: SQLContext = SQLContext.getOrCreate(sparkContext)
```
mmalohlava: I am changing this val into `@transient` to make sure we are not capturing any non-serializable entities.

vpatryshev: Oh. Thanks a lot. The class is serializable.

vpatryshev:
Wow!

mmalohlava:
Thx @vpatryshev! (i squashed that, test that on yarn, and merge)

PR #70, opened 2016-09-01
[SW-192] Add org.apache.spark.sql._ to packages imported by default in REPL

jakubhava:
Right now this package is not imported and for example when specifying saving mode in datasource api using`SaveMode.Append `
it doesn't know the class SaveMode.
It's more convenient for the user not having to import this package.


repl/src/main/scala/org/apache/spark/repl/h2o/H2OInterpreter.scala
```diff
@@ -126,6 +126,7 @@ class H2OInterpreter(val sparkContext: SparkContext, var sessionId: Int) extends
           "org.apache.spark.sql.{DataFrame, Row, SQLContext}",
           "sqlContext.implicits._",
           "sqlContext.sql",
+          "org.apache.spark.sql._",
```
mmalohlava: Kuba, will this work with 1.6 and 2.0 ? Or it is Spark version specific?

jakubhava: This package exists in both spark 1.6 and 2.0, so it shouldn't break anything

PR #71, opened 2016-09-01
[SW-193] Add scala version to pysparkling package name

jakubhava:
We need to embed the scala version in the name since spark can be built with different scala versions. We need to have packages for all of them and we need to somehow distinguish them from each other.

### DIFF
--- a/py/PUBLISHING_TO_PIPY.txt
+++ b/py/PUBLISHING_TO_PIPY.txt
@@ -36,5 +36,6 @@ python setup.py sdist upload -r pypi
 
 
 5 - Don't forget to do this for each pySparkling release ( for Spark 1.5, Spark 1.6 ... )
+    and for each spark release publish it for both scala versions 2.10 and 2.11
 
 6 - That's it!
--- a/py/README.rst
+++ b/py/README.rst
@@ -2,15 +2,15 @@ PySparkling and Spark Version
 =============================
 There exist multiple pySparkling packages, each is intended to be used with different Spark version.
 
- - h2o_pysparkling_1.6 - for Spark 1.6.x
- - h2o_pysparkling_1.5 - for Spark 1.5.x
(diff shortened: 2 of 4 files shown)


vpatryshev:
Great idea.

mmalohlava:
Are we sure about that? We know default Scala version of Spark distribution - it is 2.10 for <2.0, it is 2.11 for 2.0. I like it, but it can be confusing for users.

The question is what is the intention/motivation of encoding the scala version into python artifact? 
I would like to see the motivation.

From my point of view, It makes sense for binary (maven) artifacts, since you are linking them with your JVM program which can be Scala version specific. 
However for python artifacts we know that in the most of cases people running it with downloaded Spark (which has fixed Scala version). If people would like to use for example Spark 2.0 compiled with Scala 2.10 they have to do that manually (=build the Spark by themselves).

jakubhava:
At first, I wanted to avoid the need for build sparkling water and provide all the versions as python packages. In python package wrap sparkling water assembly jar, which is built with either scala 2.10 or 2.11, so I wanted to be transparent to the user.

It's however true that when people download spark and they plan to use pyspark anyway, they will probably download default spark, unless they have really specific needs.

So I would say I'm convinced not to merged this

mmalohlava:
It is good systematic change, however, from user point of view it can be confusing. @vpatryshev @mdymczyk any comments on this?

mmalohlava:
Oki, keep it open for now and wait for users feedback.

vpatryshev:
No opinion on my side.

Thanks,
-Vlad

jakubhava:
We decided to close this one and just write instructions how to build pysparkling in case somebody would like to use it with non default scala version ( which for spark 2.0 is 2.11 ). I already create JIRA for it https://0xdata.atlassian.net/browse/SW-211

PR #72, opened 2016-09-01
[SWPRIVATE-11] Spark SVM model support in H2O core.

mdymczyk:
Wrapper around Spark MLlib's SVM model in H2O.

Sample code demo can be found SparkSVMDemo or Spark_SVM_Model.flow.

We support Spark parameters (https://spark.apache.org/docs/latest/mllib-linear-methods.html#linear-support-vector-machines-svms) plus two modes:

regression (threshold has to be set to NaN and the response column has to be numeric)
binomial (threshold has to be set to a double value and the response column has to be an enum with domain of size 2)
Things to improve/discuss in the future:

the job is currently not cancellable. This might be a problem if a user thinks the job is running too long or simply wants to kill the job (maybe he mistyped a parameter). But adding an option to cancel might be misleading when compared with H2O models which are created from intermediate state even when the job is cancelled (impossible with Spark). Thoughts? The more I think about it adding possibility to cancel with a proper log message (no model was created) seems like the better option.
missing values - this is not supported by Spark, we can tell the user to simply use Spark for ETL or we can think about a "Spark global" strategy for this:
ignores rows with missing values (might be a lot of rows)
impute the values with median/average
in both cases would this be a per column decision or a global one? I can see user wanting to mix them depending on the column but not sure how easy it would be to implement in Flow
Caveats:

progress bar is not being reported currently, tried grabbing this info from MLlibs internals but its all on method local level...


mdymczyk:
Will be fixed in a patch PR instead.

PR #74, opened 2016-09-02
[SWPRIVATE-18] Progress bar update for SVM

mdymczyk:
Not 100% but close. SVM has one main spark job that is ran in a for loop for the number of iterations specified by the user (default 100) - it's called `treeAggregate`. There are a few other jobs that run before (3-5) that's why it's not 100% accurate but I think it's close enough to leave it as it is - otherwise we'd have to find all the jobs spark uses for SVM and keep track of all the changes they do to them.

I'm using RDD info to update a job only when a job is submitted on the appropriate RDD (we can have multiple models being built at the same time).

Thoughts?

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
@@ -16,25 +16,81 @@
  */
 package org.apache.spark.ml.spark
 
+import org.apache.spark.SparkContext
 import org.apache.spark.scheduler._
 import org.apache.spark.storage.RDDInfo
 import water.Job
+
+import scala.collection.mutable
 
 /**
   * One progress listener for each instance of model creation. Each listener gets all events so this particular one
   * will respond only to the ones corresponding to a given RDD used for training (which should not be shared) and
   * to a particular job type (depending on the algorithm).
   *
-  * It might not be 100% accurate
+  * It might not be 100% accurate depending on how many "non primary" jobs will be run.
+  *
+  * @param job Job which progress is to be updated
+  * @param rddInfo events triggered for this RDD (or its children) trigger progress change
+  * @param stageNames which stages are supposed to change the progress bar %
   */
-class ProgressListener(val job: Job[_], val rddInfo: RDDInfo, val stageNames: Iterable[String]) extends SparkListener {
+class ProgressListener(val sc: SparkContext,
+                       val job: Job[_],
+                       val rddInfo: RDDInfo,
+                       val stageNames: Iterable[String]) extends SparkListener {
+
+  var currentJobs: scala.collection.mutable.Map[Int, Int] = mutable.HashMap()
+
+  private def matchingRDD(stageInfo: StageInfo): Boolean = stageInfo.rddInfos.map(_.id).contains(rddInfo.id)
+  private def matchingRDD(stageInfos: Seq[StageInfo]): Boolean = stageInfos.exists(matchingRDD)
 
   override def onJobStart(jobStart: SparkListenerJobStart): Unit = {
-    if(jobStart.stageInfos.exists{ stage =>
-      stageNames.exists( name => stage.name.contains(name) ) && stage.rddInfos.map(_.id).contains(rddInfo.id)
-    }) {
-      job.update(1)
+    if (matchingRDD(jobStart.stageInfos)) {
+      job.update(0, s"Staring Spark job with ID [${jobStart.jobId}]")
+
+      if (jobStart.stageInfos.exists(stage => stageNames.exists(stage.name.contains))) {
+        currentJobs.put(jobStart.jobId, 1)
+      } else {
+        currentJobs.put(jobStart.jobId, 0)
+      }
     }
   }
+
+  override def onJobEnd(jobEnd: SparkListenerJobEnd): Unit = {
+    if(currentJobs.contains(jobEnd.jobId)) {
+      job.update(currentJobs.get(jobEnd.jobId).get, s"Finished Spark job with ID [${jobEnd.jobId}]")
+      currentJobs.remove(jobEnd.jobId)
+    }
+  }
+
+  var currentStages: scala.collection.mutable.Map[Int, String] = mutable.HashMap()
+
+  override def onStageSubmitted(stageSubmitted: SparkListenerStageSubmitted): Unit = {
+    if(matchingRDD(stageSubmitted.stageInfo)) {
+      updateTaskStatus(stageSubmitted.stageInfo.stageId, 0)
+    }
+  }
+
+  override def onStageCompleted(stageCompleted: SparkListenerStageCompleted): Unit = {
+    if(currentStages.contains(stageCompleted.stageInfo.stageId)) {
+      currentStages.remove(stageCompleted.stageInfo.stageId)
+      job.update(0, printStagesStatus())
+    }
+  }
+
+  override def onTaskStart(taskStart: SparkListenerTaskStart): Unit = {
+    if(currentStages.contains(taskStart.stageId)) {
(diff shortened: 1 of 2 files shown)


ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
```diff
@@ -0,0 +1,40 @@
...
+  *
+  * It might not be 100% accurate
+  */
+class ProgressListener(val job: Job[_], val rddInfo: RDDInfo, val stageNames: Iterable[String]) extends SparkListener {
+
+  override def onJobStart(jobStart: SparkListenerJobStart): Unit = {
+    if(jobStart.stageInfos.exists{ stage =>
+      stageNames.exists( name => stage.name.contains(name) ) && stage.rddInfos.map(_.id).contains(rddInfo.id)
+    }) {
+      job.update(1)
```
mmalohlava: I would call `job.update(1, msg)` - where `msg` is similar to messages which are printed by Spark UI - `msg` ~ _"Spark: processing job XXX (23/100)"_

At least user will see that there is something working

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVM.java
```diff
@@ -188,12 +192,19 @@ public void computeImpl() {
             svm.optimizer().setGradient(_parms._gradient.get());
             svm.optimizer().setUpdater(_parms._updater.get());
 
+            ProgressListener progressBar = new ProgressListener(_job, RDDInfo.fromRdd(training), asScalaIterable(Arrays.<String>asList("treeAggregate")));
```
mmalohlava: Is the `treeAggregate` constant exposed by SVM code?

ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
```diff
@@ -0,0 +1,40 @@
...
+  * One progress listener for each instance of model creation. Each listener gets all events so this particular one
+  * will respond only to the ones corresponding to a given RDD used for training (which should not be shared) and
+  * to a particular job type (depending on the algorithm).
+  *
+  * It might not be 100% accurate
+  */
+class ProgressListener(val job: Job[_], val rddInfo: RDDInfo, val stageNames: Iterable[String]) extends SparkListener {
+
+  override def onJobStart(jobStart: SparkListenerJobStart): Unit = {
+    if(jobStart.stageInfos.exists{ stage =>
```
mmalohlava: Thinking about it little bit more - on job start, we should set H2O's job total amount of work.

Then we can use `onStageCompleted` to move progress bar
and `onStageStarted` to change progress message.

mmalohlava:
I think you are on the good way, you can try to report more with the listener callbacks:
```
diff --git i/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala w/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
index 8223af9..ce3acf7 100644
--- i/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
+++ w/ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
@@ -30,11 +30,20 @@ import water.Job
 class ProgressListener(val job: Job[_], val rddInfo: RDDInfo, val stageNames: Iterable[String]) extends SparkListener {

   override def onJobStart(jobStart: SparkListenerJobStart): Unit = {
-    if(jobStart.stageInfos.exists{ stage =>
-      stageNames.exists( name => stage.name.contains(name) ) && stage.rddInfos.map(_.id).contains(rddInfo.id)
+    if (jobStart.stageInfos.exists { stage =>
+      stageNames.exists(name => stage.name.contains(name)) && stage.rddInfos.map(_.id).contains(rddInfo.id)
     }) {
       job.update(1)
     }
   }

+  override def onStageSubmitted(stageSubmitted: SparkListenerStageSubmitted): Unit = {
+    val si = stageSubmitted.stageInfo
+    job.update(1, s"${si.name} ${si.attemptId} ${si.numTasks} ${}")
+  }
+
+  override def onTaskStart(taskStart: SparkListenerTaskStart): Unit = {
+    job.update(1, taskStart.stageId + " " + taskStart.taskInfo.id)
+  }
+
 }```

However, you would need a few changes on H2O side like:
  - changing total amount of work 
  - changing only status message, not progress.

ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
mdymczyk: Hm not sure if I can change the total amount of work when the model is being built already, need to check.

Working with Spark's stages rather than jobs might be a bit confusing, though, here one job = one iteration of the algorithm that's why I can set the total amount of work to the number of iterations the user set in the UI and then just increment it by 1 with each new job.

It should be recorded on job start which job is running and incremented on job end instead but I was worried about this being run in parallel on many nodes and being harder to track but I can try.

ml/src/main/scala/org/apache/spark/ml/spark/ProgressListener.scala
mdymczyk: Hm I'm a bit confused here, `job.update(1)` will increment the progress by 1 so the user does see progress, I can update the message, too if you want of course.

mmalohlava: Yup, updating progress bar is good, but we can also expose some insight what we are doing inside: "Spark is doing XYZ".

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVM.java
mdymczyk: Unfortunately no, Spark gets the name dynamically from code so the whole name is `treeAggregate at GradientDescent:227` so from which file, which line and as what method was the job submitted. That also why it's going to be hard to track their changes in code should they change the name of the method.

The only other way I figured out is not to check the `stageName` in the listener and just progress the bar on each `jobStart` for a given RDD but then the number of such event will be bigger than the total number set in SVM algo (and there's no way to check how many jobs will be submitted by spark beforehand).

mdymczyk:
@mmalohlava actually both can be done already, total amount of work by changing the value of `_work` in `Job` (though I haven't tried that one myself) and the status message with no progress change by just passing 0 (which I'm doing now).

How about the current solution:

- changes the message on each relevant job start/finish (relevant = for this RDD)
- changes the progress only when a job with a given name finishes (for example if we want to run SVM for 100 iterations then Spark runs 100x `treeAggregate` jobs so I think it's a decent approximation)
- changes the message also on relevant stage/task submission (like Spark console). Actually some messages might not get shown as they change so fast but I'd prefer that rather than slowing the computation down.

As for changing the total number of work on the fly I still don't see how that would work?

mmalohlava:
> As for changing the total number of work on the fly I still don't see how that would work?
It is doing atomic update of Job in DKV.

mmalohlava:
Is this PR ready to be merged?

mdymczyk:
@mmalohlava yes, it's ready - just waiting for final code review :-)

In the end I don't change the total number of work and I'm sticking with the number of iterations because:

- if I decide to represent the total number of work as the total number of tasks in a stage (and change it for every stage) then the bar will be going from 0 to 100% all the time
- if I decide to represent the total number of work the total number of jobs submitted by spark (which is usually number of iterations + a few) then I would have to start at 0 for both "work done" and "total work" and just increment both at the same time so the progress bar would be at a 100% no? Unless you were thinking of a way that would allow me to figure out the total number of jobs upfront and change it (but I'm not sure it's doable in Spark)?

mdymczyk:
Ok can move them after we release 2.0 - wasn't sure about the placement since most Java classes are in Scala build dirs (all Schema V3 classes, JavaH2OContext etc).

PR #75, opened 2016-09-06
[SWPRIVATE-16] NA handling for Spark algorithms

mdymczyk:
- added NA value handling for SVM: mean imputation, skip, not allowed (old variant with an error message)
- means are calculated via RDD as H2OFrame.means() doesn't support enum columns
- changed the SVM model scoring/pojo generation to include column means when running in MeanImputation mode
- refactored tests a bit (frame creation now supports multi column frames)

### DIFF
--- a/core/src/test/scala/org/apache/spark/h2o/utils/SharedSparkTestContext.scala
+++ b/core/src/test/scala/org/apache/spark/h2o/utils/SharedSparkTestContext.scala
@@ -21,9 +21,9 @@ import java.util.concurrent.ConcurrentHashMap
 
 import org.apache.spark.SparkContext
 import org.apache.spark.h2o.{H2OConf, H2OContext, Holder}
-import org.apache.spark.sql.{SQLContext, SparkSession}
 import org.scalatest.Suite
-import water.fvec.{Chunk, FrameUtils, NewChunk, Vec}
+import water.fvec._
+import water.{DKV, Key}
 import water.parser.BufferedString
 
 import scala.reflect.ClassTag
@@ -32,9 +32,10 @@ import scala.reflect.ClassTag
   * Helper trait to simplify initialization and termination of Spark/H2O contexts.
   *
   */
-trait SharedSparkTestContext extends SparkTestContext { self: Suite =>
+trait SharedSparkTestContext extends SparkTestContext {
+  self: Suite =>
 
-  def createSparkContext:SparkContext
+  def createSparkContext: SparkContext
 
   def createH2OContext(sc: SparkContext, conf: H2OConf): H2OContext = {
     H2OContext.getOrCreate(sc, conf)
@@ -51,26 +52,51 @@ trait SharedSparkTestContext extends SparkTestContext { self: Suite =>
     super.afterAll()
   }
 
-  def buildChunks[T: ClassTag](fname: String, data: Array[T], cidx: Integer, h2oType: Array[Byte]): Chunk = {
+  def makeH2OFrame[T: ClassTag](fname: String, colNames: Array[String], chunkLayout: Array[Long],
+                                data: Array[Array[T]], h2oType: Byte, colDomains: Array[Array[String]] = null): H2OFrame = {
+    makeH2OFrame2(fname, colNames, chunkLayout, data.map(_.map(value => Array(value))), Array(h2oType), colDomains)
+  }
+
+  def makeH2OFrame2[T: ClassTag](fname: String, colNames: Array[String], chunkLayout: Array[Long],
+                                 data: Array[Array[Array[T]]], h2oTypes: Array[Byte], colDomains: Array[Array[String]] = null): H2OFrame = {
+    var f: Frame = new Frame(Key.make(fname))
+    FrameUtils.preparePartialFrame(f, colNames)
+    f.update()
+
+    for (i <- chunkLayout.indices) {
+      buildChunks(fname, data(i), i, h2oTypes)
+    }
+
+    f = DKV.get(fname).get()
+
+    FrameUtils.finalizePartialFrame(f, chunkLayout, colDomains, h2oTypes)
+
+    new H2OFrame(f)
+  }
+
+  def buildChunks[T: ClassTag](fname: String, data: Array[Array[T]], cidx: Integer, h2oType: Array[Byte]): Array[_ <: Chunk] = {
     val nchunks: Array[NewChunk] = FrameUtils.createNewChunks(fname, h2oType, cidx)
 
-    val chunk: NewChunk = nchunks(0)
-    data.foreach {
-      case u: UUID               => chunk.addUUID(
-        u.getLeastSignificantBits,
-        u.getMostSignificantBits)
-      case s: String             => chunk.addStr(new BufferedString(s))
-      case b: Byte               => chunk.addNum(b)
-      case s: Short              => chunk.addNum(s)
-      case c: Integer if h2oType(0) == Vec.T_CAT => chunk.addCategorical(c)
-      case i: Integer if h2oType(0) != Vec.T_CAT => chunk.addNum(i.toDouble)
-      case l: Long               => chunk.addNum(l)
-      case d: Double             => chunk.addNum(d)
-      case x                     =>
-        throw new IllegalArgumentException(s"Failed to figure out what is it: $x")
+    data.foreach { values =>
+      values.indices.foreach { idx =>
+        val chunk: NewChunk = nchunks(idx)
+        values(idx) match {
+          case null                                  => chunk.addNA()
+          case u: UUID                               => chunk.addUUID(u.getLeastSignificantBits, u.getMostSignificantBits)
+          case s: String                             => chunk.addStr(new BufferedString(s))
+          case b: Byte                               => chunk.addNum(b)
+          case s: Short                              => chunk.addNum(s)
+          case c: Integer if h2oType(0) == Vec.T_CAT => chunk.addCategorical(c)
+          case i: Integer if h2oType(0) != Vec.T_CAT => chunk.addNum(i.toDouble)
+          case l: Long                               => chunk.addNum(l)
+          case d: Double                             => chunk.addNum(d)
+          case x =>
+            throw new IllegalArgumentException(s"Failed to figure out what is it: $x")
+        }
+      }
     }
     FrameUtils.closeNewChunks(nchunks)
-    chunk
+    nchunks
   }
 }
 
--- a/ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
@@ -1,0 +1,134 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.ml
+
+import org.apache.spark.h2o.H2OContext
+import org.apache.spark.ml.spark.models.MissingValuesHandling
+import org.apache.spark.mllib.linalg.Vectors
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{Row, SQLContext}
+import org.apache.spark.sql.types.{DataTypes, StructField}
+import water.fvec.{Frame, H2OFrame}
+
+object FrameMLUtils {
+  /** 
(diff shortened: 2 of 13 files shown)


vpatryshev:
I think we need some kind of coordination here.

mdymczyk:
@vpatryshev you are referring to your refactoring? I'm ok with adopting your changes since most of the stuff in this commit is just moving things around in tests.

vpatryshev:
Yes, this is a good step forward, but it partially overlaps with the data handling fixes that are waiting for the merge, there's a pull request. I'd rather line up these things, I don't know, otherwise we would be committing the same stuff, code chunks fighting with each other. Michal, any updates?

jakubhava:
I would rather merge first @vpatryshev branch. I've done the review already and I'm ok with merging right away, just waiting for @mmalohlava approval and then I would put @mdymczyk behaviour on top of that.

@vpatryshev's changes seems like good base for the rest of the code build on top of that, It unifies stuff and also will play nicely with the external backend.

Would it be complicated to put your changes on @vpatryshev's branch @mdymczyk ?

But let's also wait for Michal response.

thanks!

mdymczyk:
Sure, first merge @vpatryshev changes and I'll rebase them onto this branch, no problem.

vpatryshev:
Yes, same here.

Also, I had a talk with Arno, and so the general idea of returning 0 if an
integer value is missing was considered wrong, so I had commited another
change, taking care of these issues.

Anyway, that's why I did not review @mdymczyk's change; it can happen that
most of it is actually implemented.
Sorry that we did not have enough communication here beforehand.

I actually have two more issues after talking with Arno.

First, the loop via absolute row indexes is not efficient; a nested loop,
chunks, and relative indexes inside, would make more sense. Just
performance, not sure about the actual impact, may be negligible - but this
was documented.

Second, how come UUID and Enum are not handled as such, but converted to
strings? We could as well have them.

(Or, and #3 - what's the fuss with UTF8String? Do we need, in Scala, to
have them all along with regular strings? Or just rather keep one kind
(either way), with implicit transformations when needed?)

Thanks,
-Vlad

mdymczyk:
@vpatryshev no problems.

1) true, I also saw the doc about looping through rows not being super efficient but it seemed much easier so I decided to go with it before trying something else - if it's indeed too slow we can change it

2) if you're referring to converting H2O's UUIDs to String before putting them in a DF/RDD then I guess it's because there's no UUID `DataType` in Spark, same for enums. I guess we could do it for RDDs (at least with UUID, not sure how you'd want to create enums on the fly) but don't think that would be possible with DFs and we'd stop being consistent. That's my guess. Would need to ask @mmalohlava  

3) `UTF8String` should only be used for Spark's SQL module, that's what they use internally for String types, we should not use it outside it in our code.

mdymczyk:
@vpatryshev @jakubhava how's the other refactoring going? Do you guys need any help/reviews there?

jakubhava:
I'm happy with @vpatryshev's PR in the current state, just waiting for @mmalohlava approval and merge

vpatryshev:
I'm still waiting for the merge of that previous PR.

Thanks,
-Vlad

jakubhava:
@mdymczyk I integrated the things I wanted into the master already. Could you please rebase on top of master, fix the problems and then let me know ? I will have a look and do review afterwards.

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
```diff
@@ -0,0 +1,132 @@
...
+          agg1(idx) = (otherMu - (delta * n) / (n + otherN), n + otherN)
+        } else {
+          agg1(idx) = ((mu * n + otherMu * otherN) / (n + otherN), n + otherN)
+        }
+      }
+    }
+    agg1
+  }
+
+  private[ml] def toDouble(value: Any, fieldStruct: StructField, domain: Array[String]): Double = {
```
vpatryshev: This is not very type safe; why not check value for its type, in match/case?

E.g. value match {
  case b; Byte if fieldStruct.dataType == DataTypes.ByteType => b.doubleValue
... etc ...

And I would make it less verbose.

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVM.java
```diff
@@ -100,11 +99,13 @@ public void init(boolean expensive) {
             }
         }
 
-        for (int i = 0; i < _train.numCols(); i++) {
-            Vec vec = _train.vec(i);
-            String vecName = _train.name(i);
-            if (vec.naCnt() > 0 && (null == _parms._ignored_columns || Arrays.binarySearch(_parms._ignored_columns, vecName) < 0)) {
-                error("_train", "Training frame cannot contain any missing values [" + vecName + "].");
+        if(MissingValuesHandling.NotAllowed.equals(_parms._missing_values_handling)) {
```
vpatryshev: Why .equals? Is not == good enough?

vpatryshev: You are comparing enums, right? Will == not work here?

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVM.java
```diff
@@ -174,13 +175,22 @@ public void computeImpl() {
...
-                    model._output.nfeatures()
+                    model._output.nfeatures(),
+                    _parms._missing_values_handling,
+                    h2oContext,
+                    sqlContext
             );
+            RDD<LabeledPoint> training = points._1();
             training.cache();
 
+            if(training.count() == 0 && MissingValuesHandling.Skip.equals(_parms._missing_values_handling)) {
```
vpatryshev: You are comparing enums, right? Will == not work here?

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVMModel.scala
```diff
@@ -28,6 +31,7 @@ object SVMModel {
     var interceptor: Double = .0
     var iterations: Int = 0
     var weights: Array[Double] = null
+    var numMeans: Array[Double] = null
```
vpatryshev: Using nulls in Scala is not a good practice at all. If a value is optional, use Option.
Using vars... well... it may be harder to eliminate

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVMModel.scala
```diff
@@ -47,10 +53,19 @@ class SVMModel private[svm](val selfKey: Key[SVMModel],
...
 
+  private val meanImputation: Boolean = MissingValuesHandling.MeanImputation.equals(_parms._missing_values_handling)
+
   protected def score0(data: Array[Double], preds: Array[Double]): Array[Double] = {
     java.util.Arrays.fill(preds, 0)
+
     val pred =
-      data.zip(_output.weights).foldRight(_output.interceptor){ case ((d, w), acc) => d * w + acc}
+      data.zip(_output.weights).foldRight(_output.interceptor){ case ((d, w), acc) =>
+        if(meanImputation && lang.Double.isNaN(d)) {
```
vpatryshev: So, if it's not meanImputation, it's okay to multiply by d?

mdymczyk:
Good comments, thanks @vpatryshev !

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVMModel.scala
mdymczyk: Yes it will be a NaN in that case, I can do that or throw an exception. Skipping NaN values doesn't make much sense computation wise (the MissingValuesHandling.Skip during training is skipping the whole row if it has NaNs, not singular columns).

ml/src/main/scala/org/apache/spark/ml/spark/models/svm/SVMModel.scala
mdymczyk: True but it has to work with the underlying core H2O model for Model Outputs (that's also why I had to write parts of code in Java as the underlying framework requires several completely different constructors) - not sure if Option is an... option here :-) But will look into it.

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
mdymczyk: Really like this one, thanks!

mdymczyk:
@vpatryshev @jakubhava @mmalohlava made appropriate changes according to Vlad's comments,  can I merge this in or is something still wrong? Would like to use it also for GM clustering.

mmalohlava:
@mdymczyk can you please update the branch with the latest master?

jakubhava:
I was thinking it it wouldn't be better to separate this PR to 2. The first one with updated test infrastructure and the second one with the implementation itself. To keep PR really focused on one thing.

Just an idea.

mdymczyk:
@mmalohlava sorry for the delay, completely missed that comment ;-)

@jakubhava I was thinking about it but since the whole test change isn't that big and is there only because of the NA handling feature I thought it would be ok. I can separate it into 2 PRs if you guys want but just thought it would be an overkill.

mdymczyk:
Will merge it tomorrow if there are no further complaints @jakubhava @vpatryshev @mmalohlava Need it for several other branches which are blocked by this. Could rebase off of this but then this would've to get merged too anyway...

jakubhava:
@mdymczyk I will have a look on it today ( in hour or so), sorry for the delay

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.spark.mllib.regression.LabeledPoint
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.{Row, SQLContext}
+import org.apache.spark.sql.types.{DataTypes, StructField}
+import water.fvec.{Frame, H2OFrame}
+
+object FrameMLUtils {
+  /**

+    * Converts a H2O Frame into an RDD[LabeledPoint]. Assumes that the last column is the response column which will be used
```
jakubhava: I'm thinking whether we could put this code into the converters. So when the user calls `h2oContext.asRDD[LabeledPoint](h2oFrame)` this would get executed. It's build on top of the internals but it would be nice to have this as normal conversion instead of having to call helper methods. But again, we can do JIRA for it and work on this change later so we don't block you. What do you think @mdymczyk ?

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    if (MissingValuesHandling.Skip.eq(missingHandler)) {
+      trainingRDD = trainingRDD.filter(!_.anyNull)
+    } else if (MissingValuesHandling.MeanImputation.eq(missingHandler)) {
+      // Computing the means by hand and not using frame.means() as it does not compute the mean for enum columns
+      means = movingAverage(trainingRDD, fields, domains)
+    }
+
+    (trainingRDD.map(row => {
+      val features = new Array[Double](nfeatures)
```
jakubhava: Instead of these 2 lines I like this more

   ``` val features = (0 until nfeatures).map{ i => if
        (row.isNullAt(i)) means(i) else toDouble(row.get(i), fields(i), domains(i))
      }.toArray[Double]``` It seems more functional to me since we set the features right away, but it's probably just matter of taste

core/src/test/scala/org/apache/spark/h2o/utils/SharedSparkTestContext.scala
```diff
@@ -51,26 +52,51 @@ trait SharedSparkTestContext extends SparkTestContext { self: Suite =>
     super.afterAll()
   }
 
-  def buildChunks[T: ClassTag](fname: String, data: Array[T], cidx: Integer, h2oType: Array[Byte]): Chunk = {
+  def makeH2OFrame[T: ClassTag](fname: String, colNames: Array[String], chunkLayout: Array[Long],
```
mdymczyk: will add a TODO, cool

jakubhava: I think JIRA would be better, it's easier to track than to track TODOs in the code. It's also more generic JIRA - replace all code which we use to create small frames with this builder.

mdymczyk: done https://0xdata.atlassian.net/browse/SW-250

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
mdymczyk: yes making this more robust, maybe even truly user facing, since MLlib uses LabeledPoint a lot, is a good idea - will make a JIRA for it since in that case it will need a bit more thought

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
```diff
@@ -0,0 +1,132 @@
...

+    * as the label. All other columns will be mapped to a vector and used as features. Categorical columns will be mapped to their

+    * numerical values.

+    *

+    * @param frame Input frame to be converted

+    * @param _response_column Column which contains the labels
```
mdymczyk: autgenerated with intellij since that's the convention in H2O, can change it to without - don't have any strong opinions myself

ml/src/main/scala/org/apache/spark/ml/FrameMLUtils.scala
mdymczyk: 👍

jakubhava:
Can you pls squash those 2 commits ? Also is sparkling_water_yarn branch passing ? If yes, I'm merging 👍

PR #76, opened 2016-09-07
Big reshuffle with types handling

vpatryshev:
Created an object, SupportedTypes, that is holding most of class/type data and is used for mapping.

Added ProductType and ProductMember classes, to store serializable data regarding the type of data we read in ReadConverter and elsewhere.

Also, less ad-hoc operations per row, all this is done ahead, and only accessors are called to build a tuple or a product. That unites simple product and case classes creation.
An operation of scanning indices instead of values is eliminated in several places.

ReadConverterContext now does not produce intermediate Option when reading a field; we know in advance what we need, and just call the right function. Kind of faster, less memory abuse.

All type conversions are now in ReflectionUtils.

A bunch of compiler warnings shown by IntelliJ fixed.

Added default values to SimpleType in SupportedTypes, so they can be used for extracting pure values.

Test data moved to a separate class, for common usage.

Also, added a functionality to test serializability of classes. Helped a lot.

The big problem was a bug in Scala 10.1 - see https://groups.google.com/forum/#!topic/scala-language/JNzSaqHdr1w and http://stackoverflow.com/questions/35337577/scala-reflection-with-serialization-over-spark-symbols-not-serializable


vpatryshev:
Michal, it may make sense to show this functionality/design/architecture on a big screen.

core/src/main/scala/org/apache/spark/h2o/converters/H2ODataFrame.scala
```diff
@@ -44,11 +45,12 @@ class H2ODataFrame[T <: water.fvec.Frame](@transient val frame: T,
   def this(@transient frame: T)
           (@transient sc: SparkContext) = this(frame, null)(sc)
 
+  val typesAll: Array[DataType] = frame.vecs map ReflectionUtils.dataTypeFor
 
-  val typesAll: Array[DataType] = frame.vecs().indices.map(idx => vecTypeToDataType(frame.vec(idx))).toArray
   /** Create new types list which describes expected types in a way external H2O backend can use it. This list
     * contains types in a format same for H2ODataFrame and H2ORDD */
-  val expectedTypesAll: Option[Array[Byte]] = ConverterUtils.prepareExpectedTypes(isExternalBackend, typesAll)
+  lazy val expectedTypesAll: Option[Array[Byte]] = ConverterUtils.prepareExpectedTypes(isExternalBackend, typesAll)
```
jakubhava: I'm not sure how `lazy` behaves in distributed environments. Spark `serialises` the outer scope of compute method, but not sure how it does with lazy val.

Will this work ?

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
-  ) withDefault (t => throw new scala.IllegalArgumentException(s"Type $t not supported for conversion from H2OFrame to Spark's Dataframe"))
-
-  def getNullable(dataType: DataType): (Int => Any) = (i: Int) => get(dataType)(i) orNull
-
   def numRows: Int
   def increaseRowIdx() = rowIdx += 1
+
   def hasNext = rowIdx < numRows
+
+  case class OptionReader(name: Any, apply: Int => Option[Any]) {
```
jakubhava: extra braces here

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
+  }
+
+//  type OptionReader = Int => Option[Any]
```
jakubhava: Any reason why this should be commented and not removed ?

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
+    Short -> getShort,
+    Integer -> getInt,
+    Long -> getLong,
+    Float -> getFloat,
+    Double -> getDouble,
+    Boolean -> getBoolean,
+    String -> getUTF8String,
+    UTF8 -> getUTF8String,
+    Timestamp -> getTimestamp
+  ) withDefault (t => throw new scala.IllegalArgumentException(s"Type $t not supported for conversion from H2OFrame to Spark's Dataframe"))
```
jakubhava: `ReadConverterContext` and also `InternalReadConvertorContext` were design in a first way not to depend on whether we convert data to `H2ODataFrame` or `H2ORDD`.  So maybe in this case the change of exception message would help.

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
+
+  /**
+    * For a given array of source column indexes and required data types,
+    * produces an array of value providers.
+    *
+    * @param columnIndexesWithTypes lists which columns we need, and what are the required types
+    * @return an array of value providers. Each provider gives the current column value
+    */
+  def columnValueProviders(columnIndexesWithTypes: Array[(Int, DataType)]): Array[() => Option[Any]]
 }
```
jakubhava: `ReadConverterContext` and also `InternalReadConvertorContext` were design in a first way not to depend on whether we convert data to `H2ODataFrame` or `H2ORDD`, which is not not true now because they contain specialised methods to handle Spark's `DataType`s which is only case for conversion from `H2OFrame` to Spark `DataFrame`, not for conversion to `RDD`. 

I would still like to keep them independent on this conversion, right now we split some implementations for H2ORDD and H2ODataframe in multiple classes. You could create another layer on top of ReadConverterContext which would be either for H2ORDD or H2ODataframe

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
+    String -> getUTF8String,
+    UTF8 -> getUTF8String,
+    Timestamp -> getTimestamp
+  ) withDefault (t => throw new scala.IllegalArgumentException(s"Type $t not supported for conversion from H2OFrame to Spark's Dataframe"))
+
+  def readerFor(dt: DataType): OptionReader = OptionReader(bySparkType(dt), ReaderPerType(bySparkType(dt)))
+
+  case class Reader(name: Any, apply: Int => Any)
+
+  val readerMapByName: Map[NameOfType, Reader]
```
jakubhava: This applies only to H2ORDD conversion

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
```diff
@@ -52,23 +54,43 @@ trait ReadConverterContext {
...
+
+  /**
+    * For a given array of source column indexes and required data types,
+    * produces an array of value providers.
+    *
+    * @param columnIndexesWithTypes lists which columns we need, and what are the required types
+    * @return an array of value providers. Each provider gives the current column value
+    */
+  def columnValueProviders(columnIndexesWithTypes: Array[(Int, DataType)]): Array[() => Option[Any]]
```
jakubhava: This applies only to H2ODataFrame conversion

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -37,39 +38,106 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
+    for {
+      chunk <- Option(chks(columnNum)) if !chunk.isNA(rowIdx)
+      data <- Option(read(chunk))
+    } yield data
+  }
 
-  // TODO(vlad): try to move out all this logic to prepared extractors
-  override def getString(columnNum: Int): Option[String] = get(columnNum, chunk =>
-    {
+  private def getSimple[T](defaultValue: T, read: Chunk => T)(columnNum: Int): T = {
```
jakubhava: This method doesn't seem to be used anywhere

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -37,39 +38,106 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
+  private lazy val availableReadersByName: Map[NameOfType, Reader] =
+    OptionReaders flatMap twoReaders.tupled
+
+  lazy val readerMapByName: Map[NameOfType, Reader] = availableReadersByName withDefaultValue DefaultReader
+
+//  lazy val readerMap: Map[SupportedType, Reader] = availableReaders withDefaultValue DefaultReader
+
+  def columnValueProviders(columnIndexesWithTypes: Array[(Int, DataType)]): Array[() => Option[Any]] = {
+    for {
+      (i, ft) <- columnIndexesWithTypes
```
jakubhava: `dataType` or something like that would be maybe more readable instead of `ft`. I know it's implied by the context, but still from my point of view things like this should be obvious after a first quick look :)

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -37,39 +38,106 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
+    List((s"Option[$key]":NameOfType) -> Reader(s"Option[$key]":NameOfType, op),
+      key                            -> Reader(key, (col: Int) => {
+        op.apply(col).getOrElse(Defaults(key))
+      })
+    )
+
+  private lazy val availableReadersByName: Map[NameOfType, Reader] =
+    OptionReaders flatMap twoReaders.tupled
+
+  lazy val readerMapByName: Map[NameOfType, Reader] = availableReadersByName withDefaultValue DefaultReader
```
jakubhava: If I understand it right all the `readerMapByName`, `availableReadersByName`, `twoReaders`, `Defaults` and `OptionReaders` applies only to conversion to H2ORDD, in conversion to H2ODataFrame we get the right method right in the `ReadConverterContext`. Would it be possible to separate the code for conversion to H2ODataFrame and H2ORDD ? ( please see the relevant comment bellow ) It makes it hard to read ( at least for me ) since right now we mix more things together and things are split in different files.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -37,39 +38,106 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
-    } else {
-      assert(assertion = false, s"Should never be here, type is ${vector.get_type_str()}")
-      // TODO(vlad): this is temporarily here, to provide None is string is missing
-      null
-    }
-    })
+    if (chunk.isNA(rowIdx)) defaultValue else read(chunk)
+  }
+
+  override def getLong(columnNum: Int): Option[Long] = {
```
jakubhava: I would write it just in one like, just to be consistent with the rest of the code 

`override def getLong(columnNum: Int): Option[Long] = getOption(_.at8(rowIdx))(columnNum)`

core/src/main/scala/org/apache/spark/h2o/converters/H2ODataFrame.scala
vpatryshev: I am not sure at all whether this stuff will work. Laziness can be kicked
out, it's not an issue; as soon as the thing serializes. Since there are no
tests for that, we don't know for sure how it behaves.

Thanks,
-Vlad

vpatryshev:
Great comments, thanks, will work on fixing all this.

jakubhava:
Thanks. I really like this architecture change though, it's becoming to be more scala-like, which is great. Thanks for all this!

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
vpatryshev: fixed.

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
vpatryshev: fixed. thanks!

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
vpatryshev: I see. I have changed the wording of the exception, but regarding using DataType, it's only being referenced from H2ODataFrame, not from H2ORDD. 
I support, though, the idea of not mentioning DataType at all in ReadConvertorContext; let me see if I can move it out to H2ODataFrame.

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
vpatryshev: Yes, it was. Now reorganized, that last part was missing. Now it's shared, since it's the same functionality.

core/src/main/scala/org/apache/spark/h2o/converters/ReadConverterContext.scala
vpatryshev: Yes; moved out DataTypes out of it.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: Yep. Hope there's no such thing here anymore.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: Oh, thanks a lot. It works now.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: Actually, most of them are being used in both use cases. It was a little bit misleading, but now it's all essential.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: My preference too.

vpatryshev:
Just committed the updates. Pls take a look. Thanks for your comments!

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -17,59 +17,127 @@
...
+    Vec.T_CAT -> categoricalString,
+    Vec.T_UUID -> uuidString,
+    Vec.T_STR -> plainString
+  ) withDefault((t:Byte) => {
+    assert(assertion = false, s"Should never be here, type is $t")
+    (_:Chunk) => null
+  }
+    )
+
+  def string(chunk: Chunk): String = StringProviders(chunk.vec().get_type())(chunk)
```
jakubhava: Maybe this method and `StringProviders` could be private too ?

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -17,59 +17,127 @@
...
+  } toMap
+
+  def columnValueProviders(columnIndexesWithTypes: Array[(Int, SimpleType[_])]): Array[() => Option[Any]] = {
+    for {
+      (columIndex, supportedType) <- columnIndexesWithTypes
+      reader = OptionReaders(byBaseType(supportedType))
+      provider = () => reader.apply(columIndex)
+    } yield provider
+  }
 }
```
jakubhava: When external backend will be merged on one sunny day, I could actually reuse most of the code in the `InternalReadConverterContext` in the `ExternalReadConverterContext` to  be consistent. It almost seems like all the code can be reused but corresponding `ExtractorsTable` will be implemented there with relevant extractors.

What about putting these generic methods right to `ReadConvertorContext` ? The only specific code to `InternalReadConvertorContext` compared to general `ReadConverterContext` is `ExtractorMap`. We could do it either as part of this commit or I can do it when merging the future external backend. It's not a big deal.

I'm just interested in your opinion and if you would have anything against this step. Thanks!

jakubhava:
I added a few last comments, otherwise looks good to me. 👍

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -17,59 +17,127 @@
...
+      case (t, getter) => t -> returnSimple(t.defaultValue, getter) _
+    } toMap
+
+  private lazy val SimpleReaders: Map[SimpleType[_], Reader] = SimpleReadersMap withDefault
+    (t => throw new scala.IllegalArgumentException(s"Type $t conversion is not supported in Sparkling Water"))
+
+  private def plainReaderFor(supportedType: SimpleType[_], optReader: OptionReader): Reader =
+    (col: Int) => optReader(col).getOrElse(supportedType.defaultValue)
+
+  lazy val readerMapByName: Map[NameOfType, Reader] = (OptionReaders ++ SimpleReaders) map {
```
jakubhava: This is just question, not an improvement request :) I see that in `ReadConverterContext` you defined 
readerMapByName using `def` as `def readersMapByDefined: Map[NameOfType, Reader]` and here you implemented it using `val`. I was searching for this but I could find any explanation, can I just ask you what does it internally do and what is it good for when you implement something defined using `def` with `val` ?

Thanks a lot for answer!. Sorry for putting it here as another comment, but I thought it will be easier to raise my question close to real code example.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: Sure! Fixing.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: Well, now (at last) I see what "Internal" means here. :) Yes, we can move it over there, or, I would suggest, to a separate trait that both Internal and External extend. I could do it right now, but it's not exactly clear to me which parts are worth moving. All except ExtractorMap? If so, that would be totally gorgeous. OTOH, I would not rush with this; as soon as External starts forming, it will be much easier to figure out which parts are actually worth sharing.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
vpatryshev: See, `def` in a trait without implementation is just a general idea of something being provided in implementations. The implementation may be def, or val, or lazy val. Def will reevaluate on every reference, which makes no sense. So either val or lazy val. I made implementation lazy val for the reason of avoiding bad initialization order. Scala people blame JVM, but I would blame scala compiler. It happens pretty frequently that a member of a class refers another member, and if that member is not initialized yet (just because it's below, textually), it is only discovered in runtime, and it is null. Not sure if there is a bug filed, but well, our situation is special, we are targeting a range of compilers, as I understand.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
jakubhava: Cool, that's great! Let's do it when "External" backend will come into game, as you said, it will be probably easier to determine which parts will be shared.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
jakubhava: Thanks for great explanation 👍

jakubhava:
@vpatryshev I created new branch based on this one because I needed to do some minor changes there - cleaned up tests a bit and rebased on master ( spark 2.0).  That branch based on yours was merged few minutes ago. That's why I'm closing this PR. Finally, the change is in master 👍

vpatryshev:
Well, this sounds great to me. Let's see when the stuff is merged; probably will need another round of upgrades, or maybe not.

jakubhava:
It's merged already. I'm now changing my PR with preparation for external backend so I can merge that too so we are again a bit closer to merging it completely

PR #82, opened 2016-09-22
Vp glm test in scala

vpatryshev:
Copied, converted to Scala, cleaned up GLMtest from H2O.


mmalohlava:
Why the code is in Sparkling Water? It should be in h2o-3/h2o-scala - what do you think?

Or should we put h2o-scala project into sparkling water?

vpatryshev:
Ok, sure it's not related to Spark; but all the code in h2o-3 is all Java;
I have not seen any h2o-scala; does it exist?

Thanks,
-Vlad

jakubhava:
yup, it exists - https://github.com/h2oai/h2o-3/tree/master/h2o-scala

PR #83, opened 2016-09-25
[SW-206] Support of Spark 2.0

mmalohlava:


core/src/main/scala/org/apache/spark/h2o/utils/LogUtil.scala
```diff
@@ -45,9 +50,9 @@ object LogUtil extends org.apache.spark.Logging {
       logWarn()
     } else {
       LogManager.getLogger("water.default").setLevel(Level.toLevel(level))
-      Log._level = levelIdx
+      // FIXME: Log._level = levelIdx
```
mdymczyk: Why can't this class be in `water.util`? This way we'll need a minor change in h2o.

core/src/main/scala/org/apache/spark/listeners/H2OSparkListener.scala
```diff
@@ -16,13 +16,9 @@
 */
 package org.apache.spark.listeners
 
-import org.apache.spark.scheduler.{SparkListener, SparkListenerEvent, SparkListenerExecutorAdded}
+import org.apache.spark.scheduler.{SparkListener, SparkListenerExecutorAdded}
 
 trait H2OSparkListener extends SparkListener {
```
mdymczyk: Guess we can just remove this class altogether now?

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -89,4 +91,14 @@ class H2OConf(@transient val sc: SparkContext) extends Logging with InternalBack
 
 object H2OConf {
   def apply(sc: SparkContext) = new H2OConf(sc)
+
+  private var _sparkConfChecked = false
+
+  def sparkConfChecked = _sparkConfChecked
+
+  def checkSparkConf(sparkConf: SparkConf): SparkConf = {
```
mdymczyk: Maybe a different name, since this doesn't really check the conf for anything?

mdymczyk:
Besides the small comments it looks ok to me. Maybe @jakubhava can find something especially in the interpreter and python ;-)

examples/src/integTest/scala/water/sparkling/itest/IntegTestHelper.scala
```diff
@@ -115,14 +120,14 @@ object LocalTest extends Tag("water.sparkling.itest.LocalTest")
...
-trait IntegTestStopper extends org.apache.spark.Logging {
+trait IntegTestStopper {
 
   def exitOnException(f: => Unit): Unit ={
     try {
       f
     } catch {
       case t: Throwable => {
-        logError("Test throws exception!", t)
+        //logError("Test throws exception!", t)
```
jakubhava: Why we don't log here ? Is the `printStactTrace` enough ?

core/src/main/scala/org/apache/spark/h2o/converters/SupportedDataset.scala
```diff
@@ -41,15 +41,15 @@ object SupportedDataset {
     val typeTag = ttag
     override def toH2OFrame(sc: SQLContext, frameKeyName: Option[String]): H2OFrame = {
       val tpe = ttag.tpe
-      val constructorSymbol = tpe.declaration(nme.CONSTRUCTOR)
+      val constructorSymbol = tpe.declaration(nme.CONSTRUCTOR) // To be compatible with Scala 2.10
```
jakubhava: Just a question, why is here the comment // To be compatible with Scala 2.10 ? Would this be done in different way in scala 2.11 ? 

I have no problem with this, I'm just don't fully understand this short piece of code.

Thanks!

jakubhava:
Some minor comments, otherwise looks good to me 👍

examples/src/main/scala/org/apache/spark/examples/h2o/AirlinesWithWeatherDemo.scala
```diff
@@ -35,6 +35,7 @@ object AirlinesWithWeatherDemo extends SparkContextSupport {
     // Configure this application
     val conf: SparkConf = configure("Sparkling Water: Join of Airlines with Weather Data")
 
+    FIXME replace with spark Session
```
mmalohlava: Thx, Kuba, i preserve separation of individual steps - you have to look at final diff.

jakubhava: Oh right, sorry. I went commit by commit.

mmalohlava: I will probably merge it as it is without squashing. WDYT?

core/src/main/scala/org/apache/spark/h2o/utils/LogUtil.scala
mmalohlava: Because of Spark: `private[spark] trait Logging {`

core/src/main/scala/org/apache/spark/listeners/H2OSparkListener.scala
mmalohlava: Let's keep it for now.

examples/src/main/scala/org/apache/spark/examples/h2o/AirlinesWithWeatherDemo.scala
jakubhava: jup, I would keep the commits separated like this

mmalohlava: oki, fixed

examples/src/integTest/scala/water/sparkling/itest/IntegTestHelper.scala
mmalohlava: because Logging class of Spark is private

core/src/main/scala/org/apache/spark/h2o/converters/SupportedDataset.scala
mmalohlava: Because we need to compile for Spark 2.10 and 2.11. I tried to switch to new Spark 2.11 API but it is not compatible with 2.10

PR #85, opened 2016-09-26
Prepare the code for external backend core implementation

jakubhava:
This PR introduces introduces this changes:
- Several minor fixes
- Remove duplicated code used by `ProductRDDConverter` and `DataSetConverter`
- Use new converter api in `DatasetConverter`
- unify the `DatasetConverter` with the rest of converters
- add implicit conversion for datasets
- fix problem of getting wrong configuration in `H2ODataFrame` and `H2ORDD`
- move creation of `ConvertorContext` into `H2ORDDLike`
- Send whole list of expected types and list of selectedColumns in both `H2ODataFrame` and `H2ORDD` to h2o backend

Some fixes affect both backends and some only external backend. 

This needs to be merged before the PR with external backend core implementation is merged.


core/src/main/scala/water/api/RDDs/RDDsHandler.scala
```diff
@@ -75,7 +75,7 @@ class RDDsHandler(val sc: SparkContext, val h2oContext: H2OContext) extends Hand
       throw new H2ONotFoundArgumentException(s"RDD with ID '${s.rdd_id}' does not exist, can not proceed with the transformation!")
     }
     // TODO(vlad): take care of the cases when the data are missing
-    val rdd = sc.getPersistentRDDs.get(s.rdd_id).get
+    val rdd = sc.getPersistentRDDs(s.rdd_id)
     val h2oFrame = if(s.h2oframe_id == null) convertToH2OFrame(rdd, None) else convertToH2OFrame(rdd,Some(s.h2oframe_id.toLowerCase))
```
vpatryshev: this can _probably_ be rewrtitten as
val h2oFrame = convertToH2OFrame(rdd, Option(s.h2oframe_id) map (_.toLowerCase))

vpatryshev:
Wrote several comments, nothing essential; the whole pr is very good, I think. But we need tests.

jakubhava:
Provided fixes based on @vpatryshev suggestions. I'm also happy to write tests for it, I just don't see what kind of tests I should create for this PR. Maybe, can you point me to some @vpatryshev please so I have an idea ? Thanks!

vpatryshev:
I made a couple of notes. Just two classes actually: DatasetConverter and
H2OFrameFromRDDProductBuilder.

Thanks,
-Vlad

jakubhava:
Made up-to-date with current master

PR #87, opened 2016-09-27
External cluster implementation

jakubhava:
- implements core external cluster functionality
- add support for testing external cluster locally with existing tests
- add support to start h2o instances on yarn from sparkling-water

Test can be run both locally and on yarn. Locally means that h2o cluster is started on the same machine and on yarn means the the cluster is started on yarn. In both cases we  need to set H2O_EXTENDED_JAR environmental property to path to either extended h2o jar or h2o driver jar.

To run tests on external backend locally:
```export H2O_EXTENDED_JAR=`./gradlew -q extendJar -PdownloadH2O`
 ./gradlew build -x check -PbackendMod=external```

To run tests on external backend on yarn:
```export H2O_EXTENDED_JAR=`./gradlew -q extendJar -PdownloadH2O=hadoop_version`
 ./gradlew build -x check -PbackendMode=external -PstartH2OClusterOnYarn```

### CODE REVIEW

**Comment by mdymczyk** [comment]
@jakubhava I'm a bit behind with external cluster thing status but wanted to label this PR, does it need a review? some other blockers?

**Comment by jakubhava** [comment]
Let's say it's WIP - I'm finalising the branch on h2o side now. Once it's done, I'll finish this one and let you guys know when it's ready for proper review

**Comment by jakubhava** [comment]
The last parts I’m about to change will be in `ExternalH2OBackend`:
- remove`-disown` argument when starting h2o cluster on yarn from sparkling water to ensure this cluster is killed when spark gets killed 
- use cluster file info instead of parsing console output of h2odriver to get yarn app id and cluster’s ip and port.

Otherwise ready to be reviewed

File: core/src/integTest/scala/water/sparkling/itest/local/H2OContextLocalClusterSuite.scala [hunk 1]
```diff
@@ -37,9 +37,20 @@ class H2OContextLocalClusterSuite extends FunSuite
...
     val conf = defaultSparkConf.setJars(swassembly :: Nil)
     sc = new SparkContext("local-cluster[3,2,2048]", "test-local-cluster", conf)
-    hc = H2OContext.getOrCreate(sc, new H2OConf(sc))
+
+    // start h2o cloud in case of external cluster mode
+    if(testsInExternalMode(sc.getConf)){
+      startCloud(3, sc.getConf)
+    }
+
+    hc = H2OContext.getOrCreate(sc, new H2OConf(sc).setNumOfExternalH2ONodes(3))
```
vpatryshev: This code is more or less repeated in H2OContextLocalSuite, differs only in a constant; makes sense to write a common function.

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
```diff
@@ -0,0 +1,100 @@
...
+  // if all connections are currently used or reuse the existing free one. Programmer needs to put the connection back to the
+  // pool of available connections using the method putAvailableConnection
+  private class PerOneNodeConnection(val nodeDesc: NodeDesc) {
+
+    private def getConnection(nodeDesc: NodeDesc): ByteChannel= {
+      ExternalFrameUtils.getConnection(nodeDesc.hostname, nodeDesc.port)
+    }
+    // ordered list of connections where the available connections are at the start of the list and the used at the end.
+    private val availableConnections = new java.util.concurrent.ConcurrentLinkedQueue[(ByteChannel, Long)]()
+    def getAvailableConnection(): ByteChannel = {
```
vpatryshev: Is it a getter? Does it have a side effect? If not, Scala naming convention suggests not to use parentheses.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -0,0 +1,138 @@
...
+    self
+  }
+
+  def h2oClusterHost = {
+    val value = sparkConf.getOption(PROP_EXTERNAL_CLUSTER_REPRESENTATIVE._1)
+    if (value.isDefined) {
+      Option(value.get.split(":")(0))
+    } else {
+      None
+    }
```
vpatryshev: This is not Scala style. The right Scala way is a one-liner:

sparkConf.getOption(PROP_EXTERNAL_CLUSTER_REPRESENTATIVE._1) map (_.split(":")(0))

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  def h2oClusterPort = {
+    val value = sparkConf.getOption(PROP_EXTERNAL_CLUSTER_REPRESENTATIVE._1)
+    if (value.isDefined) {
+      Option(value.get.split(":")(1))
+    } else {
+      None
+    }
+  }
```
vpatryshev: Please use Scala style.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -0,0 +1,138 @@
...
+        |  clientBasePort       : ${clientBasePort}
+        |  h2oClientLog         : ${h2oClientLogLevel}
+        |  nthreads             : ${nthreads}""".stripMargin
+}
+
+object ExternalBackendConf {
+
+  val PROP_EXTERNAL_H2O_MEMORY = ("spark.ext.h2o.hadoop.memory", "6g")
+
+  val PROP_EXTERNAL_CLUSTER_HDFS_DIR = ("spark.ext.h2o.external.hdfs.dir", null.asInstanceOf[String])
```
vpatryshev: Can you not use nulls in Scala? It's not a good style. Seems like the value does not matter here, right?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -0,0 +1,138 @@
...
+  val PROP_EXTERNAL_CLUSTER_H2O_DRIVER = ("spark.ext.h2o.external.h2o.driver", null.asInstanceOf[String])
+
+  /** Yarn queue on which external cluster should be started */
+  val PROP_EXTERNAL_CLUSTER_YARN_QUEUE = ("spark.ext.h2o.external.yarn.queue", null.asInstanceOf[String])
+
+  /** ip:port of arbitrary h2o node to identify external h2o cluster */
+  val PROP_EXTERNAL_CLUSTER_REPRESENTATIVE = ("spark.ext.h2o.cloud.representative", null.asInstanceOf[String])
+
+  /** Number of nodes to wait for when connecting to external H2O cluster */
+  val PROP_EXTERNAL_H2O_NODES = ("spark.ext.h2o.external.cluster.num.h2o.nodes", null.asInstanceOf[String])
```
vpatryshev: All these nulls... you should not use them in Scala. Same with 'asInstanceOf' - another case of bad Scala.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  def cloudMembers = H2O.CLOUD.members().map(NodeDesc(_))
+
+  private[this] def getH2OClientConnectionArgs(conf: H2OConf): Array[String] = {
+    if (conf.h2oCluster.isDefined) {
+      Array("-flatfile", saveAsFile(conf.h2oCluster.get).getAbsolutePath)
+    }else{
+      Array()
+    }
```
vpatryshev: Not like this. It's an option. Treat it properly. E.g.
conf.h2oCluster map (cluster => Array("-flatfile", saveAsFile(cluster).getAbsolutePath))) getOrElse Array()

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -0,0 +1,170 @@
...
+
+class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with ExternalBackendUtils with Logging {
+
+  private var yarnAppId: String = _
+  def launchH2OOnYarn(conf: H2OConf): (String, String) = {
+    val cmdToLaunch = Seq[String](
+      "hadoop",
+      "jar",
+      conf.h2oDriverPath.get,
+      conf.YARNQueue.map("-Dmapreduce.job.queuename=" + _ ).getOrElse(""),
```
vpatryshev: :) yep.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -0,0 +1,170 @@
...
+      "-J", "-md5skip",
+      "-mapperXmx", conf.mapperXmx,
+      "-output", conf.HDFSOutputDir.get,
+      "-disown" // close the driver after cluster formation
+    )
+
+    logDebug("Command used to start H2O on yarn: " + cmdToLaunch.mkString)
+
+    import scala.sys.process._
+    val stdout = new StringBuffer()
```
vpatryshev: This is nice but misleading. Could you give a less misguiding name?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    logDebug(stdout.toString)
+    logDebug(stderr.toString)
+
+    // parse required information
+    val clusterName = stdout.toString.split("\n").find(_.contains("Job name")).get.split("'")(1)
+    val arbitraryNodeIPPort = stdout.toString.split("\n").find(_.contains("H2O node")).get.split(" ")(2)
+    val split = stdout.toString.split("\n").find(_.contains("For YARN users, logs command is")).get.split(Array(' ','''))
+    yarnAppId = split(split.length-1)
+    assert(proc == 0, s"Process ended with return value $proc")
```
vpatryshev: If you say what exactly process ended badly, users will be less puzzled by this message.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -0,0 +1,170 @@
...
+    if(hc.getConf.HDFSOutputDir.isDefined){
+      try {
+        val hdfs = org.apache.hadoop.fs.FileSystem.get(hc.sparkContext.hadoopConfiguration)
+        hdfs.delete(new Path(hc.getConf.HDFSOutputDir.get), true)
+      }catch {
+        case e: Exception => log.error(e.getMessage)
+      }
+      import scala.sys.process._
+      // kill the job
+      Seq[String]("yarn", "application", "-kill", yarnAppId).mkString(" ").!
```
vpatryshev: s"yarn application -kill $yarnAppId" ! would look more readable

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalReadConverterContext.scala
```diff
@@ -0,0 +1,68 @@
...
+                                    val nodeDesc: NodeDesc, expectedTypes: Array[Byte], selectedColumnIndices: Array[Int])
+  extends ExternalBackendUtils with ReadConverterContext {
+  override type DataSource = ExternalFrameReaderClient
+
+  private val socketChannel = ConnectionToH2OHelper.getOrCreateConnection(nodeDesc)
+  val externalFrameReader = new ExternalFrameReaderClient(socketChannel, keyName, chunkIdx, selectedColumnIndices, expectedTypes)
+
+  private val numOfRows: Int =  externalFrameReader.getNumRows
+
+  override def numRows: Int = numOfRows
```
vpatryshev: You can use lazy val numRows = externalFrameReader.getNumRows, and all your caching is done.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalReadConverterContext.scala
```diff
@@ -0,0 +1,68 @@
...
+
+  override def returnOption[T](read: DataSource => T)(columnNum: Int): Option[T] = {
+    val value = read(externalFrameReader)
+    if(externalFrameReader.isLastNA){
+      None
+    }else{
+      Option(value)
+    }
```
vpatryshev: Totally bad name. 
Also, you can as well rewrite it as Option(read(externalFrameReader)) filter (!_.isLastNA)

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalReadConverterContext.scala
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  override def returnSimple[T](ifMissing: String => T, read: DataSource => T)(columnNum: Int): T = {
```
vpatryshev: Bad name too, I think.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -36,99 +36,37 @@ class InternalReadConverterContext(override val keyName: String, override val ch
 
   /** Number of rows in this partition */
   lazy val numRows = chks(0)._len
-  
-  private def returnOption[T](read: Chunk => T)(columnNum: Int): Option[T] = {
+
+  private def underlyingFrame = DKV.get(Key.make(keyName)).get.asInstanceOf[Frame]
+
+  override def returnOption[T](read: DataSource => T)(columnNum: Int): Option[T] = {
```
vpatryshev: Weird, weird name.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalReadConverterContext.scala
```diff
@@ -36,99 +36,37 @@ class InternalReadConverterContext(override val keyName: String, override val ch
...
     for {
       chunk <- Option(chks(columnNum)) if !chunk.isNA(rowIdx)
       data <- Option(read(chunk))
     } yield data
   }
 
-  private def returnSimple[T](ifMissing: String => T, read: Chunk => T)(columnNum: Int): T = {
+  override def returnSimple[T](ifMissing: String => T, read: DataSource => T)(columnNum: Int): T = {
```
vpatryshev: Weird name...

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
```diff
@@ -119,8 +128,9 @@ object ConverterUtils extends ConverterUtils {
 
   def getReadConverterContext(keyName: String, chunkIdx: Int,
                               extra: Option[ExternalBackendInfo]): ReadConverterContext = {
-    val converterContext = if (extra.isDefined) { // metainfo external cluster is not empty => use external cluster
-      throw new NotImplementedError("Not implemented at the moment")
+    val converterContext = if (extra.isDefined) {
+      // metainfo external cluster is not empty => use external cluster
+      new ExternalReadConverterContext(keyName, chunkIdx, extra.get.chksLocation(chunkIdx), extra.get.expectedTypes, extra.get.selectedColumnIndices)
```
vpatryshev: Pls use Scala style

core/src/main/scala/org/apache/spark/h2o/converters/ConverterUtils.scala
```diff
@@ -149,27 +159,18 @@ object ConverterUtils extends ConverterUtils {
...
 
 object ExternalBackendInfo{
-  def apply(chksLocation: Option[Array[NodeDesc]],
-            expectedTypes: Option[Array[Byte]],
+  def apply(chksLocation: Option[Array[NodeDesc]], expectedTypes: Option[Array[Byte]],
             selectedColumnIndices: Array[Int]): Option[ExternalBackendInfo] = {
 
     if(chksLocation.isDefined){
-      Some(new ExternalBackendInfo(chksLocation.get, expectedTypes.get, selectedColumnIndices))
+      Some(new ExternalBackendInfo(chksLocation.get, expectedTypes.get,  selectedColumnIndices))
```
vpatryshev: Pls use scala style

core/src/main/scala/org/apache/spark/h2o/converters/LabeledPointConverter.scala
```diff
@@ -46,9 +45,17 @@ private[converters] object LabeledPointConverter extends Logging with ConverterU
...
-    val vecTypes = Array.fill(maxNumFeatures + 1)(Vec.T_NUM)
 
-    convert[LabeledPoint](hc, rdd, keyName, fnames, vecTypes, perLabeledPointRDDPartition(maxNumFeatures))
+    // in case of internal backend, store regular vector types
+    // otherwise for external backend store expected types
+    val expectedTypes = if(hc.getConf.runsInInternalClusterMode){
+      Array.fill(maxNumFeatures + 1)(Vec.T_NUM)
+    }else{
+      Array.fill(maxNumFeatures + 1)(ExternalFrameUtils.EXPECTED_DOUBLE)
+    }
```
vpatryshev: Code duplication; why not do it like this:
val expectedType = if (hc...) Vec.T_NUM else External...EXPECTED_DOUBLE
and then val expectedTypes = Array.fill(maxNumFeatures+1)(expectedType)

core/src/main/scala/org/apache/spark/h2o/converters/LabeledPointConverter.scala
```diff
@@ -65,12 +72,13 @@ private[converters] object LabeledPointConverter extends Logging with ConverterU
   def perLabeledPointRDDPartition(maxNumFeatures: Int)
                                  (keyName: String, vecTypes: Array[Byte], uploadPlan: Option[immutable.Map[Int, NodeDesc]])
                                  (context: TaskContext, it: Iterator[LabeledPoint]): (Int, Long) = {
+    val asArr = it.toArray[LabeledPoint] // need to buffer the iterator in order to get its length
```
vpatryshev: Pretty bad name; you just specify the data type here, not the meaning of it.

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
```diff
@@ -57,25 +65,29 @@ private[converters] object PrimitiveRDDConverter extends Logging with ConverterU
   def perPrimitiveRDDPartition[T]() // extra arguments for this transformation
                                  (keyName: String, vecTypes: Array[Byte], uploadPlan: Option[immutable.Map[Int, NodeDesc]]) // general arguments
                                  (context: TaskContext, it: Iterator[T]): (Int, Long) = { // arguments and return types needed for spark's runJob input
+
+    val asArr = it.toArray[Any] // need to buffer the iterator in order to get its length
```
vpatryshev: Again, bad name.

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
```diff
@@ -57,25 +65,29 @@ private[converters] object PrimitiveRDDConverter extends Logging with ConverterU
...
+      case n: Char => con.put(0, n)
+      case n: Short => con.put(0, n)
+      case n: Int => con.put(0, n)
+      case n: Long => con.put(0, n)
+      case n: Float => con.put(0, n)
+      case n: Double => con.put(0, n)
+      case n: String => con.put(0, n)
+      case n: java.sql.Timestamp => con.put(0, n)
+      case _ => con.putNA(0)
     }
```
vpatryshev: This chunk of code was already somewhere above.
Why not extract a method/function for it?

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -78,12 +86,12 @@ trait SparkDataFrameConverter extends Logging with ConverterUtils {
   def perSQLPartition(types: Seq[(Seq[Int], StructField, Byte)])
                      (keyName: String, vecTypes: Array[Byte], uploadPlan: Option[immutable.Map[Int, NodeDesc]])
                      (context: TaskContext, it: Iterator[Row]): (Int, Long) = {
+    val asArr = it.toArray[Row] // need to buffer the iterator in order to get its length
```
vpatryshev: bad name; also, why do you want arrays all around? a list would be enough, I think
(and I wonder why Iterators are being passed all around; a Seq would be way better)

core/src/main/scala/org/apache/spark/h2o/utils/ReflectionUtils.scala
```diff
@@ -173,7 +176,7 @@ case class ProductMember(name: String, typeName: NameOfType) {
   override def toString = s"$name: $typeName"
 }
 
-case class ProductType(members: Array[ProductMember]) {
+case class ProductType(members: Array[ProductMember], types: Array[Class[_]]) {
```
vpatryshev: Why an additional argument? Let's make these "types" (actually, classes) members of ProductMember. That's what that class is for.

core/src/test/scala/org/apache/spark/h2o/utils/ExternalClusterModeTestHelper.scala
```diff
@@ -0,0 +1,84 @@
...
+  private def launchSingle(cloudName: String, ip: String): Process = {
+    val cmdToLaunch = Seq[String]("java", "-jar", h2oJar, "-md5skip", "-name", cloudName, "-ip", ip)
+    import scala.sys.process._
+    Process(cmdToLaunch).run()
+  }
+
+  def startCloud(cloudSize: Int, cloudName: String, ip: String): Unit = {
+    nodeProcesses = (1 to cloudSize).map { _ => launchSingle(cloudName, ip) }
+    // Wait 2 seconds to ensure that h2o nodes are created earlier than h2o client
+    Thread.sleep(2000)
```
jakubhava: This method does something different - it starts h2o using cli

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
jakubhava: I would like to use null for the sake of consistency with rest of the configuration properties at least as part of this change. I'm open up to eliminating `null`s as part of another commit targeting particularly this change.

core/src/integTest/scala/water/sparkling/itest/local/H2OContextLocalClusterSuite.scala
jakubhava: Yup, will have a look on it and will try to generalise  it

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
jakubhava: see comment above. Agreed, but it should be targeted to another commit since there are also cases like that in the previous sparkling water code

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: yup

core/src/main/scala/org/apache/spark/h2o/converters/LabeledPointConverter.scala
jakubhava: Yup, it make sense in this case. I was putting this piece of code amongst all the converters, but at this particular one the simplification make sense 👍

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
jakubhava: I was thinking about it. Will try to generalise

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
jakubhava: Iterator are passed because it's Spark api we have to play with. I need to buffer the array in advance because I need to get the number of elements before the iteration over them which is not possible with iterator. 

Agree that array is not necessary, can change to `List`

jakubhava:
Thanks @vpatryshev for the good comments, will work on it tomorrow

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
```diff
@@ -0,0 +1,100 @@
...
+      availableConnections.add((sock, System.currentTimeMillis()))
+
+      // after each put start this cleaning thread
+      val waitTime = 1000 // 1 second
+      new Thread(){
+        override def run(): Unit = {
+          Thread.sleep(waitTime)
+          availableConnections.asScala.filter{
+          case (socket, lastTimeUsed) =>  if (System.currentTimeMillis() - lastTimeUsed > waitTime) {
+            socket.close()
```
mdymczyk: It's been a while since I used `ConcurrentLinkedQueue` so will ask just in case, is a scenario like this possible:

1) We have 1 connection in the queue
2) The `Thread.sleep(waitTime)` wakes up and the iterator grabs the element (through `peek` i would assume)
3) A context switch occurs before the `if(System.currentTimeMillis() - lastTimeUsed > waitTime)` and `getAvailableConnection` is called, it pops the channel and returns it
4) We get switched back to the original thread, it sees that `System.currentTimeMillis() - lastTimeUsed > waitTime` is true and closes the socket, which we returned in `getAvailableConnection`.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterContext.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  val socketChannel = ConnectionToH2OHelper.getOrCreateConnection(nodeDesc)
+  val externalFrameWriter = new ExternalFrameWriterClient(socketChannel)
+
+  /**
+    * This method closes the communication after the chunks have been closed
+    */
+  override def closeChunks(): Unit = {
+    externalFrameWriter.waitUntilAllWriten()
+    ConnectionToH2OHelper.putAvailableConnection(nodeDesc, socketChannel)
```
mdymczyk: What if the above line (`waitUntilAllWriten` - btw typo in `Writen`) fails with an exception? We won't release the connection?

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
```diff
@@ -54,22 +55,30 @@ class H2ORDD[A <: Product: TypeTag: ClassTag, T <: Frame] private(@(transient @p
...
+      // prepare expected types for selected columns
+      Option(ExternalFrameUtils.prepareExpectedTypes(productType.types))
+    } else {
+      None
+    }
+  }
 
   // Check that H2OFrame & given Scala type are compatible
-  if (!productType.isSingleton) {
+  if (!productType.isSingleton) { // Unnecessary check
```
mdymczyk: If it's unnecessary then lets remove it?

core/src/main/scala/org/apache/spark/h2o/converters/H2OFrameFromRDDProductBuilder.scala
```diff
@@ -101,20 +101,28 @@ object H2OFrameFromRDDProductBuilder{
   private[converters] def perTypedDataPartition[T<:Product]()
                                                            (keyName: String, vecTypes: Array[Byte], uploadPlan: Option[immutable.Map[Int, NodeDesc]])
                                                            (context: TaskContext, it: Iterator[T]): (Int, Long) = {
+    val asArr = it.toArray[Product] // need to buffer the iterator in order to get its length
```
mdymczyk: All `it.toArray` should only be done in external mode I think, no reason to waste memory in internal mode where the length isn't used, right?

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
```diff
@@ -0,0 +1,100 @@
...
+
+    def putAvailableConnection(sock: ByteChannel): Unit = {
+      availableConnections.add((sock, System.currentTimeMillis()))
+
+      // after each put start this cleaning thread
+      val waitTime = 1000 // 1 second
+      new Thread(){
+        override def run(): Unit = {
+          Thread.sleep(waitTime)
+          availableConnections.asScala.filter{
```
jakubhava: It is a filter since we are filtering the connections and removing the old/closed ones so they are not hanging there. I think `filter` here makes semantically sense. If I use map I would have to store somewhere items to be deleted and delete them in another step

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterContext.scala
jakubhava: If that happens, connection is not released but will be removed by cleaning thread in `ConnectionHelper`. Do you think it's sufficient or have any other ideas ? 

btw: Thanks for typo!

core/src/main/scala/org/apache/spark/h2o/converters/H2OFrameFromRDDProductBuilder.scala
jakubhava: yes! thanks

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
mdymczyk: @jakubhava But we're not removing anything here as filter doesn't change the underlying collection and you're not assigning its result anywhere. It does require you to return, unnecessarily, a `Boolean`. Running:

```
availableConnections.asScala.foreach { case (socket, lastTimeUsed) =>  
    if (System.currentTimeMillis() - lastTimeUsed > waitTime) socket.close()
}```

would give you the same result but in a more concise manner.

The closed connections will get removed from the queue in `getAvailableConnection()` when you do `poll` and check for `isOpen` but if I understand the code correctly that is also a bug as `ConcurrentLinkedQueue` is a FIFO so you can have such a scenario:

1) insert a channel via `putAvailableConnection`
2) wait 500ms, insert another channel via `putAvailableConnection`
3) wait 501ms, the thread from 1) kicks in and closes channel from 1)
4) someone calls `getAvailableConnection`, the queue is not empty, `poll` grabs the channel from 1) which is already closed and decides to call `getConnection` to get a new connection even though there's a good, open connection from 2)

Comment (jakubhava) [inline core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala]: I wanted to remove the closed connections here and also in the cleaning thread -> they are not same. So this is different problem where I just forgot to return new collection. I can keep all the cleaning in the cleaning thread, but since I have to do this loop anyway it fits here well too.

Thanks for the input

Comment (mdymczyk) [inline core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala]: What do you want to do with the returned collection, though? Simply changing `availableConnections` to a var and reassigning it will cause synchronization problems.

Comment (jakubhava) [inline core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala]: good point, that's not way to go. I'll have a look on it later and will try to improve it

File: core/src/main/java/org/apache/spark/h2o/JavaH2OContext.java [hunk 31]
```diff
@@ -45,16 +48,16 @@
  * can be set in H2O configuration class H2OConf
  *
  */
-public class JavaH2OContext {
+public class JavaH2OContext implements Serializable {
```
mmalohlava: This is dangerous! Why JavaH2OContext/H2OContext needs to be `Serializable`? I was looking at `SparkContext` it is not serializable (it is by design since you do not want to pass it around cluster).

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mmalohlava: Please, another commit to change this one. Plus it is not critical problem right now.

core/src/test/scala/org/apache/spark/h2o/utils/ExternalClusterModeTestHelper.scala
mmalohlava: btw: 2sec for big clouds/busy machines is not enough. Probably for tests is ok, but still it can be dangerous

gradle/scala.gradle
```diff
@@ -63,6 +63,32 @@ sourceSets {
     }
 }
 
+//
+// Initial task checking setup of all properties required
+// by scala tests
+//
+task checkScalaTestEnv << {
```
mmalohlava: Do not use `<<` - use:
```
task checkScalaTestEnv {
  doLast {
  }
}
```

core/src/test/scala/org/apache/spark/h2o/utils/ExternalClusterModeTestHelper.scala
jakubhava: ok, thanks. Will increase that

core/src/main/java/org/apache/spark/h2o/JavaH2OContext.java
jakubhava: It was changed just recently when I was debugging one of the problems and it complained that `JavaH2OContext` is not serializable. We already have `H2OContext` implementing serialisable but having `sparkContext` as `@transient` field and that's why I decided to make this class also implement `Serializable` too even though I thought it's not a good idea.

I'm absolutely up for it having it non-serializable. It makes sense - does it mean that `H2OContext` should be changes as well ?

Thanks for pointing this out!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalReadConverterContext.scala
jakubhava: I put comment in the code explaining why it can't be lazy

jakubhava:
@vpatryshev for this feature ( communication with external H2O cluster ) we are reusing all existing tests. There is commit in this branch allowing us to run all the tests on the external H2O cluster. Also the ExternalFrame API is already tested on H2O side.

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: @vpatryshev It's not related to this PR so I will remove this comment here. But do we need this check here in the future ? The reason I could see why it's needed is that in case when `Product` `RDD` has only one field, it doesn't matter on it's and corresponding Frame name and we don't compare them as we do in cases where there is 2 and more fields in the RDD. Was that the intention ? If yes, I guess it make sense to have it like this, but I would put a comment in the code explaining this behaviour.

core/src/test/scala/org/apache/spark/h2o/utils/ExternalClusterModeTestHelper.scala
jakubhava:
```
  lazy val clusterStartTimeout = sys.props.getOrElse("cluster.start.timeout", 

sys.env.getOrElse("cluster.start.timeout", "6000")).toInt
```
Default value is now 6 seconds + user get override the default value by specifying env variable `cluster.start.timeout`

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: yes, good point, thanks!

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
jakubhava: Change has been done. It's more readable now hopefully and also we do this only in case of external cluster

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
jakubhava: Done

core/src/main/scala/org/apache/spark/h2o/converters/H2ORDD.scala
jakubhava: btw: I removed that comment for now, we can address it later in another PR

core/src/main/scala/org/apache/spark/h2o/converters/H2OFrameFromRDDProductBuilder.scala
jakubhava: Done 👍 Now it's done only for internal H2O backend and hopefully it's more readable

core/src/main/scala/org/apache/spark/h2o/utils/ReflectionUtils.scala
jakubhava: Done, it was my bad understanding. Now it's part of `ProductMember`

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
jakubhava: Thanks for the bug with FIFO. I'm now using  concurrent `LinkedBlockingDeque` with getLast() and replaced `filter` with `foreach` as you suggested 👍

core/src/main/scala/org/apache/spark/h2o/backends/external/ConnectionToH2OHelper.scala
jakubhava: I changed the `ConnectionToH2OHelper` a bit.

I realized we don't need to do the cleaning in a cleaning thread. I can simply just loop over the `availableConnections` right away and remove the closed connections.

Because there is no cleaning thread which might compete with the main thread, I do not need  concurrent structure such as `ConcurrentLinkedQueue` because `putAvailableConnection` and ` getOrCreateConnection` are synchronized.

That is also reason why I can do the filter again instead of foreach and reassign the `availableConnections` since we can be sure it's safe

jakubhava: But still need to run tests on it and make sure my assumption is correct

mdymczyk:
Looks ok but as @vpatryshev noticed we might have missed something due to no new tests. Even if we are reusing the old ones with new backend there still might be bugs that those tests weren't designed to catch (for example the concurrency issues with connections I pointed out in my previous review).

We'd need some integration tests with reasonable load/connections.

jakubhava:
Yes, agree that `ConnectionToH2OHelper` can be tested. I will write some tests for it. 

From my point of view, we don't need new integration tests, just JUunit tests for example for `ConnectionToH2OHelper` as you pointed out. Integration tests testing the connection and communication in general to external h2o cluster are the same as integration tests for normal sparkling-water. Those are larger tests testing integration of whole platform with various scenarios, on big and small data.

So I can definitely work on more Unit tests for some of the classes, but integration tests are not necessary as I see it

mdymczyk:
The difference between the old impl and the new external backend is that we're going to go over the wire now, which might be a problem (or not) with larger volumes of data/slower networks and from what I remember we don't really test for that since this wasn't really an issue before?

jakubhava:
It's true it's different behaviour, maybe it will worth having new tests. What aspects of network communication would you like to tests exactly ?

btw: I still have one tests scenario in my mind for long time - I need to test killing spark executors in the middle of jobs - I need support from @abal5 for it so we can run it as part of our test infrastructure. But that's next step for me once all the comments have been addressed here ( But still before the merging )

mdymczyk:
We could start with a simple scenario - start a spark cluster, h2o cluster with a few nodes and do some modelling on larger data (GBs at least). This would be a good start and, imho, enough to push this out the door. Later we can iterate with more sophisticated ones (slow network, nodes dying etc).

jakubhava:
I probably still don't see it, but the exact scenario is tested for example by this integration test https://github.com/h2oai/sparkling-water/blob/master/examples/src/integTest/scala/water/sparkling/itest/yarn/ChicagoCrimeTestSuite.scala

The data for full ChicagoCrimeTest suite isn't definitely small

mdymczyk:
Ah ok missed that one, then it looks good 👍

jakubhava:
Thanks for this discussion though, now I see that few more tests can be definitelly written 👍
 
- unit tests for `ConnectionToH2OHelper`
- spark nodes dying 

These 2 need to be available before we merge this

mmalohlava:
Minor problem:

1. Got :
```
scala> val h2oContext = H2OContext.getOrCreate(sc, conf)
16/12/01 07:33:21 WARN SparkSession$Builder: Use an existing SparkSession, some configuration may not take effect.
org.apache.spark.h2o.backends.external.H2OClusterNotRunning:

H2O Cluster is not running or couldn't be connected to. Provided configuration:
   cloud name            : michal

   cloud representative  : localhost:54321


 It is possible that in case you provided only cloud name h2o is not able to cloud up because multicast
 communication is limited in your network. In that case please consider setting "spark.ext.h2o.cloud.representative
"
 property and starting nodes in h2o cluster with -flatfile option describing the cluster
```

2. corrected H2OConf

3. call `val h2oContext = H2OContext.getOrCreate(sc, conf)`

4. got :
```
h2oContext: org.apache.spark.h2o.H2OContext =

Sparkling Water Context:
 * H2O name: michal
 * cluster size: 0
 * list of used nodes:
  (executorId, host, port)
  ------------------------

  ------------------------

  Open H2O Flow in browser: http://null:0 (CMD + click in Mac OSX)```

**Comment by jakubhava** [comment]
Good catch. I saw this behaviour before too but didn't have time to fix it. It's because `H2OContext` wasn't started successfully for the first time. It should be easy fix - don't set `H2OContext` instance if this exception is thrown. Will work on the fix when I have the time,

**Comment by jakubhava** [comment]
The problem with configuration is fixed. Because H2O can't be stopped on the JVM, even after unsuccessful connection to external h2o cluster we have h2o client running. Therefore we have to tell the user to restart the job or spark shell and apply new configuration.

File: core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala [hunk 33]
```diff
@@ -57,25 +65,15 @@ private[converters] object PrimitiveRDDConverter extends Logging with ConverterU
...
-      }
-      con.increaseRowCounter()
-    }
+    iterator.foreach {con.putAnySupportedType(0, _)}
     //Compress & write data in partitions to H2O Chunks
     con.closeChunks()
-
     // Return Partition number and number of rows in this partition
-    (context.partitionId, con.numOfRows)
+    (context.partitionId, iterator.size)
```
mmalohlava: `iterator.size` call can trigger exception, no? We should trakc number of rows during `iterator.foreach` and then use it here.

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -144,14 +153,13 @@ trait SparkDataFrameConverter extends Logging with ConverterUtils {
...
       }
-      con.increaseRowCounter()
     })
 
     //Compress & write data in partitions to H2O Chunks
     con.closeChunks()
 
     // Return Partition number and number of rows in this partition
-    (context.partitionId, con.numOfRows)
+    (context.partitionId, iterator.length)
```
mmalohlava: The same as above:
  - the iterator.length/iterator.size should not be used - instead of that pls compute number of items in foreach call.

core/src/main/scala/org/apache/spark/h2o/converters/DatasetConverter.scala
```diff
@@ -26,7 +26,7 @@ import scala.language.{implicitConversions, postfixOps}
 import scala.reflect.runtime.universe._
 
 
-trait DatasetConverter extends Logging {
+trait DatasetConverter extends Logging with Serializable{
```
mmalohlava: Just for tracking: why do we need Serializable here.

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
jakubhava: We need to know the size before because of external backend. We convert iterator to list, get the size and then convert it to a new iterator.

core/src/main/scala/water/support/H2OFrameSupport.scala
```diff
@@ -37,13 +37,24 @@ trait H2OFrameSupport extends JoinSupport {
     splitter.getResult
   }
 
+  def withLockAndUpdate[T <: Frame](fr: T)(f: T => Any): T = {
```
mmalohlava: Nice 👏

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
jakubhava: same as above. Fro my point of view there's reason why it's done like this

core/src/main/scala/org/apache/spark/h2o/converters/SupportedRDDConverter.scala
```diff
@@ -29,7 +29,7 @@ import scala.reflect.runtime.universe._
   * This converter just wraps the existing RDD converters and hides the internal RDD converters
   */
 
-trait SupportedRDDConverter{
+trait SupportedRDDConverter extends Serializable{
```
jakubhava: Scala creates bunch of java classes and one of those needs to be Serialisable because otherwise Spark complains that task is not serializable

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
mmalohlava: I am just saying, that we are breaking iterator invariant by knowing internal implementation If somebody changes internal implementation, we can be affected here. From my point of view, it is safer to pass here number of invocations from foreach above.

jakubhava: oki, will do. It's true that we are changing iterator default behaviour

core/src/main/scala/org/apache/spark/h2o/converters/SupportedRDDConverter.scala
mmalohlava: oki, so that means converters are transferred as part of tasks.

jakubhava: yup

core/src/main/scala/org/apache/spark/h2o/converters/PrimitiveRDDConverter.scala
mmalohlava: :+1: it is mostly to support maintainability

vpatryshev: Iterator in Scala is not something that a) guarantees to provide size (it can be unknown), b) is usable after sizing (TraversableOnce, e.g.) c) it may not even have a size.

If you have an iterator, using size may be destructive. If you know the size ahead, and it's not heavy, it may be better to use a different type. Seq, for instance.

jakubhava: It was a good point. I'm not changing iterator behaviour anymore. Thanks for pointing it out!

core/src/main/java/org/apache/spark/h2o/JavaH2OContext.java
mmalohlava: Sorry, i am still confused why JavaH2OContext needs to be Serializable - it should live only on client, never send via wire.

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -27,7 +28,7 @@ import org.apache.spark.repl.h2o.H2OInterpreter
   * Configuration holder which is representing
   * properties passed from user to Sparkling Water.
   */
-class H2OConf(@transient val sc: SparkContext) extends Logging with InternalBackendConf {
+class H2OConf(@transient val sc: SparkContext) extends Logging with ExternalBackendConf with InternalBackendConf {
```
mmalohlava: You do not need there to pass `sc`. `SparkConf` should be enough. I am right?

mmalohlava: At least you do not need to store SparkContext as field, just extracting SparkConf should be enough.

core/src/main/scala/org/apache/spark/h2o/converters/SupportedRDDConverter.scala
mmalohlava: interesting, that means they need to carry a state...

core/src/main/java/org/apache/spark/h2o/JavaH2OContext.java
jakubhava: I doesn't have to. I just missed this one during merging the commits

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
jakubhava: yup, I would prefer 2nd idea because it's also previous API and it's easier to call H2OConf(sc) then H2OConf(sc.conf)

PR #88, opened 2016-09-27
[SW-196] Fix wrong output of __str__ on H2OContext

jakubhava:
This PR fixes the problem of getting wrong output from `__str__` method on h2o context. 

The check for `has_attr` always returned false since it was called on `self` and field `is_initialized` was set on h2o, so the field weren't available in all cases.
This fixes the problem by setting the` is_initialized` attribute on `h2o_context` instance


mdymczyk:
LGTM

LGTM

mmalohlava:
Thx Kuba, pls merge into master, and cherry-pick into corresponding release branches (rel-2.0, rel-1.6). Than kyou!!!

PR #89, opened 2016-09-27
[SW-207] Use correct spark conf in SpreadRDDBuilder

jakubhava:
Right now the wrong `SparkConf` was used in `SpreadRDDBuilder` in sparkling water for Spark 1.5.

This PR fixes this problem by passing the builder `sparkConf` from h2oContext instead of from sparkContext. It is because we set additional properties to `sparkConf` in h2o context and we need to have these fields available.


jakubhava:
I know we decided not to update 1.5, but this was really easy fix. Can I merge into rel-1.5 ?

mmalohlava:
@jakubhava yes please.

PR #90, opened 2016-09-27
[MINOR] Move Java classes to java source dirs

mdymczyk:
- Moves all *.java files to src/main/java
- Adds src/main/java to source dirs in gradle


gradle/scala.gradle
```diff
@@ -49,12 +49,18 @@ test {
 sourceSets {
     main {
         scala {
-            srcDir "src/main/scala_$scalaBaseVersion"
+            srcDirs += ["src/main/scala_$scalaBaseVersion", "src/main/java"]
+        }
+        java {
+            srcDirs = []
         }
     }
```
jakubhava: Should we put java sources under java source dir ?
```main {
          scala {
                   srcDir "src/main/scala_$scalaBaseVersion"
         }
         java {
             srcDir "src/main/java"
          }
      }
``` 

And just to point out - `srcDir "src"` and `srcDirs += []` is equal, so we can stick wit any of those

mdymczyk: No, putting it in separate sets (java and scala) won't work because the Java classes depend on Scala classes and they won't see each other.

Hm I tried `srcDirs ["src/main/scala_$scalaBaseVersion", "src/main/java"]` but it didn't work (which was weird, the API has a srcDirs() method but I got some gradle exceptions) and `srcDir` doesn't take sets (only single dirs). Do you know what should be the exact syntax?

I did notice I was adding `main` in the `test` sources, though :-)

jakubhava: Good point regarding dependencies!

`srcDir` expects one resource dir and add it to the existing resource directories
`srcDirs = ` sets new set of source directories
`srcDirs += ` should add to the existing source directories

At least from what I know

jakubhava:
Good with me!

gradle/scala.gradle
mmalohlava: You do not need this line. Scala plugin by default apply java plugin.

jakubhava: So it seems that it is sufficient just to do `srcDir "src/main/scala_$scalaBaseVersion"` ?

mdymczyk: Hm do you mean I don't need the java part with srcDir = []? Or that I dont need to add the java source folder to scala srcDirs? Because I tried it without adding "src/main/java" and was getting compilation errors.

jakubhava: I think that you don't need to specify java part, but still need to put src/main/java to scala source. 

But regarding the very first comment, I'm trying to use Scala classes from java source dir and vice versa and it seems to see each other

mmalohlava: Hmm, look at JoinMethod - it is java class in the core and still compiled.

jakubhava:
@mmalohlava yup, I was trying to say that cross compilation should work

gradle/scala.gradle
mdymczyk: @mmalohlava yes but that's because it has no dependencies on Scala code. `JavaH2OContext` for example does (`SupportedRDD$`) and needs to be in the same srcDir as Scala sources.

Also the `java { srcDirs = [] }` part is necessary because otherwise it will try to compile Java classes twice (once for Scala compilation since I added it to Scala's `srcDirs` and once for Java source dirs -> this will fail because no Scala classes in the Java `srcDirs`).

mmalohlava: ahh, i see
so maybe at the end is better to keep java files in scala directory and do not move them. 
Nasty, but easy solution

mdymczyk: @mmalohlava well with this setting all works fine even when classes are in appropriate directories (java/scala) but if you think this config might cause trouble then I don't see any problems in keeping everything in scala folder.

mdymczyk:
so can I merge this or should I close the PR without merging @mmalohlava ? For me I'd like to have a clear separation but I'm ok with ditching it if you find it unnecessarily complex

jakubhava:
@mmalohlava @mdymczyk what about this change? I'm ok with both keeping it like it is or separating it. Just would like to either close or merge so we don't have hanging PRs:)

mmalohlava:
Will be merge in the next release.

PR #91, opened 2016-09-27
[SW-208] Generate all temporary and output artefacts into build dir

jakubhava:
This PR does exactly what it says. All temporary artefacts created during assembling python package and the output artefacts are put inside `build` directory.

### DIFF
--- a/py/build.gradle
+++ b/py/build.gradle
@@ -18,7 +18,10 @@ dependencies {
 // Create a file with version for Python dist task
 //
 task createVersionFile << {
-    File version_file = new File(getProjectDir(), "version.txt")
+    if(!getBuildDir().exists()){
+        getBuildDir().mkdir()
+    }
+    File version_file = new File(getBuildDir(), "version.txt")
     def version_txt = version.replace("-SNAPSHOT","")
     version_file.write(version_txt)
 }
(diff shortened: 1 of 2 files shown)


py/setup.py
```diff
@@ -10,7 +10,7 @@
     long_description = f.read()
 
 # Get the version from the relevant file
-with open(path.join(here, 'version.txt'), encoding='utf-8') as f:
+with open(path.join(here, 'build', 'version.txt'), encoding='utf-8') as f:
```
mdymczyk: same as above, shouldn't it be `getBuildDir()`?

py/build.gradle
```diff
@@ -185,12 +189,7 @@ check.dependsOn integTestPython
 // Cleanup
 //
 task cleanPython(type: Delete) {
-    delete file("dist/"),
-            file("h2o_pysparkling_${version.substring(0, version.lastIndexOf('.'))}.egg-info/"),
-            fileTree(dir: projectDir, include: '**/*.pyc'),
-            file("version.txt"),
-            file("h2o/"),
-            file("sparkling_water/")
+    delete file("build/")
```
mdymczyk: shouldn't this be `getBuildDir()`? Is it always `build`?

py/setup.py
jakubhava: `setup.py` is python code and doesn't have any information of outside world, that's why I can't use groovy `getBuildDir` method

py/build.gradle
jakubhava: it's always`build`, but good catch using `getBuildDir()` is better. Thanks!

py/setup.py
mdymczyk: ah right that's python, my bad.

mdymczyk:
LGTM

LGTM

py/build.gradle
mmalohlava: +1 for getBuildDir()

mmalohlava:
OKI, thx Kuba. Please merge it into master and cherry-pick directly into rel-2.0, rel-1.6, please. Use `-x` for cherry-pick

PR #92, opened 2016-09-27
[SW-200] Add flows from Paris to flow dir

jakubhava:
This PR add 2 additional flows to flow UI. IT doesn't add Spark SVM I used in Budapest and Paris since the current one seems to be newer. Also I added only flows from Paris since they are updated versions from Budapest.

There are 2 flows:
DataSourceAPI_Example - shows how to use h2o as spark's data source
Ham_Or_Spam_Pipeline - example of spark/h2o pipeline in flow. We load data from online repository, so anybody can actually try this flow without having to download example data. Also it demonstrates flow function `asDataFrame`

Tests for these flows will come in the next PR together with tests for the rest of flows


mdymczyk:
Hm I've been thinking about adding more textual description to the SVM flow, maybe we could also add some to this? Or do you think the code is enough? If so then LGTM.

jakubhava:
If you want we can improve SVM flow as part of this PR, sounds good with me 👍

mdymczyk:
Nah I think it's good for now.

mmalohlava:
Looks good. Please merge into master, and also rel-2.0 (if it works with rel-1.6, then also there).

PR #93, opened 2016-09-27
[SW-212] Fix deprecation warning in gradle build system

jakubhava:
Gradle is deprecating use of ant-based compiler and options connected to it.
 `scalaCompileOptions.useCompileDaemon` is now set to false automatically and will be completely removed in Gradle 3.0

`scalaCompileOptions.useAnt` is also automatically set to false ( trying to assign it will trigger the deprecation warning ) and will also be removed in Gradle 3.0

### DIFF
--- a/gradle/scala.gradle
+++ b/gradle/scala.gradle
@@ -3,8 +3,6 @@ apply from: "$rootDir/gradle/utils.gradle"
 
 // Activate Zinc compiler and configure scalac
 tasks.withType(ScalaCompile) {
-    scalaCompileOptions.useCompileDaemon = false
-    scalaCompileOptions.useAnt = false
     scalaCompileOptions.additionalParameters = ['-target:jvm-1.6']
 }
 


mdymczyk:
LGTM

LGTM

gradle/scala.gradle
```diff
@@ -3,8 +3,6 @@ apply from: "$rootDir/gradle/utils.gradle"
 
 // Activate Zinc compiler and configure scalac
 tasks.withType(ScalaCompile) {
-    scalaCompileOptions.useCompileDaemon = false
```
mmalohlava: Is it safe? What are defaults?

jakubhava: Default values for both options are `false`, so should be safe.

mmalohlava: Oki, perfect, Please merge into master and rel-2.0, rel-1.6

PR #94, opened 2016-09-27
[SW-169] Fix deprecation warnings

jakubhava:
This PR fixes deprecation warnings by introducing desired meta-annotaions.

This PR also introduces new trait and object. New trait `CrossScalaShared` is used to specify methods which each scala-version specific implementation should have and `CrossScalaUtils` is object implemented for each scala version with implementation tailored for specific scala.

This way we can utilities functions which will work on both scala versions and won't produce deprecations warnings/other problems in both cases.


mdymczyk:
LGTM

mmalohlava:
Oki,looks good!  @jakubhava please merge into master, and also cherrypick directly into rel-2.0.
Please run `cherry-pick` with option `-x` to record the original commit in the commit message.

PR #96, opened 2016-09-30
[SW-XXX] Start working on SW/Steam integration server

mdymczyk:
To run:

1) Steam:
- get this Steam branch https://github.com/h2oai/steam/tree/MD_sw_integration 
- in the etc/vagrant/HDPYarn do vagrant up and vagrant ssh
- do "go build"
- run with sudo `env PATH=$PATH:/usr/local/hadoop/bin YARN_CONF_DIR=/usr/local/hadoop/etc/hadoop SPARK_HOME=/usr/local/spark ./steamY serve master --superuser-name=vagrant --superuser-password=superuser`

! The Steam vagrant env is set up with Java7 so please build the below steps using Java7

2) Spark version:
- get this branch (`MD_steam_integration`)
- run `./gradlew clean install -x test -x integTest`

3) Sample engine application:
- get the droplet from https://github.com/h2oai/h2o-droplets/tree/MD_sw_steam_integ
- [optional] add code to `water.SparklingWater` class (_NOT_ `water.droplet.SparklingWaterDroplet`!)
- from sparkling-water-droplet run `./gradlew externalLauncher`
- zip the jar in `build/libs`
- upload the ZIP through Steam's UI _remember_ to choose Spark as engine
- fill out the other input fields and launch cluster

Possible problems: I think I have something wrong with my YARN/mac setup (network) so I had to add in Sparkling Water in the `water.messaging.driver.Driver` in the Spark Launcher this conf `.setConf("spark.ext.h2o.client.ip", "localhost")`

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -43,7 +43,9 @@ dependencies {
   compile "org.scala-lang:scala-library:${scalaVersion}"
 
   // Steam RPC
-  compile "com.github.briandilley.jsonrpc4j:jsonrpc4j:1.4.6"
+  compile("com.github.briandilley.jsonrpc4j:jsonrpc4j:1.4.6") {
+      exclude(group: "com.fasterxml.jackson.core")
+  }
   compile "javax.portlet:portlet-api:2.0"
 
   // And use scalatest for Scala testing
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -42,6 +42,10 @@ dependencies {
   // And Scala library
   compile "org.scala-lang:scala-library:${scalaVersion}"
 
+  // Steam RPC
+  compile "com.github.briandilley.jsonrpc4j:jsonrpc4j:1.4.6"
+  compile "javax.portlet:portlet-api:2.0"
+
   // And use scalatest for Scala testing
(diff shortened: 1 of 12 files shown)


mmalohlava:
Good first iteration!

There are lot of open questions:
  - do we need really to hardcode "Steam" inside SW? In fact, it is generic idea of publishing information to some "API" (hence i was slightly preferring a generic solution based on publishing messages to a bus)
  - is this going to be in the core code? Or is it going to be an optional extension?

mdymczyk:
@mmalohlava good questions:

1) yes I will change the name from Steam to something else (didn't have a better name at the time)

2) I would prefer to push it out into some external dependency - that's why I'm writing a part of it in Java so for example h2odriver could reuse it - the idea is that both h2odriver and SW (and any other JVM process) would reuse the server/client part and just implement an appropriate driver (SparkLauncher for SW and RunJar for h2odriver).

3) as per your other comments - I might try to make this even more generic with just a single method "handleMessage" on this side so Steam just pushes a message to it (almost no dependency on Steam side then) and this side would figure out how to handle it (using some metadata from the payload like the name of the method or something).

PR #98, opened 2016-10-03
Improve shell scripts

alexott:
This PR implements 2 things:

- Don't fail immediately if SPARK_HOME isn't set: Sometimes the Spark is installed via package manager, or some other means, like, Cloudera parcells.  In this case the `SPARK_HOME` isn't set explicitly, but `spark-shell` & `spark-submit` are available in the `PATH`.  This patch adds for this case, and fail only if both `SPARK_HOME` isn't set, and `spark-submit` isn't available in the `PATH`.  It also fixes explicit use of `FAT_JAR` variable instead of using `FAT_JAR_FILE` that points to the full path to `FAT_JAR`;
- Don't change current directory in scripts: Sometimes you need to run the `spark-shell` or other command from some directory with data, etc., but all scripts right now change directory to the `TOP_DIR`.

### DIFF
--- a/bin/pysparkling
+++ b/bin/pysparkling
@@ -10,5 +10,5 @@ checkSparkVersion
 
 
 PYTHONPATH=$PY_EGG_FILE:$PYTHONPATH \
-$SPARK_HOME/bin/pyspark --py-files $PY_EGG_FILE "$@"
+pyspark --py-files $PY_EGG_FILE "$@"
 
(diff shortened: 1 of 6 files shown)


mdymczyk:
LGTM thanks @alexott

jakubhava:
looks good to me too. 👍

mmalohlava:
Oki, @jakubhava pls merge it

PR #99, opened 2016-10-07
[SW-221] Unlocks the SVM model after driver finishes.

mdymczyk:


mdymczyk:
@jakubhava think this is enough? Don't want to pester Michal :-)

PR #100, opened 2016-10-07
[SW-230] Fix windows scripts

jakubhava:
This commit fixes several problems in windows scripts

- find instead of findstr was used to find string
- gradle.properties has linux line ending, by using type command we print the result on screen first with correct line endings
  and then it can be processed later
- scala version wasn't detected ( it tried to get scala version from sparkVersion field)

### DIFF
--- a/bin/sparkling-env.cmd
+++ b/bin/sparkling-env.cmd
@@ -5,13 +5,13 @@ if not defined TOPDIR (
     exit /b -1
 )
 
-rem Version of this distribution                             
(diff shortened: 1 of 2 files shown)


hachifm:
@jakubhava the sparkling-shell2.cmd should work with the Spark 1.6.0, right? I downloaded the edited file and tried to instal it again, but the same problem seems to keep happening. 
I also downloaded the 1.5 version and called for the sparkling-shell, no success. I'm totally new to all the terms, got assigned today to work with Machine learning, so I might be doing something wrong.

PR #101, opened 2016-10-10
[SW-232] rsparkling: fix CRAN note on invalid url

jangorecki:
Fix for a NOTE on invalid url
```
* checking CRAN incoming feasibility ... NOTE
Maintainer: ‘Erin LeDell <erin@h2o.ai>’

New submission

Non-FOSS package license (file LICENSE)

Found the following (possibly) invalid URLs:
  URL: https://jira.h2o.ai
    From: DESCRIPTION
    Status: Error
    Message: libcurl error code 7
    	Failed to connect to jira.h2o.ai port 443: Connection timed out
```
we do have ssl on our jira, but not on jira.h2o.ai, so it cannot be used here

### DIFF
--- a/r/rsparkling/DESCRIPTION
+++ b/r/rsparkling/DESCRIPTION
@@ -19,5 +19,5 @@ LazyData: true
 NeedsCompilation: no
 SystemRequirements: Java (>= 1.6)
 URL: https://github.com/h2oai/sparkling-water/tree/master/r/rsparkling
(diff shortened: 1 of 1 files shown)


jakubhava:
Good catch!
Before merging, @jangorecki can you pls create JIRA for this in Sparkling Water project and rename this commit as `[SW-XYZ ] Commit msg`, where `XYZ` is JIRA number

PR #109, opened 2016-10-13
[SW-237] Fix classloading issues in repl

jakubhava:


jakubhava:
Integration tests aren't passing at the moment for this branch, I know the cause and working on a fix

jakubhava:
More I think about it, we actually don't have to do it like the original code. I'm doing the binding instead of `processLine` since we have already spark context available and configured. In spark case this is the place where they create spark context for the first time and they do also additional configuration there which we don't have to do.

What I can do is to wrap `SparkSession.builder().gerOrCreate().sqlContext` and similar calls into helper method in an object and call it using `command` instead of binding, but that would be semantically the same as the binding.

Also if I'm not wrong processLine is available only on scala 2.11 so BaseH2OInterpreter couldn't be used as base for scala 2.10 and scala 2.11.

@mmalohlava what  do you think ?

jakubhava:
It is passing integration tests on yarn

mmalohlava:
:+1: oki, i am OK with working solution.

PR #111, opened 2016-10-14
Update and rename README.rst to README.md

Mistobaan:
WIP trying to change the format from rst to md and updating the versions. Here just for early comments

### DIFF
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -43,7 +43,7 @@ FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
 PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py2.7.egg"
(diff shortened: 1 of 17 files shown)


angela0xdata:
The reason for .rst is so that it is compatible with Sphinx-built documentation. The eventual output for Sparkling Water documentation is an html user guide similar to the Steam and H2O user guides. Our Sphinx theme doesn't support md, so we're switched to rst. Users who view the readme files on github will have no negative experience with reading rst vs md.

Mistobaan:
Seems you can use markdown files now in sphinx:

- http://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rst

angela0xdata:
That's correct, but md does not work with the sphinx template that we use.

jakubhava:
@angela0xdata so md is not supported in our documentation ? In that case can you @Mistobaan change this PR to just contain documentation fixes ( without converting it to md ) ?  Also can you pls create a JIRA for it in sparkling-water and put the jira number as part of the commit message as [SW-JIRA_NUM] Update Python documentation?

jakubhava:
Closing this PR since it's completely out of date

PR #113, opened 2016-10-20
add missing suggested pkgs, strict pkgs import, closes SW-238

jangorecki:
On my machine unit test will still fail but later on invalid spark dependencies ([SW-209](https://0xdata.atlassian.net/browse/SW-209)?). This will at least eliminate first error. It moves `Depends` to `Imports` and explicitly list imported functions in NAMESPACE. Starting from sparklyr 0.4.17 there should be no `Warning: replacing previous import` when building/installing rsparkling.
Note: I changed `strict_version_check=TRUE` https://github.com/h2oai/sparkling-water/pull/113/files#diff-4960598cc8d9bbb9a6bd79deec004716L15 according to the comment, @ledell what was the reason of FALSE?
This change also assume that we can maintain NAMESPACE manually instead of roxygen. If this would be problem it should be possible to use roxygen `#' @importFrom` but maintaining lists of imports is IMO easier in NAMESPACE file.


r/rsparkling/rsparkling.Rproj
```diff
@@ -13,4 +13,4 @@ StripTrailingWhitespace: Yes
 BuildType: Package
 PackageUseDevtools: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
-PackageRoxygenize: rd,collate,namespace
+PackageRoxygenize: rd,collate
```
mmalohlava: Do we need this file in repository?

r/rsparkling/DESCRIPTION
```diff
@@ -9,10 +9,11 @@ Authors@R: c(
   person("Jakub", "Hava", role = c("aut"), email = "jakub@h2o.ai"),
   person("RStudio", role = c("cph")),
   person("H2O.ai", role = c("cph")))
-Depends:
+Imports:
     utils,
     sparklyr (>= 0.3),
     h2o (>= 3.8.2.9)
```
mmalohlava: Does it mean that users need to explicitly import h2o package? What is expected user pattern?

r/rsparkling/rsparkling.Rproj
jangorecki: I'm not using it, it make sense only if you want to sync your rstudio settings, quite developer-specific. I can do git rm and add to gitignore.

mmalohlava: I do not know what is common R community practice - keep and share, or remove? In Java world we do not share user specific settings...

r/rsparkling/DESCRIPTION
jangorecki: Yes. This change is a matter if rsparkling is acting as API without the need to call h2o functions directly by user. If rsparkling can be used in client mode where it use only own methods then why force user to attach h2o to `searchpaths()`. This is more a good practice, some people say it is _anti-social_ as it can cause conflicts in downstream dependencies built on top of rspkarling. In this case it is pretty unlikely, so it isn't as important. Let me know if you want to move h2o and/or sparklyr to depends.

r/rsparkling/rsparkling.Rproj
jangorecki: I wouldn't keep it, @erin @navdeep-G do you need to keep `rsparkling.Rpoj` in git history?

navdeep-G: @jangorecki @ledell I don't keep it.

r/rsparkling/DESCRIPTION
jangorecki: If we are moving those deps from Depends to Imports I will also need to update `r/examples/example_rsparkling.R`, so PR in current state is not yet ready to merge.

jakubhava:
This looks good to me, but I'm still starting with R. Can we squash and merge ? ( Before merging at some point later - @jangorecki please create JIRA for it and put the jira number into the commit message )

jangorecki:
@jakubhava it was in single commit already. I just hit "update branch" and now there are two commits :( will rebase to latest master and force push. "SW-238" is already in commit message.
It still needs approval from @ledell about `strict_version_check=TRUE` change.
It also requires to update examples/scripts by adding `library(h2o)` or `library(sparklyr)` as those packages are not attached anymore when attaching rsparkling.

jangorecki:
We can go with `strict_version_check=TRUE`, so now only reviews needed

PR #110, opened 2016-10-13
[SW-217] Add unit testing framework to RSparkling

navdeep-G:
*This PR adds a unit testing framework for RSparkling (Note: It is not complete, but a good start)
*To run simply call `./gradlew testR` in top directory of sparkling water

PR #116, opened 2016-10-21
[SW-209] [RSparkling] Improve handling of Sparkling Water package dependencies

navdeep-G:
Right now, the file https://github.com/h2oai/sparkling-water/blob/master/r/rsparkling/R/package.R
enforces a specific version of Sparkling Water package.
However, we should:
verify compatibility of Sparkling Water version with passed `spark_version` argument
configure version of Sparkling Water automatically (rely on user specified version)


r/rsparkling/R/package.R
```diff
@@ -1,13 +1,24 @@
-
 #' @import sparklyr h2o utils
 NULL
 
 # define required spark packages
 spark_dependencies <- function(spark_version, scala_version, ...) {
+  sw_version = getOption("rsparkling.sparklingwater.version", default = "1.6.7")
```
jakubhava: I'm thinking if it wouldn't be easier to throw error - warn the user - that the version has to be set as `+options(rsparkling.sparklingwater.version = "2.0.0")` in order to use sparkling-water. It's more explicit and we don't have to update default version.

Because like this, we would need to update default version with each sparkling-water release ( right now it should be 2.0.0 )

What do you think @navdeep-G, is this doable ? What do you @mmalohlava think ? I know we discussed that in the office, so wanted to know your input on this. 

btw: thanks @navdeep-G for this!

r/rsparkling/R/package.R
```diff
@@ -1,13 +1,24 @@
...
+
+  if(as.package_version(spark_version)$major != as.package_version(sw_version)$major){
+    stop(cat(paste0("Major version of Sparkling Water does not correspond to major Spark version.
+    \nMajor Sparkling Water Version = ",as.package_version(sw_version)$major,
+    "\nMajor Spark Version = ",as.package_version(spark_version)$major)))
+  }
+  if(as.package_version(spark_version)$minor != as.package_version(sw_version)$minor){
```
jakubhava: I think that we shouldn't stop the application if the minor versions differ, I would just warn in that case. We should stop only in case when major version differ.

r/rsparkling/R/package.R
```diff
@@ -1,13 +1,24 @@
...
+     stop(cat(paste0("Minor version of Sparkling Water does not correspond to minor Spark version.
+     \nMinor Sparkling Water Version = ",as.package_version(sw_version)$minor,
+     "\nMinor Spark Version = ",as.package_version(spark_version)$minor)))
+  }
   sparklyr::spark_dependency(packages = c(
-    sprintf("ai.h2o:sparkling-water-core_%s:1.6.7", scala_version),
-    sprintf("ai.h2o:sparkling-water-ml_%s:1.6.7", scala_version),
-    sprintf("ai.h2o:sparkling-water-repl_%s:1.6.7", scala_version)
+    sprintf("ai.h2o:sparkling-water-core_%s:%s", scala_version,sw_version),
```
jakubhava: We don't check scala_version sparkling-version compatibility - this problem happened to me today - t occurred when testing https://github.com/h2oai/sparkling-water/pull/110

```
:: USE VERBOSE OR DEBUG MESSAGE LEVEL FOR MORE DETAILS
Exception in thread "main" java.lang.RuntimeException: [unresolved dependency: ai.h2o#sparkling-water-core_2.11;1.6.7: not found, unresolved dependency: ai.h2o#sparkling-water-ml_2.11;1.6.7: not found, unresolved dependency: ai.h2o#sparkling-water-repl_2.11;1.6.7: not found]
        at org.apache.spark.deploy.SparkSubmitUtils$.resolveMavenCoordinates(SparkSubmit.scala:1066)
        at org.apache.spark.deploy.SparkSubmit$.prepareSubmitEnvironment(SparkSubmit.scala:294)
        at org.apache.spark.deploy.SparkSubmit$.submit(SparkSubmit.scala:158)
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:124)
        at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)```

My `SPARK_HOME` points to spark 2.0 which uses scala 2.11 as default one. The same check is done by sparklyr. So right now it's fetching non-existing package from maven central `sparkling-water-repl_2.11;1.6.7`

We need to check that when user wants to use sparkling water 1.6 and lower, scala 2.10 should be used and scala 2.11 for sparkling water 2.0 - otherwise throw an error.

File: r/rsparkling/R/package.R (hunk 1 above)
Comment (jangorecki) [inline r/rsparkling/R/package.R]: [this](https://0xdata.atlassian.net/browse/SW-209?focusedCommentId=37711&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-37711) chunk would avoid hardcoding sparkling version but adjust it to matching spark version. We still need to hardcode mapping of spark-sparkling versions.

File: r/rsparkling/R/package.R (hunk 2 above)
Comment (jangorecki) [inline r/rsparkling/R/package.R]: this "minor" isn't really minor
```
#as.package_version("1.2.3")$major
[1] 1
as.package_version("1.2.3")$minor
#[1] 2
```
real minor is "patchlevel" here
```
as.package_version("1.2.3")$patchlevel
#[1] 3
```

Comment (jakubhava) [inline r/rsparkling/R/package.R]: Ok, good to know!

File: r/rsparkling/R/package.R (hunk 1 above)
Comment (jakubhava) [inline r/rsparkling/R/package.R]: @jangorecki But it still hardcodes the versions in the mapping - with each new version, you would need to change the mapping and thus RSparkling. From my point of view, explicit setting is better.

Regarding 
> We still need to hardcode mapping of spark-sparkling versions.
 sparkling-water version tells us for which spark and scala it's being targeted, so this may not be necessary too

File: r/examples/example_rsparkling.R [hunk 4]
```diff
@@ -15,13 +15,17 @@
 # devtools::install_github("h2oai/sparkling-water", subdir = "/r/rsparkling")
 # devtools::install_github("rstudio/sparklyr")
 
+#Set version of spark. 
+#Note: At the time of spark_connect sparklyr will call the spark_dependencies function in the rsparkling package.
+options(rsparkling.sparklingwater.version = "1.6.7")
```
mmalohlava: What i do not like on this one is that it forces explicit order of operations: you need to setup version first and then call  `spark_connect`. There was already a problem report in community of forum which was reporting error in RSparkling, because of wrong operations order. This even more stress user. 

But I understand that `spark_connect(master = ..., extensions = add(rsparkling(version="1.6.8")) )` is not best one as well.

r/rsparkling/R/package.R
navdeep-G: Assuming this is fine with all parties? The `requested changes` blocks from merge later. ;)

r/examples/example_rsparkling.R
jakubhava: I actually like the `spark_connect(master = ..., extensions = add(rsparkling(version="1.6.8")) )` more than the `options` one, but I got the feeling that it wasn't possible to do, or am I wrong @navdeep-G ?

r/rsparkling/R/package.R
jakubhava: yup, it was my bad understanding of `minor` here

r/examples/example_rsparkling.R
navdeep-G: @jakubhava Yes, I could not find a way to pass the `version` to `extensions`

r/rsparkling/R/package.R
navdeep-G: @jakubhava Are you thinking of something like this:

`  sw_version = getOption("rsparkling.sparklingwater.version", default = NULL)
  if(is.null(sw_version)){
    stop("Sparkling Water version is not set. Please choose a correct version")
  }`

navdeep-G:
@jakubhava @mmalohlava @jakubhava @ledell I have added the following to `package.R` to ensure a user passes in a SW version:
https://github.com/h2oai/sparkling-water/pull/116/commits/35628b1d55af268170d5c4805fd85830d3eeff33

navdeep-G:
@jangorecki One more look and were good.

r/rsparkling/R/package.R
```diff
@@ -2,13 +2,25 @@
 #' @importFrom sparklyr spark_dependency register_extension invoke_static invoke spark_connection spark_dataframe sdf_register spark_context
 #' @importFrom h2o h2o.getFrame h2o.getId h2o.init
 
+#' @import sparklyr h2o utils
```
jangorecki: Instead of importing whole namespace use `#' @importFrom pkg fun`, we don't want to get into namespace collisions. When changing imports you also need to render NAMESPACE each time, otherwise it won't take effect.

navdeep-G:
All good now I believe.

jangorecki:
Currently test script was failing. I removed it, it is going to be replaced anyway in coming PR.

ledell:
@mmalohlava @jangorecki @navdeep-G Is this ready to be merged?

jangorecki:
@ledell yes

PR #117, opened 2016-10-28
consider adding name parameter to as_spark_dataframe

javierluraschi:
Some of our users expected a friendly default name once they convert their h2o frames back to spark data frames with `as_spark_dataframe`. Consider setting the default name to the name of the variable `x` which is what `sparklyr` does during `copy_to` and similar methods.

### DIFF
--- a/r/rsparkling/R/h2o_context.R
+++ b/r/rsparkling/R/h2o_context.R
@@ -42,7 +42,7 @@ h2o_flow <- function(sc) {
 as_h2o_frame <- function(sc, x) {
   # sc is not actually required since the sc is monkey-patched into the Spark DataFrame
   # it is kept as an argument for API consistency
-  
+
   # Ensure we are dealing with a Spark DataFrame (might be e.g. a tbl)
   x <- sparklyr::spark_dataframe(x)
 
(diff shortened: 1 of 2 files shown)


mdymczyk:
Looks good to me, @ledell @jangorecki any comments?

jakubhava:
looks good. We just need to create a JIRA for it 👍

jangorecki:
1. feature could be tested
2. it is safer to use `paste(deparse(substitute(x), width.cutoff=500L), collapse="")` as `deparse` might not return scalar.

mmalohlava:
Thank you @javierluraschi !

mmalohlava:
@jangorecki ups, sorry, i just merge it. Do you want to revert it?

jangorecki:
no, this is minor thing that can be added later

javierluraschi:
@jangorecki we use this same code in `sparklyr` (https://github.com/rstudio/sparklyr/blob/8d142c4afa058feb5bf3132836a17bda65c177a2/R/dplyr_spark.R#L81) so if there are cases that are not handled correctly, I would love to know. Which case do you have in mind that might break this? Thanks!

jangorecki:
@javierluraschi try the following
```r
f = function(x, name=deparse(substitute(x))) name
f(x = {
  a = 1L
  data.frame(a)
})
```
also see `?deparse` and `args(as.data.frame.vector)`

PR #118, opened 2016-11-01
[SW-223] H2OModel/Algo and H2OPipeline recatoring

mdymczyk:
This PR will have quite a few changes to H2OAlgo/Model and H2OPipline (possibly removed). Considering splitting it into multiple ones but let's see how it goes.


mdymczyk:
@vpatryshev Spark does it in a peculiar way because they don't need it, their ML algorithms all work on `Vector[Double]`, H2O's ML algorithms can have columns of mixed types. The reason I'm doing it this way is to make the API more Spark like where you would set the `featuresColumn(name)` for an ML algorithm (also H2O algorithm) which contains a `Vector` (or `Array`, `Seq` etc.) with all the features. I'm open to suggestions how this can be defined in a more Scala idomatic way.

vpatryshev:
It should be worked on on a case by case basis, I guess. There may be
logical/type problems with H2O algorithms. I see a lot of transformations
enum->int-double and back in our code. These operations are not reasonable.
Scala can do better.

Thanks,
-Vlad

mdymczyk:
@vpatryshev "I guess", sorry Vlad I can't go with just I guess :-) The problem is that H2O's and Spark MLlib's APIs/approaches are a bit different, don't think it's a Scala problem but more a Spark one:

1) Spark's algorithms work *only* on numeric vectors (so they are ok with `Vector[Double]`) whereas H2O's algorithms can work on different datatypes (String, Int, Double) for the same algorithm
2) Spark's algorithms have all features in one column which contains `Vector[Double]`, whereas H2O's algorithms work on a Frame which has each feature in a separate column

Personally in SW I'd prefer to have an API for our wrappers which is consistent with MLlib's API even if it means a small hack.

This isn't a language (Scala) problem, since in Scala I could simply make a `Array[Any]`, which is not a supported type for a column in Spark (not sure why, might be because Tungsten wouldn't be able to optimize it, not sure).

There are 2 approaches I see:

1) go with the approach I'm doing here, a bit hacky but in line with Spark's API
2) mix both approaches, use `featureColumns(Array[String])` in H2O wrappers which resembles Spark's API but uses multiple columns instead of one and use those columns for training/predicting.

Or I'm open to other solutions but they have to be generic and user friendly, not a case-by-case solutions.

@jakubhava @mmalohlava thoughts welcome.

vpatryshev:
I see there are problems. Sure. But any solution that just drops the type
of data is, I am sure, a bad solution.

Thanks,
-Vlad

mdymczyk:
Well I'm trying to preserve the type, in a pretty clunky way I must admit, since `AnyType` has both the value field and the type field and I'm planning to convert this vector to an H2O frame where each column would have an appropriate type inferenced from the type field. In my 2nd solution (multiple feature columns) this would be handled automatically by our current converters. 

The only place we'd loose the actual type check is in the DataFrame column.

vpatryshev:
Well, yes, the biggest problem lies in H2O Frames...

Thanks,
-Vlad

jakubhava:
@mdymczyk if you don't mind, I will close this PR but will possible re-use some ideas. I started to working on the pipelines in sw so it will be overlapping a bit.

My first iteration is:
1) Use MOJO when possible in pipelines instead of model requiring h2o environment 
2) Get rid of OneTimeTransformers
3) Get rid of H2OPipeline

PR #119, opened 2016-11-01
Fixed spark version extraction

vpatryshev:
Taking care of NPE and IOOBE.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
@@ -19,6 +19,7 @@ package org.apache.spark.h2o.utils
 
 import org.apache.spark.internal.Logging
 import org.apache.spark.SparkContext
+import language.postfixOps
 
 /**
   * Support methods for H2OContext.
@@ -63,14 +64,22 @@ private[spark] trait H2OContextUtils extends Logging{
     * For example, for 1.6.1 returns 1.6
     */
   def buildSparkMajorVersion = {
-    val stream = getClass.getResourceAsStream("/spark.version")
-    val version = scala.io.Source.fromInputStream(stream).mkString
-    if (version.count(_ == '.') == 1) {
-      // e.g., 1.6
-      version
-    } else {
-      // 1.4
-      version.substring(0, version.lastIndexOf('.'))
+    val VERSION_FILE: String = "/spark.version"
+    val stream = getClass.getResourceAsStream(VERSION_FILE)
+    
+    stream match {
+      case null => s"Unknown spark version: $VERSION_FILE missing"
+      case s =>
+        val version = scala.io.Source.fromInputStream(s).mkString
+        
+        if (version.count('.'==) <= 1) {
+          // e.g., 1.6 or "new"
+          version
+        } else {
+          // 1.4
+          version.substring(0, version.lastIndexOf('.'))
+        }
     }
+    
   }
 }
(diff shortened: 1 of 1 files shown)


jakubhava:
I'm not sure if this is necessary since we always put `spark.version` file into the resources. The case where it's missing should never happen. If we decide to check it anyway, I would rather check assembly jar whether it contains the file or not as part of gradle build ( because if it's missing that means something went wrong with the build )

vpatryshev:
Jakub, if it does not happen, we can as well throw an exception explaining in details what's wrong. Why make developers life hard? Do you like it? I talked with Michal, he agreed that we should return some text. Are you going to check assembly jars all around the world? Or? What exactly is your suggestion? I'll gladly implement it. But please let not keep uses in the dark regarding the problems, ok? Just share your idea, and let's make sure Michal has the same opinion. I personally don't care either way. I don't want to throw obscure exceptions, that's it. Do you like obscure exceptions? What's their purpose?

mmalohlava:
Checking on both sides makes good sense:
  - at build time to produce correct artifacts
  - at runtime since people can bundle sparkling-water-core into FatJar and have different fat jarring process with different filters (e.g., skipping `spark.properties` for some reason).

Thinking about it more - probably we should have an option to disable the compatibility check..

vpatryshev:
Disabling compatibility check is not in the scope of this small change. I have no opinion, probably Michal knows better whether it makes sense. Or we could return and Option[String].

What's exactly your suggestion regarding the current code?

jakubhava:
@vpatryshev I understand your point of view, but your message feels like filled with emotions instead of explanation why and why not it's bad or good. I'm always opened to other ideas, the fact I wrote the first comment doesn't mean I will end up liking it et the end after ideas of others, I like to collaborate. Also I can't know what people discussed in the office and what not, since I'm remote.

vpatryshev:
Jakub, sure, we may have talked about something, but it was not reflected anywhere.

So, what happened. I updated SPARK on my machine yesterday, built an app with gradlew, and it started throwing this error. Of course the cause is somewhere outside. But we should report it properly, not keep it enigmatic. I guessed the cheapest solution (I'm not its fan, but well, I'm trying to be soft) was to provide some human-readable text. I would prefer throwing an exception instead. So if you agree with me, I'll replace returning that nice text (which hardly anybody will read, for obvious reason) with an exception. But please come up with some suggestion, if you don't feel this current solution is good. :)

jakubhava:
@vpatryshev thanks for explanation! 

I think this solution is fine. It makes sense that we should check it at build and runtime as @mmalohlava pointed out. 

I'm not sure whether to choose the message or the exception. From my point of view an exception should be thrown - but on the other hand - no code doesn't depend on the spark version directly so I can imagine that at some points, the exception would be too much and text makes sense. If you decided to go with the text with @mmalohlava we can keep it like that.

vpatryshev:
Let's ask Michal:

Michal, do you mind if I throw an exception instead?

mmalohlava:
+1 for exception - we need to make sure that users running SW on right version of Spark. When we can see wrong version or no version => throw exception.

Later we should introduce an option to override this.

vpatryshev:
Thank you all; now an exception is thrown!

jakubhava:
@vpatryshev Could you please squash the commits, create JIRA for it, put the JIRA number - [SW_NUMBER] - to commit as prefix to the message and rebase on top of current master ? Then we can merge 👍

vpatryshev:
OK, sure.

Thanks,
-Vlad

PR #121, opened 2016-11-02
[SW-248] Python 3 incompatibilities

mdymczyk:
Fixes:
* Links to Sparkling Water and PySparkling (used links to 2.0 now)
* Fixed some print statements in PySparkling
* Fixed some Python3 incompatibilities in the tensorflow demo, should be still ok with Python2


mdymczyk:
@jakubhava yes I ran it with Python 2.7 and 3.5 before pushing

PR #125, opened 2016-11-03
Add new testing framework for rsparkling

navdeep-G:
* This PR adds a new testing framework for rsparkling (different than the testing framework in the branch `NG_SW217`)
* To run the tests one simply does the following `R -f testthat.R`. Note, running tests requires suggested dependencies (normally optional), also listed in `testthat.R`.
* Also, this testing framework is completely independent of Sparkling Water testing framework.


r/rsparkling/tests/testthat/test_transforms.R
```diff
@@ -0,0 +1,100 @@
+library(rsparkling)
```
jangorecki: we don't need to load rsparkling again, we already did in init script

r/rsparkling/tests/testthat/test_transforms.R
```diff
@@ -0,0 +1,100 @@
...
+
+})
+
+
+test_that("Test transformation from dataframe to h2o frame", {
+   sc <- spark_connect(master = "local[*]")
+   mtcars_tbl <- copy_to(sc, mtcars, overwrite = TRUE)
+   mtcars_hf_name <- as_h2o_frame(sc, mtcars_tbl, name = "frame1")
+
+   expect_equal(h2o::h2o.getId(mtcars_hf_name), "frame1")
```
jangorecki: instead of `h2o::` better IMO to add `library(h2o)` in the init `tests/testthat.R`, otherwise there will be a lot of `h2o::` in future calls to h2o

r/rsparkling/tests/testthat.R
```diff
@@ -0,0 +1,7 @@
+if(requireNamespace("testthat", quietly = TRUE)){
```
navdeep-G: I would opt for just the library calls then. No `if(){}` required.

r/rsparkling/tests/testthat/test_transforms.R
navdeep-G: I thought that was common practice, no?

r/rsparkling/tests/testthat/test_transforms.R
navdeep-G: Good idea

r/rsparkling/tests/testthat/test_transforms.R
jangorecki: This is required when running just R scripts in `tests`. If there is testthat.R then it will process all tests as a childs, so with testthat you can attach packages globally for testthat tests.

navdeep-G:
All is updated. @jangorecki please take one more look.

navdeep-G:
@jangorecki Do you mean in general or for CRAN?

ledell:
@navdeep-G Is this ready to be merged?  Looks ok to me.

navdeep-G:
Looks good to me. @jangorecki @mmalohlava ?

PR #130, opened 2016-11-10
making tests portable, and fixes to pass R check, closes SW-259

jangorecki:
rsparkling tests are now portable and _escaped on CRAN_, few minor fixes had been made to pass R check, also additional argument `strict_version_check` was added to make tests working (they can eventually be removed when we improve handling `h2o_context` in rsparkling.
I also added few comments in-line for clarify the changes.
To run unit tests we need to, should run on your local machine.
```sh
R CMD build r/rsparkling
export NOT_CRAN="true"
R CMD check rsparkling_0.1.0.tar.gz
```
To change spark/sw version use env vars:
```sh
# default
export SPARKLINGWATER_VERSION="1.6.7"
export SPARK_VERSION="1.6.2"
# or
export SPARKLINGWATER_VERSION="2.0.0"
export SPARK_VERSION="2.0.1"
```
other values has not been tested, if software in specific version is missing it will be installed (if valid).

### DIFF
--- a/r/rsparkling/R/h2o_context.R
+++ b/r/rsparkling/R/h2o_context.R
@@ -17,10 +17,9 @@ h2o_context.spark_connection <- function(x, strict_version_check = TRUE) {
 }
 
 #' @export
-h2o_context.spark_jobj <- function(x) {
-  h2o_context(spark_connection(x))
+h2o_context.spark_jobj <- function(x, strict_version_check = TRUE) {
+  h2o_context.spark_connection(spark_connection(x), strict_version_check=strict_version_check)
 }
-
 
 #' Open the H2O Flow UI in a browser
 #'
@@ -37,20 +36,22 @@ h2o_flow <- function(sc) {
 #'
 #' @param sc Object of type \code{spark_connection}.
 #' @param x A \code{spark_dataframe}.
-#' @param name The name of the H2OFrame
+#' @param name The name of the H2OFrame.
+#' @param strict_version_check (Optional) Setting this to FALSE does not cross check version of H2O and attempts to connect.
 #' @export
-as_h2o_frame <- function(sc, x, name=NULL) {
+as_h2o_frame <- function(sc, x, name=NULL, strict_version_check=TRUE) {
   # sc is not actually required since the sc is monkey-patched into the Spark DataFrame
   # it is kept as an argument for API consistency
 
   # Ensure we are dealing with a Spark DataFrame (might be e.g. a tbl)
   x <- spark_dataframe(x)
-
+  
   # Convert the Spark DataFrame to an H2OFrame
-  if(is.null(name)){
-    jhf <- invoke(h2o_context(x), "asH2OFrame", x)
-  }else{
-    jhf <- invoke(h2o_context(x), "asH2OFrame", x, name)
+  hc <- h2o_context(x, strict_version_check=strict_version_check)
+  jhf <- if(is.null(name)) {
+    invoke(hc, "asH2OFrame", x)
+  } else {
+    invoke(hc, "asH2OFrame", x, name)
   }
 
   key <- invoke(invoke(jhf, "key"), "toString")
@@ -62,15 +63,15 @@ as_h2o_frame <- function(sc, x, name=NULL) {
 #' @param sc Object of type \code{spark_connection}.
 #' @param x An \code{H2OFrame}.
 #' @param name The name to assign the data frame in Spark.
-#'
+#' @param strict_version_check (Optional) Setting this to FALSE does not cross check version of H2O and attempts to connect.
 #' @export
-as_spark_dataframe <- function(sc, x, name = deparse(substitute(x))) {
+as_spark_dataframe <- function(sc, x, name = paste(deparse(substitute(x)), collapse=""), strict_version_check=TRUE) {
   # TO DO: ensure we are dealing with a H2OFrame
 
   # Get SQLContext
   sqlContext <- invoke_static(sc, "org.apache.spark.sql.SQLContext", "getOrCreate", spark_context(sc))
   # Get H2OContext
-  hc <- h2o_context(sc)
+  hc <- h2o_context(sc, strict_version_check=strict_version_check)
   # Invoke H2OContext#asDataFrame method on the backend
   spark_df <- invoke(hc, "asDataFrame", h2o.getId(x), TRUE, sqlContext)
   # Register returned spark_jobj as a table for dplyr
--- a/r/rsparkling/R/package.R
+++ b/r/rsparkling/R/package.R
@@ -5,9 +5,9 @@ NULL
 
 # define required spark packages
 spark_dependencies <- function(spark_version, scala_version, ...) {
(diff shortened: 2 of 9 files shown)


r/rsparkling/R/h2o_context.R
```diff
@@ -17,11 +17,10 @@ h2o_context.spark_connection <- function(x, strict_version_check = TRUE) {
 }
 
 #' @export
-h2o_context.spark_jobj <- function(x) {
-  h2o_context(spark_connection(x))
+h2o_context.spark_jobj <- function(x, strict_version_check = TRUE) {
+  h2o_context.spark_connection(spark_connection(x), strict_version_check=strict_version_check)
```
jangorecki: because we already convert `x` with `spark_connection()` we can explicitly direct to `h2o_context.spark_connection` method, this avoid extra dispatching via `h2o_context`

r/rsparkling/R/h2o_context.R
```diff
@@ -37,20 +36,22 @@ h2o_flow <- function(sc) {
 #'
 #' @param sc Object of type \code{spark_connection}.
 #' @param x A \code{spark_dataframe}.
-#' @param name The name of the H2OFrame
+#' @param name The name of the H2OFrame.
+#' @param strict_version_check (Optional) Setting this to FALSE does not cross check version of H2O and attempts to connect.
 #' @export
-as_h2o_frame <- function(sc, x, name=NULL) {
+as_h2o_frame <- function(sc, x, name=NULL, strict_version_check=TRUE) {
```
jangorecki: New argument is required for now because `as_h2o_frame` call `h2o.init()` in fixed version available in maven repo(?). At the moment it is sparkling water limitation AFAIK. Argument could be removed when we handle this better.

r/rsparkling/R/h2o_context.R
```diff
@@ -62,15 +63,15 @@ as_h2o_frame <- function(sc, x, name=NULL) {
 #' @param sc Object of type \code{spark_connection}.
 #' @param x An \code{H2OFrame}.
 #' @param name The name to assign the data frame in Spark.
-#'
+#' @param strict_version_check (Optional) Setting this to FALSE does not cross check version of H2O and attempts to connect.
 #' @export
-as_spark_dataframe <- function(sc, x, name = deparse(substitute(x))) {
+as_spark_dataframe <- function(sc, x, name = paste(deparse(substitute(x)), collapse=""), strict_version_check=TRUE) {
```
jangorecki: new argument also required to run tests

r/rsparkling/tests/testthat.R
```diff
@@ -4,5 +4,11 @@ library(sparklyr)
 library(h2o)
 library(rsparkling)
 
-test_check("rsparkling")
+options(rsparkling.sparklingwater.version = Sys.getenv("SPARKLINGWATER_VERSION","1.6.7"),
+        rsparkling.spark.version =          Sys.getenv("SPARK_VERSION","1.6.2"))
```
jangorecki: `rsparkling.spark.version` is the new option which was only used in tests. I've added this here so we have single place to control spark and sw version. Both are using env vars to make it easy to change in CI.

r/rsparkling/tests/testthat.R
```diff
@@ -4,5 +4,11 @@ library(sparklyr)
...
 
+if(identical(Sys.getenv("NOT_CRAN"), "true")) { # testthat::skip_on_cran
```
jangorecki: `testthat::skip_on_cran` must be used locally in each tests, I extract exact logic of it and put it here so it works globally on all tests now.

r/rsparkling/R/h2o_context.R
navdeep-G: I think this argument is necessary. Otherwise you limit a user to a specific version of H2O. Sparkling Water can handle any minor version AFAIK. Major version I am not sure. cc @mmalohlava

r/rsparkling/tests/testthat/test_transforms.R
```diff
@@ -3,98 +3,89 @@ context("Test transformations of H2O frames and Spark frames in rsparkling")
...
-  df_tbl = as_spark_dataframe(sc,df_hex)
-
-  expect_equal(as.data.frame(tally(df_tbl))$n,nrow(df_hex))
-  expect_equal(ncol(df_tbl),ncol(df_hex))
-  expect_equal(colnames(df_tbl),colnames(df_hex))
-
-
+  df <- as.data.frame(t(c(1,2,3,4,"A"))) # workaround for sparklyr#316
+  sdf <- copy_to(sc, df)
+  hf <- as_h2o_frame(sc, sdf, strict_version_check=FALSE)
```
jangorecki: `strict_version_check=FALSE` required to run tests because user h2o version might be different that h2o fixed in SW, other changes in this files are more cosmetic.

r/rsparkling/R/h2o_context.R
```diff
@@ -37,20 +36,22 @@ h2o_flow <- function(sc) {
...
-  if(is.null(name)){
-    jhf <- invoke(h2o_context(x), "asH2OFrame", x)
-  }else{
-    jhf <- invoke(h2o_context(x), "asH2OFrame", x, name)
+  hc <- h2o_context(x, strict_version_check=strict_version_check)
+  jhf <- if(is.null(name)) {
+    invoke(hc, "asH2OFrame", x)
+  } else {
+    invoke(hc, "asH2OFrame", x, name)
   }
```
navdeep-G: Variable names are good here. Good addition.

r/rsparkling/R/package.R
```diff
@@ -20,13 +20,13 @@ spark_dependencies <- function(spark_version, scala_version, ...) {
      "\nMinor Spark Version = ",as.package_version(spark_version)$minor)))
   }
   spark_dependency(packages = c(
-     sprintf("ai.h2o:sparkling-water-core_%s:%s", scala_version, sw_version),
-     sprintf("ai.h2o:sparkling-water-ml_%s:%s", scala_version, sw_version),
-     sprintf("ai.h2o:sparkling-water-repl_%s:%s", scala_version, sw_version)
+    sprintf("ai.h2o:sparkling-water-core_%s:%s", scala_version, sw_version),
+    sprintf("ai.h2o:sparkling-water-ml_%s:%s", scala_version, sw_version),
+    sprintf("ai.h2o:sparkling-water-repl_%s:%s", scala_version, sw_version)
   ))
```
navdeep-G: No change here. Why the diff?

navdeep-G:
This PR looks good to me. Can we get it merged before EOD? @jangorecki @mmalohlava ?

r/rsparkling/R/package.R
jangorecki: @navdeep-G only indenting change to be consistent

r/rsparkling/R/h2o_context.R
mmalohlava: In the context of this PR it is enough. With Jan we discussed different strategies.

PR #132, opened 2016-11-12
Add strict version check to h2o_flow()

navdeep-G:
* This PR adds a `strict_version_check` parameter to `h2o_flow`.
* This is needed to invoke H2O flow from a rsparkling instance if H2O versions do not match via the `h2o.init()` call.

### DIFF
--- a/r/rsparkling/R/h2o_context.R
+++ b/r/rsparkling/R/h2o_context.R
@@ -26,9 +26,10 @@ h2o_context.spark_jobj <- function(x, strict_version_check = TRUE) {
 #' @inheritParams h2o_context
 #'
 #' @param sc Object of type \code{spark_connection}.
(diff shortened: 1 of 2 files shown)


r/rsparkling/man/as_h2o_frame.Rd
```diff
@@ -1,4 +1,4 @@
-% Generated by roxygen2: do not edit by hand
+% Generated by roxygen2 (4.1.1): do not edit by hand
```
mmalohlava: Is this change necessary?

navdeep-G: Docs need to be updated since we added an extra parameter to `h2o_flow()`. I have re-updated the docs with the latest version of roxygen2.

navdeep-G:
@jangorecki We can squash with merge in Github

navdeep-G:
Okay to merge?

PR #139, opened 2016-11-22
[BUILD] Gradle upgrade

mmalohlava:
The motivaiton for this change is to speed up build and execute basic building
tasks in parallel.

```bash
./gradlew --parallel clean build -x check scalaStyle
```

Minor updates:
  - removing deprecated `<<` construct for defining Task actions

TODO: sparkling-water-py:distPython needs to be change to support parallel build


gradle/itest.gradle
```diff
@@ -2,6 +2,8 @@
 // Integration tests support
 //
 
+
+// FIXME: this is not necessary anymore with new version of Idea
```
jakubhava: If not necessary, should we remove all occurrences of this code in Gradle code ?

jakubhava:
Regarding `:sparkling-water-py:distPython` what exactly there is blocking the parallel execution ? I studied that in the morning and it seems that project and task dependencies are fine for Gradle parallel execution. In `distPython` task we are accessing the compile configuration but it's part of the project dependency, or am I wrong ?

jakubhava:
And found a cool think - gradle python plugin https://engineering.linkedin.com/blog/2016/08/introducing--py-gradle--an-open-source-python-plugin-for-gradle

https://github.com/linkedin/pygradle/blob/master/docs/python.md

mmalohlava:
Regarding `distPython` - the problem is `doFirst` which is executed in parallel with actual task action. I tried to remove it, but it was leading to more problems. Still need to investigate that.

Regarding python plugin - looks great, but for our case not sure if it is useful - we are doing lot of SW-specific things - verification of Spark, preparation of environment, download of H2O package. In our case, the base plugin is pretty simple (just defines individual common stages)

PR #140, opened 2016-11-22
[MINOR] Add tags of sw and h2o for latest official sparkling-water re…

jakubhava:
…lease into readme

People who wants to build sparkling-water and h2o manually need to ensure they are using the same sparkling-water and h2o version ( for example in external cluster ). These links points to sparkling-water release tag and corresponding h2o release tag.

This was small change so I decided to make a pull request first and then decide whether we approve or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,15 @@ pip install future
 > To build only a specific module, use, for example, `./gradlew :sparkling-water-examples:build`.
 > To build and test a specific module, use, for example,  `./gradlew :sparkling-water-examples:check`.
 
+Developers who want to build sparkling water and also h2o manually need to ensure ensure they are using
+correct versions of both dependencies. 
+
+Here we provide links to tags of sparkling water latest official release and corresponding h2o official release.
+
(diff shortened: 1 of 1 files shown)


mmalohlava:
These are hardcoded value. We will forget to update them and it will make confusion.

jakubhava:
Good point. We could generate it automatically based on gradle.properties in sparkling water ( that's all we need ). We could have README.template, however before each push we would need to ensure that the script which generates the README.md would be run, which is another thing to remember.

Maybe it's not a good change at the end.

mmalohlava:
Thinking about it more:
  - release page already links directly Spark and H2O versions
  - the README.md file is a static file, hard to regenerate in repo (i do not know if GitHub provide some kind of dynamicity here)
  - with each release we can push a generated RELEASE.md file describing relation ship between version - i would not care about master since in our case it is development branch.

What do you think?

jakubhava:
Yup, we can only make it for release branches. Actually as you said, it makes more sense 👍

jakubhava:
Since we have this PR https://github.com/h2oai/sparkling-water/pull/143, I think that this change is no longer required, can I close it ? Because of the previous PR the user can say `-PdownloadAndExtend` and it will download correct h2o version automatically

The added value of this change would be really tiny.

PR #141, opened 2016-11-23
[MINOR][SCALASTYLE] fix scala style warning

jakubhava:
This is ugly change, but doesn't change the semantic of the code at all, it just fixes a lot of scala style warnings. It started to be annoying to see a lot of warnings every time we build the project.


mmalohlava:
Let's postpone this change little bit after release and porting all changes into rel-1.6.

jakubhava:
Yup, no hurry for this change 👍

jakubhava:
I was thinking more about this change. It doesn't seem to be smart to change all the spacing one comment at one to me now.

I was thinking whether it's possible to set some kind of policy which would require to change all spacing in a file when that file has been changed because of other changes. Or anything else similar so we are force to change this at least incrementally.

What do you think @mmalohlava @mdymczyk ? Should I close this PR or are we planning to use it at some point ?

mmalohlava:
@jakubhava good point - we should run check style on each change and remove all warnings on changed files. Thx for idea!

PR #142, opened 2016-11-23
[SW-266] H2OContext should not implement Serializable

jakubhava:
H2OContext now doesn't implement `Serializable`, but we need to make `ConverterUtils` `Serializable` because inner class generated by scala to java conversion  references instance of `SparkDataFrameConverter`, `PrimiveRDDConverter` and so on in outer scope.

There is a commit in https://github.com/h2oai/sparkling-water/pull/87 removing `ConverterUtils` and splitting the code in in into `ReadConverterCtxUtils` and `WriteConverterCtxUtils` for better readability and code separation. As a side effect it also removes the need of having `ConverterUtils `and related classes `Serializable`

So maybe we can wait, don't merge this PR and wait for PR-87 ? What do you think @mmalohlava ?

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -59,7 +59,7 @@ import scala.util.control.NoStackTrace
   * @param conf H2O configuration
   */
 class H2OContext private (@(transient @param @field) val sparkContext: SparkContext, @(transient @param @field) conf: H2OConf) extends Logging
-  with Serializable with SparkDataFrameConverter with SupportedRDDConverter with DatasetConverter
+  with SparkDataFrameConverter with SupportedRDDConverter with DatasetConverter
   with H2OContextUtils { self =>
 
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -59,7 +59,7 @@ import scala.util.control.NoStackTrace
   * @param conf H2O configuration
   */
 class H2OContext private (@(transient @param @field) val sparkContext: SparkContext, @(transient @param @field) conf: H2OConf) extends Logging
-  with Serializable with SparkDataFrameConverter with SupportedRDDConverter with DatasetConverter
```
mmalohlava: I am not sure about it - this will hide `Serializable` but H2OContext still remains `Serializable`.

jakubhava: You're right. This change doesn't do actually anything.

Why we need `ConverterUtils` to be `Serializable`
> We need to make `ConverterUtils` `Serializable` because inner class generated by scala to java conversion references instance of SparkDataFrameConverter, PrimiveRDDConverter and so on in outer scope.

I completely removed the need for the `Serializable` in  #87 by separating ConverterUtils and making it just helper objects instead of traits. If it's ok, I would close this PR and wait for #87 as extracting the change from the whole PR seems like a lengthy process. Is it ok with you ?

jakubhava:
This change will be brought in by #87

PR #143, opened 2016-11-24
Add support for extending H2O/H2ODrriver jar files by classes require…

jakubhava:
…d by sparkling water. The files can be downloaded which simplifies deployment

### DIFF
--- a/assembly-h2o/build.gradle
+++ b/assembly-h2o/build.gradle
@@ -1,0 +1,172 @@
+import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+
+import java.util.zip.ZipEntry
+import java.util.zip.ZipInputStream
+
+apply plugin: 'com.github.johnrengelman.shadow'
+apply plugin: 'base'
+
+description = """
+h2o and h2o Driver Assembly. This assembly extends provided jar by sparkling-water specific classes needed in external
+backend mode.
+
+User can specify location to either h2o or h2odriver jar via H2O_ORIGINAL_JAR env variable.
+No works is done if H2O_ORIGINAL_JAR is not set.
+
+User can also use gradle command line argument -PdownloadAndExtend which will download correct h2o version for current
+sparkling water from our repository if it hasn't been already downloaded.
+If this is run without any argument, then basic h2o jar is downloaded,
+but this command takes also argument specifying hadoop version, such as cdh5.4. In case the argument is specified, instead
+of downloading h2o jar this command will download h2o driver jar for specified hadoop version.
+
+Gradle property has higher priority if both H2O_ORIGINAL_JAR and -PdownloadAndExtend are set
+
+Few examples:
+
+1)
+export H2O_ORIGINAL_JAR = ...
+./gradlew build -x check
+
+In this case the jar to be extended is located using the provided environment variable
+
+2)
+./gradlew build -x check -PdownloadAndExtend
+
+In this case the jar to be extended is h2o jar and is downloaded from our repository first
+
+3)
+./gradlew build -x check -PdownloadAndExtend=cdh5.4
+
+In this case the jar to be extended is h2o driver jar for provided hadoop version and is downloaded from our repository first
+
+4)
+./gradlew build -x check -PdownloadAndExtend=abc
+
+This case will throw an exception since such hadoop version is not supported
+
+5)
+export H2O_ORIGINAL_JAR = ...
+./gradlew build -x check -PdownloadAndExtend
+
+This version will ignore environment variable and jar to be extended will be downloaded from our repository. The same
+holds for version with argument.
+"""
+
+configurations {
+    artifactsToMerge
+}
+
+def originalJar = getOrDownloadOriginalJar()
+
+dependencies {
+    artifactsToMerge("ai.h2o:h2o-scala_${scalaBaseVersion}:${h2oVersion}") { transitive = false }
+    artifactsToMerge("org.scala-lang:scala-library:${scalaVersion}")
+    originalJar = getOrDownloadOriginalJar()
+    if (originalJar != null) {
+        artifactsToMerge files(originalJar)
+    }
+}
+
+task extendJar(type: ShadowJar) {
+    configurations = [project.configurations.artifactsToMerge]
+    if (originalJar != null) {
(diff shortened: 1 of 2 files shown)


h2o-assembly/build.gradle
```diff
@@ -0,0 +1,137 @@
...
+
+Gradle property has higher priority if both H2O_ORIGINAL_JAR and -PdownloadAndExtend are set
+
+"""
+
+configurations{
+  artifactsToMerge
+}
+
+ext.supportedHadoopVersions = ["cdh5.2","cdh5.3", "cdh5.4", "cdh5.5", "cdh5.6", "cdh5.7", "cdh5.8", "hdp2.1",
```
mmalohlava: You do not need to list them here, since it is hard to synchronize the list with actual list in H2O. Just keep it simple - if user specify wrong version, the script reports error

h2o-assembly/build.gradle
```diff
@@ -0,0 +1,137 @@
...
+No works is done if H2O_ORIGINAL_JAR is not set.
+
+User can also use gradle command line argument -PdownloadAndExtend which will download correct h2o version for current
+sparkling water from our repository if it hasn't been already downloaded.
+If this is run without any argument, then basic H2O.jar is downloaded,
+but this command takes also argument specifying hadoop version, such as cdh5.4. In case the argument is specified, instead
+of downloading H2O.JAR this command will download h2o driver jar for specified hadoop version.
+
+Gradle property has higher priority if both H2O_ORIGINAL_JAR and -PdownloadAndExtend are set
+
```
mmalohlava: Example in description would be helpful.

h2o-assembly/build.gradle
```diff
@@ -0,0 +1,137 @@
...
+description = """
+H2O and H2O Driver Assembly. This assembly extends provided jar by sparkling-water specific classes needed in external
+backend mode.
+
+User can specify location to either h2o or h2odriver jar via H2O_ORIGINAL_JAR env variable.
+No works is done if H2O_ORIGINAL_JAR is not set.
+
+User can also use gradle command line argument -PdownloadAndExtend which will download correct h2o version for current
+sparkling water from our repository if it hasn't been already downloaded.
+If this is run without any argument, then basic H2O.jar is downloaded,
```
mmalohlava: H2O.jar, H2O.JAR or h2o.jar ?

h2o-assembly/build.gradle
```diff
@@ -0,0 +1,137 @@
...
+      logger.info("\"H2O_ORIGINAL_JAR\" property set to $originalJar")
+      originalJar
+    }else{
+      return null
+    }
+  }
+}
+
+def String downloadAndUnpackH2OJar(String urlStr, String fileName, String matchPattern){
+  def saveDir = new File("private")
```
mmalohlava: Keep it in build directory.

mmalohlava: file("$buildDir/private")

settings.gradle
```diff
@@ -5,6 +5,7 @@ include 'examples'
 // Disable for now
 include 'ml'
 include 'assembly'
+include 'h2o-assembly'
```
mmalohlava: keep the same prefix please: `assembly-h2o`. It is more handy since if i am looking for that i can type assembly and shell will offer me alternatives. Think about it as hierarchical space...

h2o-assembly/build.gradle
jakubhava: Sorry, I will unify it. Thanks!

h2o-assembly/build.gradle
jakubhava: yup will do

h2o-assembly/build.gradle
jakubhava: good point. we can actually discover that - wrong version means that url won't lead to download h2o/h2odriver jar

settings.gradle
jakubhava: It makes sense, thanks!

jakubhava:
@mmalohlava All proposed changes are in. I also cleaned spacing and other minor things in the code to be consistent.

jakubhava:
Another change is in. I put also version to downloaded jars names. Since we cache them, it could happen that the user would update sparkling-water repo via `git pull`, `the h2o.jar` or `h2odriver.jar` won't be deleted ( unless the user would call `./gradlew clean`) and therefore h2o jars with wrong versions would be used. This is eliminated now and can't happen

assembly-h2o/build.gradle
```diff
@@ -0,0 +1,172 @@
...
+
+This version will ignore environment variable and jar to be extended will be downloaded from our repository. The same
+holds for version with argument.
+"""
+
+configurations {
+    artifactsToMerge
+}
+
+def originalJar = getOrDownloadOriginalJar()
```
mmalohlava: why it is here? and also in `ext`

jakubhava: What do you mean please ? I had a look and I'm not putting it in `ext` manually

I need to keep the `originalJar` at this scope for example because of this call
```
if (originalJar != null) {
    build.dependsOn extendJar
}
```

mmalohlava: ahh, oki, sorry my fault, i looked below and see `originalJar` in `dependencies` section and confused it with `ext`.

assembly-h2o/build.gradle
```diff
@@ -0,0 +1,172 @@
...
+ * Method determining whether the h2o or h2o driver jar is available.
+ * Location to the jar is returned if it's available, oterwise null is returned
+ *
+ * If the user specified Gradle argument downloadAndExtend then the method will try do download the jar and  return path to it
+ */
+def String getOrDownloadOriginalJar() {
+    if (project.hasProperty("downloadAndExtend")) {
+        def arg = project.downloadAndExtend
+        if (arg == null || arg.isEmpty()) {
+            def url = "http://h2o-release.s3.amazonaws.com/h2o/${h2oMajorName != "master" ? "rel-${h2oMajorName}" : "master"}/${h2oBuild}/h2o-${h2oMajorVersion}.${h2oBuild}.zip"
```
mmalohlava: I was looking into S3 storage: you do not need to download the `zip` file. use this one:

https://s3.amazonaws.com/h2o-release/h2o/rel-turin/4/Rjar/h2o.jar

PR #145, opened 2016-12-04
[SW-268] Relocation of Jetty classes into `ai.h2o` package

mmalohlava:
The H2O is using old version of Jetty (because of Java6) which is not compatible
with DBC Jetty (prepended on Spark classpath).

### DIFF
--- a/assembly/build.gradle
+++ b/assembly/build.gradle
@@ -24,6 +24,8 @@ jar {
 // Shadow jar settings
 shadowJar {
   relocate 'javassist', 'ai.h2o.javassist'
+  // Relocate Eclipse Jetty to avoid collisions with Spark Jetty (newer version)
+  relocate 'org.eclipse.jetty', 'ai.h2o.org.eclipse.jetty'
   dependencies {
     // This has to be specific list of dependencies
     // Do not forget that Sparkling Water is intended 


mmalohlava:
See also Jira https://0xdata.atlassian.net/browse/SW-268 for more info.

jakubhava:
So H2O still has Java 6 ? :(

mmalohlava:
Yup, no java 6 - no answer from Cran guys. However, this is temporary solution - we will get rid of it in SW-269.

tomkraljevic:
I don't know how important this is for sparkling water, but it probably means that anybody that was using a jetty security configuration for LDAP or Kerberos before won't work anymore, or at least not without a tweak.

Notice the class name is explicitly listed here below.

ldaploginmodule {
    org.eclipse.jetty.plus.jaas.spi.LdapLoginModule required
    debug="true"
    useLdaps="false"
    contextFactory="com.sun.jndi.ldap.LdapCtxFactory"
    hostname="ldap.0xdata.loc"
    port="389"
    bindDn="cn=admin,dc=0xdata,dc=loc"
    bindPassword="0xdata"
    authenticationMethod="simple"
    forceBindingLogin="true"
    userBaseDn="ou=users,dc=0xdata,dc=loc";
};

Adding Mark Chan, in case one of the customers he's working with runs in to an issue.

mmalohlava:
Good point, however, it should work since the class will be still available. But honestly, i really hate re-allocation, however, this is only way how to progress right now. There is SW-269 which is targeting revert of this change whenever CRAN guys confirms they installed java version > v6.

@mklechan please let me know if you explore any problems

mmalohlava:
But again, it is ridiculous that a single repo is forcing us to screw our code.

PR #146, opened 2016-12-04
Update README.md

niketdoshi:
Modified the repos URL. Now it correctly points to Turing build 7


jakubhava:
Thanks for reporting this @niketdoshi!

I think we could actually automatically generate these as part of our release script `make-dist.sh`, what do you think @mmalohlava ?

PR #147, opened 2016-12-05
[SW-267][FIX] Few fixes to task extending h2o jars

jakubhava:
- remove unnecessary semicolon
 - don't depend on file name but check if it's h2o driver jar file based on availability of class specific only to h2o driver jar
 - merge service files
 - Add `Main-Class` attribute to the extended jar


jakubhava:
I'm closing this PR and cherry-picking it on the branch with external cluster since it belongs there ( also the branch would be ready to merge any time afterwards and testing for me will be easier )

PR #148, opened 2016-12-08
[SW-272] Fix deployment issue of Sparkling Water on Azure/CDH

mmalohlava:
The problem was that `SparkContext#addJar` method improperly
register given path (representing `sparkling_water_assembly.jar` from egg file) in file server.
The problematic part of code:
```scala
if (master == "yarn" && deployMode == "cluster") {
  val fileName = new org.apache.hadoop.fs.Path(uri.getPath).getName()
  try {
    env.rpcEnv.fileServer.addJar(new File(fileName))
  } catch {
// ...
```

The code in fact register given file location under current directory (because of `new File(Path.getName())`).

Big warning - the fix use internal Spark API. For "yarn/cluster" mode it re-registers `sparkling_water_assembly.jar`
key under its original location provided by egg file.

Change:
  - Use abs path to point to resource

### DIFF
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -24,9 +24,9 @@ class Initializer(object):
         jvm = sc._jvm
         # Add Sparkling water assembly JAR to driver
         sw_jar_file = Initializer.__get_sw_jar()
-        url = sc._jvm.java.net.URL("file://{}".format(sw_jar_file))
+        url = jvm.java.net.URL("file://{}".format(sw_jar_file))
         # Assuming that context class loader is always instance of URLClassLoader ( which should be always true)
-        cl = sc._jvm.Thread.currentThread().getContextClassLoader()
+        cl = jvm.Thread.currentThread().getContextClassLoader()
 
         # explicitly check if we run on databricks cloud since there we must add the jar to the parent of context class loader
         if cl.getClass().getName()=='com.databricks.backend.daemon.driver.DriverLocal$DriverLocalClassLoader':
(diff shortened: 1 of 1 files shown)


mmalohlava:
Please look at it - this is kind of hack, and issue should be already fixed by SPARK-18160 (https://github.com/apache/spark/commit/3c24299b71e23e159edbb972347b13430f92a465)

py/pysparkling/initializer.py
```diff
@@ -18,8 +20,11 @@ def load_sparkling_jar(spark_context):
 
     @staticmethod
     def __add_sparkling_jar_to_spark(sc):
+        jsc = sc._jsc
+        jvm = sc._jvm
         # Add Sparkling water assembly JAR to driver
-        url = sc._jvm.java.net.URL("file://"+Initializer.__get_sw_jar())
+        sw_jar_file = Initializer.__get_sw_jar()
+        url = sc._jvm.java.net.URL("file://{}".format(sw_jar_file))
```
jakubhava: `jvm.java.net.URL("file://{}".format(sw_jar_file))` would be shorter since we have already defined `jvm` above it

py/pysparkling/initializer.py
```diff
@@ -18,8 +20,11 @@ def load_sparkling_jar(spark_context):
...
         # Assuming that context class loader is always instance of URLClassLoader ( which should be always true)
         cl = sc._jvm.Thread.currentThread().getContextClassLoader()
```
jakubhava: `sc._jvm` can be replaces by `jvm`

py/pysparkling/initializer.py
```diff
@@ -30,9 +35,21 @@ def __add_sparkling_jar_to_spark(sc):
             cl.addURL(url)
 
         # Add Sparkling water assembly JAR to executors
```
jakubhava: I wrote this comment when writing this code, but I would change it to something like this:

> Add Sparkling Water Assembly JAR to Spark's file server so executors can fetch it when they need to use the dependency.

because we aren't actually adding any jars to executors at this phase

py/pysparkling/initializer.py
```diff
@@ -30,9 +35,21 @@ def __add_sparkling_jar_to_spark(sc):
...
-        sc._jsc.addJar(Initializer.__get_sw_jar())
+        jsc.addJar(sw_jar_file)
+
+        # SW-272: For yarn and cluster deployments we need to modify location the jar file to a driver cache location
```
jakubhava: tiny thing - location **of** the jar

PR #151, opened 2016-12-12
[SW-276] Fix classloading problem of classes not defined in repl when…

jakubhava:
… repl is enabled

Explanation is here https://0xdata.atlassian.net/browse/SW-276


mmalohlava:
Need to be tested on Databricks and Azure environments.

mmalohlava:
And with Hive.

mmalohlava:
Sorry, i accidentally closed the PR.

jakubhava:
Yup, I'm now running yarn and script tests in jenkins. Already tested in DBC and fix works as expected

mmalohlava:
We need more automatic tests for that - DBC, Azure. No manual testing...it is too error prone.

jakubhava:
@mmalohlava agree!

PR #152, opened 2016-12-12
[SW-263] Allow running build in parallel

jakubhava:
Only task and project dependencies are allowed for projects to be decoupled. We were using configuration dependency

### DIFF
--- a/py/build.gradle
+++ b/py/build.gradle
@@ -11,9 +11,8 @@ ext {
 description = "PySparkling - Sparkling-Water Python Package"
 
 dependencies {
-  compile project(path: ':sparkling-water-assembly', configuration: 'shadow')
+    compile project(path: ':sparkling-water-assembly')
 }
-
 //
 // Create a file with version for Python dist task
 //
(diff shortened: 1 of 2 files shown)


py/build.gradle
```diff
@@ -111,8 +107,8 @@ task distPython(type: Exec, dependsOn: checkPythonEnv) {
         }
 
         copy {
-            from "${configurations.compile.join(',')}"
```
mmalohlava: I would prefer to keep declarative dependency ...

jakubhava: me to, but why I decided to do this: Because we can depend on specific configuration anymore, we depend on all artefacts produced by `:sparkling-water-assembly`. I can use them ( there's around 190 artefacts ) and filter for sparkling-water assembly jar. It seems cleaner to me to declare it like this then filtering almost 200 hundred artefacts. But I can the the filtering, it actually worked for me too.

jakubhava: Should I do the second approach anyway ?

jakubhava: So actually the filtering won't work. If we declare dependency as `compile project(:sparkling-water-assembly` then the configuration would contain all jars but sparkling-water-all jars ( the output of shadowJar is missing there ) So nicest solution can be to change sparkling-water-assemblt to contain the all jar and do the filter or just ensure that it will produce only all file as artifact )

jakubhava: So last update: not even this is possible, if we specify the project as dependency the configuration contains complete source set ( before when just shadow configuration was specified as a dependency, it contained just output of that task thus only desired jar file.)

We can't use the the jars provided by this source set since it wouldn't contain changes introduced by shadow jar such as the relocation of classes
So i don't know of any better solution without coupling the projects again. By accessing the configuration of different project we couple the projects. So this seems like the simplest and single functional solution I know of

jakubhava: @mmalohlava please any ideas on this ? It would be good to make the build parallel together with your PR https://github.com/h2oai/sparkling-water/pull/195, it would move PySparkling build  a bit forward again!

mmalohlava: PR 195 has higher priority. Build time is not bottle neck right now. Keep this change opened.

jakubhava: oki 👍

PR #161, opened 2016-12-22
[PUBDEV-3808] Cache Spark's dataframe before we do any conversion. Be…

jakubhava:
…haviour without caching can be nondeterministic. Seems like a Spark bug, but for now caching the dataframe does the work

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/converters/DatasetConverter.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/converters/DatasetConverter.scala
@@ -30,6 +30,7 @@ private[h2o] object DatasetConverter extends Logging {
 
   /** Transform Spark's Dataset into H2O Frame */
   def toH2OFrame[T <: Product](hc: H2OContext, ds: Dataset[T], frameKeyName: Option[String])(implicit ttag:TypeTag[T]) = {
+    ds.cache()
     val tpe = ttag.tpe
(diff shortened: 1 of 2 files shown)


mdymczyk:
Cool, I'll run some test in a few hours, I'm just curious to why only some columns were swapped - I find it very weird, have you found out a reason for that @jakubhava ?

jakubhava:
Not yet. It really feels like a bug in Spark which occurs on some specific
data. Maybe the nulls are causing it. I want to report it to Spark
community and see if they can explain or file up jira for it.

mmalohlava:
Agree, we need to know why columns are swapped. Caching data is good-enough hotfix, but it means doubling memory.

KristinaPlazonic:
Hi, original reporter here. Caching the dataframe doesn't seem to be the correct fix, as it spends resources and changing the h2o build from 5.10.0.10 to 5.10.1.2 seems to have fixed the issue for us (rather than this pull) for Spark 2.0 and 1.5 (Sparkling Water 1.5.16). Please see the [JIRA comment](https://0xdata.atlassian.net/browse/PUBDEV-3808?focusedCommentId=40703&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-40703)

mmalohlava:
I am going to close this PR, since we have slightly different workaround.

PR #165, opened 2017-01-04
[SW-XXX] DeepWater demo in SW

mdymczyk:
My idea is to implement this http://www.wildml.com/2015/12/implementing-a-cnn-for-text-classification-in-tensorflow/ with a twist (use Word2Vec for word embeddings) so it's not just a copy paste of his Python code. It's a sentiment analysis CNN learning the sentiments based on the Rotten Tomatoes reviews.

Eventually we will have 2 demos:

* an ipython notebook with the whole pipeline: getting the data, cleaning it, making the word2vec vectors, building the TF architecture, saving it, running the whole thing in DeepWater

* a Scala script that will do the same as above except architecting the TF network, we'll use a metadata file created in the ipython notebook. The reason is simple - there's no Java/Scala API for TensorFlow


mdymczyk:
Note to self: Spark puts Protobuf-Java 2.5.0 on the classpath, we need 3.1.0+ for DeepWater's TF - need to figure out how to do this.

Note2: need to add a better way to add `deepwater-all.jar` to `--jars` in `sparkling-shell`

h2o-ops:
Hi @mdymczyk, your PR title "DeepWater demo in SW" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #166, opened 2017-01-11
Remove python version from egg so it works with 3.x

mdymczyk:
Until now we were only building pysparkling for 2.7 it seems? And the pysparkling shell had that version hardcoded - think it's about time to support also 3.5 (anyway I need this as I set up all my env in a Python3.5 conda env).

This isn't 100% ready as sometimes, while switching conda env's the `setup.py` part of the gradle build does not recognize the change, I have no idea why though as the `environment` variable seems to be properly set and when I try to run the `pythonVersion.py` script from that task it always prints the right version. Also when I tried rewriting that task from `Exec` task to using a simple `["python", otherArgs].execute()` it also always works but for that I would also need to rewrite the other python tasks - I'm ok with that but not sure if that's the way to go?

tl;dr seems like gradle's exec task has some problems with conda

### DIFF
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -54,11 +54,12 @@ FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
 
-python_version=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
-PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py${python_version}.egg"
-PY_EGG_FILE="$TOPDIR/py/build/dist/$PY_EGG"
-
-checkPyEggFile ${python_version}
+python_version=`../py/pythonVersion.sh`
+if [ ! -z $python_version ]; then
+    PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py${python_version}.egg"
+    PY_EGG_FILE="$TOPDIR/py/build/dist/$PY_EGG"
+    checkPyEggFile ${python_version}
+fi
 
 # Default master
 DEFAULT_MASTER="local[*]"
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -21,6 +21,12 @@ function getMasterArg() {
     done
 }
 
+function checkPyEggFile() {
+  if [ ! -f $PY_EGG_FILE ]; then
+    echo "WARN	Running pysparkling using Python $1 requires ${PY_EGG_FILE} egg file which does not exist. This might be because your project was built using a different Python version. Build the project using appropriate Python version or change your version."
+  fi
+}
+
 if [ -z $TOPDIR ]; then
   echo "Caller has to setup TOPDIR variable!"
   exit -1
@@ -48,9 +54,11 @@ FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
 
-PYTHON_VERSION=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
-PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py$PYTHON_VERSION.egg"
+python_version=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
+PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py${python_version}.egg"
 PY_EGG_FILE="$TOPDIR/py/build/dist/$PY_EGG"
+
+checkPyEggFile ${python_version}
 
 # Default master
 DEFAULT_MASTER="local[*]"
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -47,7 +47,9 @@ FAT_JAR="sparkling-water-assembly_$SCALA_VERSION-$VERSION-all.jar"
 FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
-PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py2.7.egg"
+
+PYTHON_VERSION=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
+PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py$PYTHON_VERSION.egg"
 PY_EGG_FILE="$TOPDIR/py/build/dist/$PY_EGG"
 
 # Default master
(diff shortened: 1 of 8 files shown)


bin/sparkling-env.sh
```diff
@@ -47,7 +47,9 @@ FAT_JAR="sparkling-water-assembly_$SCALA_VERSION-$VERSION-all.jar"
 FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
-PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py2.7.egg"
+
+PYTHON_VERSION=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
```
jakubhava: We should do this also on windows script equivalents to be consistent.

Maybe it would be nice to print an explanation what happened in case the script couldn't find desired python egg. The user could have changed the python version and I can imagine he/she would be confused when script which they have been using before doesn't work.

bin/sparkling-env.sh
```diff
@@ -47,7 +47,9 @@ FAT_JAR="sparkling-water-assembly_$SCALA_VERSION-$VERSION-all.jar"
...
+PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py$PYTHON_VERSION.egg"
```
jakubhava: Tiny thing - just to be consistent `${PYTHON_VERSION}`

mdymczyk: Done

bin/sparkling-env.sh
mdymczyk: @jakubhava We don't have a `pysparkling` script for Windows, though and we're not setting it (the egg file location) in the Windows version - should we? Was it not done before for any reason?

Added a warn message in the script, not failing the script though as some uses of it don't need PySparkling.

jakubhava:
I would say it wasn’t done because most of us and our users don’t use Windows and it wasn’t reported as missing features. I would just create new jira for it since it’s not urgent.

Thanks!

py/build.gradle
```diff
@@ -3,6 +3,13 @@ apply from: "$rootDir/gradle/utils.gradle"
 ext {
     FS = File.separator
     FPS = File.pathSeparator
+
+    def command = ["python", "py/pythonVersion.py"]
```
mmalohlava: Why do you need separated python file? You can invoke python code directly no?

I am asking since it is inconsistent with code above.

bin/sparkling-env.sh
```diff
@@ -47,9 +53,13 @@ FAT_JAR="sparkling-water-assembly_$SCALA_VERSION-$VERSION-all.jar"
 FAT_JAR_FILE="$TOPDIR/assembly/build/libs/$FAT_JAR"
 major_version=`echo $VERSION | cut -d . -f 1,2`
 version_without_snapshot=`echo $VERSION | cut -d - -f 1`
-PY_EGG="h2o_pysparkling_${major_version}-${version_without_snapshot}-py2.7.egg"
+
+python_version=`python -c 'import sys; print(".".join(str(x) for x in sys.version_info[0:2]))'`
```
mmalohlava: What if user does not have python and he/she is not interested in pysparkling. We should handle this situation.

jakubhava: good point with this one 👍

mmalohlava: Hint: https://gist.github.com/rishabhmhjn/3a51828ae2cd9c1f19bc

py/build.gradle
mdymczyk: Yes, I was having problems while invoking the contents of pythonVersion.py directly from the gradle script as above using the `-c` option, not sure if I was doing something wrong but I didn't manage to get it to work :-/

bin/sparkling-env.sh
mdymczyk: Good point, will change it,

py/build.gradle
jakubhava: the original approach in build.gradle didn't work as well ? 

```commandLine getOsSpecificCommandLine(["python", "setup.py", "egg_info", "...other...")```

Comment (mdymczyk) [inline py/build.gradle]: @jakubhava that one works fine, I was having problems with the `-c` option in my case.

Comment (jakubhava) [inline py/build.gradle]: yup, that works, but I mean if similar approach in your case `commandLine getOsSpecificCommandLine(["python", "c"])` didn't work as well ( not considering the stdout at the moment )

Comment (mdymczyk) [inline py/build.gradle]: yes that's what I meant, tried it and didn't work but maybe I passed the parameters incorrectly (i.e. "-c" and "(STUFF)" separately and it should be together as one string, maybe applied too many quotes etc) hm... Though I think I tried quite a fee combinations

Comment (jakubhava) [inline py/build.gradle]: ah oki, that's interesting though. Thanks for explanation!

**Comment by mdymczyk** [comment]
@mmalohlava I moved the python version check to the script, where I also added a check if python is installed

**Comment by mmalohlava** [comment]
@mdymczyk Thinking more about this change, i do not understand it. What the scope of distribution - if i download sparkling water from webpage, it will contain version 2.7. So it will not be detected by launch script. 

Do we need to encode the python version inside the name of egg file?
Furthermore, the PR says "Remove python version from egg" - however the diff shows: 
```
eggs file("${project.buildDir}/dist/h2o_pysparkling_${version.substring(0, version.lastIndexOf('.'))}-${version.replace("-SNAPSHOT","")}-py${PYTHON_VERSION}.egg").getAbsoluteFile()
```

Could we remove Python version from egg filename?

mdymczyk:
@mmalohlava the name of the `egg` file is automatically set by the method generating it and it automatically adds the `-py2.7` or `-py3.5` suffix depending on which Python version was added. Whether or not we can override this behaviour to remove the suffix is a good question, haven't looked into it. Maybe @jakubhava knows? So if you run `./gradlew clean build` from a terminal which has Python set to 3.5 and then try to run `pysparkling` it won't work because the script has hardcoded `2.7` and the egg file will have the suffix `3.5`.

The line `eggs file("${project.buildDir}/dist/h2o_pysparkling_${version.substring(0, version.lastIndexOf('.'))}-${version.replace("-SNAPSHOT","")}-py${PYTHON_VERSION}.egg").getAbsoluteFile()` does not decide the `egg` name, it just adds the already generated `egg` file to the `eggs` custom gradle configuration which we are using in python tests.

tl;dr: I agree that removing the `py` suffix is a good idea (as long as we plan on making both versions compatible) but I'm not sure how to; I agree, it shouldn't say "removing" but "setting the proper python version for the egg file dependency".

mmalohlava:
My main problem:  what to put into binary
  - if we build SW with python 2.7 => users with Python 3.5 are screwed 
  - and vice versa:  build with 3.5 => 2.7 users have a problem

mmalohlava:
oki, spoke with Pasha - it seems we need to build universal egg file

mdymczyk:
@mmalohlava good point regarding the package. Did @st-pasha say how to do that? I always thought that egg files are Python version specific and you cannot make one for both Python 2 and Python 3 even if the codebase is the same?

mmalohlava:
hmm, it seems that egg is Python version specific since it includes `pyc` files. And we cannot use Wheel since Spark does not support them and we cannot use source distribution since it cannot contain binary files. Any ideas?

mdymczyk:
@mmalohlava http://spark.apache.org/docs/latest/api/python/pyspark.html#pyspark.SparkContext.addPyFile think it would work if we just zipped, instead of egg, our files and distribute the zip?

If not the distributing multiple eggs would be another, more annoying, option.

jakubhava:
So this PR seems like to be replaced by https://github.com/h2oai/sparkling-water/pull/195 ? Can we close it but still keep the communication here as reference ?

PR #167, opened 2017-01-11
Adds zip64 to assembly-h2o

mdymczyk:
Running `./gradlew extendJar -x check -PdownloadH2O=cdh5.4` without that option fails because of our zip being too large.

### DIFF
--- a/assembly-h2o/build.gradle
+++ b/assembly-h2o/build.gradle
@@ -95,6 +95,7 @@ dependencies {
 }
 
 task extendJar(type: ShadowJar, dependsOn:[":sparkling-water-ml:jar", ":sparkling-water-core:jar"]) {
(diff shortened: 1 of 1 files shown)


mdymczyk:
@jakubhava do you have any particular tests in mind (jenkins?) or just manual test?

jakubhava:
Just manual tests would be fine. I would just try to build it and run tests locally on it. Just to be little sure the assembly is correct with `zip64`. It don't see what could be wrong, but just to be sure

mdymczyk:
@jakubhava ah ok, I was using it (jar built with this config) to test the bug report we got recently and it seemed to work ok.

jakubhava:
cool! 👍

PR #168, opened 2017-01-12
Fix extended SW uber jar

mdymczyk:
On case insensitive systems (such as MacOS X) there's a conflict when sending the `h2odriver_extended.jar` to Hadoop: `META-INF/LICENSE` (a license file) and `META-INF/license` (a folder containing some licenses). It makes the whole job fail to run.

Other solution would be to rename `license` folder to `licenses` but the uber jar also contains `META-INF/LICENSE.txt` which seems like the same content (Apache 2.0 license).

Similar, old, issue in Mahout for reference: https://issues.apache.org/jira/browse/MAHOUT-780


jakubhava:
We just need to check and run tests on newly created jar file. It should contain all the required classes, but better to test it.

PR #170, opened 2017-01-15
[FIXES] SW-303 adds support for Spark's DecimalType

mdymczyk:
Basic `DecimalType` support since Spark 2.x seems to use it in places where Spark 1.6 used `DoubleType`.


mmalohlava:
Thx @mdymczyk ! Looks good!

PR #171, opened 2017-01-17
[SW-305] Support of java.sql.Date

mmalohlava:


mmalohlava:
I am not happy about allocation in external cluster, we can remove it later.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
```diff
@@ -55,6 +55,7 @@ class ExternalWriteConverterCtx(nodeDesc: NodeDesc, totalNumOfRows: Int) extends
   override def put(colIdx: Int, data: Float) = externalFrameWriter.sendFloat(data)
   override def put(colIdx: Int, data: Double) = externalFrameWriter.sendDouble(data)
   override def put(colIdx: Int, data: java.sql.Timestamp) = externalFrameWriter.sendTimestamp(data)
+  override def put(colIdx: Int, data: java.sql.Date) = externalFrameWriter.sendTimestamp(new java.sql.Timestamp(data.getTime))
```
jakubhava: To get rid of the allocation we could do 
`
override def put(colIdx: Int, data: java.sql.Date) = externalFrameWriter.sendLong(data.getTime)
` 
because in writer context on h2o side we handle timestamp in the same way as the rest of numbers

jakubhava:
btw: Thanks for looking into this!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
mmalohlava: Good point, in the best case, i would add a new method `sendDate` but sending long is good workaround.

mmalohlava: Thx Kuba!

jakubhava: Yup, `sendDate` would be the best, we can comment it properly here and meanwhile add this method on the h2o side

PR #172, opened 2017-01-17
[SW-310][SW-303] Unit test to verify correct behavior

mmalohlava:


mmalohlava:
I would keep both tests since they were reported as separated bugs. As a proof that it is fixed.

Issue #127, opened 2016-11-07
Sparkling Water on CDH 5.9's Spark 1.6


instanceofme:
Hi,

On Sparkling Water 1.6.8 & CDH 5.9's Spark 1.6, when creating a H2OContext, I get the folowing, which I could not reproduce with other versions of CDH or Spark:

```
java.lang.AbstractMethodError
     at org.apache.spark.Logging$class.log(Logging.scala:50)
     at org.apache.spark.h2o.backends.internal.InternalH2OBackend.log(InternalH2OBackend.scala:31)
     at org.apache.spark.Logging$class.logWarning(Logging.scala:70)
     at org.apache.spark.h2o.backends.internal.InternalH2OBackend.logWarning(InternalH2OBackend.scala:31)
     at org.apache.spark.h2o.backends.SharedBackendUtils$class.checkAndUpdateConf(SharedBackendUtils.scala:54)
     at org.apache.spark.h2o.backends.internal.InternalH2OBackend.checkAndUpdateConf(InternalH2OBackend.scala:40)
     at org.apache.spark.h2o.H2OContext.<init>(H2OContext.scala:83)
     at org.apache.spark.h2o.H2OContext$.getOrCreate(H2OContext.scala:262)
     at org.apache.spark.h2o.H2OContext$.getOrCreate(H2OContext.scala:277)
```

This reminds me of [SW-176](https://0xdata.atlassian.net/browse/SW-176), where CDH 5.7's Spark 1.6.0 had an additional method on some class (compared to Spark's official source code).

Smalinuxer:
same problem

jakubhava:
Sorry for very delayed response. Thanks for the bug report. We'll have a look on it and let you know 👍

edgararuiz:
Hi @jakubhava, any updates on this bug?

mdymczyk:
@edgararuiz I think I found a problem and pushed a PR https://github.com/h2oai/sparkling-water/pull/174 We need to test it against an actual CDH 5.9.1 cluster yet, though.

mdymczyk:
Hey guys, sorry for the delay - I had a look at this issue and it seems like the Spark code used in CDH 5.9 changes some some internals because of which apps compiled against vanilla Spark 1.6 won't work... I'm working on a fix but don't have an ETA yet, sorry.

mdymczyk:
[...]

PR #174, opened 2017-01-23
Issue 127 (CDH 5.9) fix

mdymczyk:
Stop the class server only if present - Spark 1.6.x in CDH 5.9 does not use it.

Still needs to be tested on an actual CDH cluster. I was able to bootstrap SparklingWaterDriver with Spark 1.6.0-CDH5.9.1 as dependency, though.

Edit: this is still a relevant fix (for CDH 5.8+) but doesn't fix the original issue #127


mdymczyk:
@jakubhava true, I wanted to just get it to work, completely forgot we can simply pass `repl.enabled=false`. Closing.

PR #175, opened 2017-01-24
Adds missing method to Spark's 1.6.2 Logging required by CDH

mdymczyk:
CDH 5.9 rewrote the Logging class used by Spark (and us), changing some method signature, this PR adds those methods to a new trait and uses the new trait for logging instead.

### DIFF
--- a/repl/src/main/scala/org/apache/spark/h2o/H2OLogging.scala
+++ b/repl/src/main/scala/org/apache/spark/h2o/H2OLogging.scala
@@ -1,0 +1,99 @@
+/*
+* Licensed to the Apache Software Foundation (ASF) under one or more
+* contributor license agreements.  See the NOTICE file distributed with
+* this work for additional information regarding copyright ownership.
+* The ASF licenses this file to You under the Apache License, Version 2.0
+* (the "License"); you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+package org.apache.spark.h2o
+
+import org.apache.log4j.{Level, LogManager, PropertyConfigurator}
+import org.apache.spark.Logging
+import org.apache.spark.util.Utils
+import org.slf4j.impl.StaticLoggerBinder
+
+/**
+  * This is a copy/paste of https://github.com/cloudera/spark/blob/master/core/src/main/scala/org/apache/spark/Logging.scala
+  * Cloudera's 1.6.0-cdh5.9 Spark changed some of the Logging methods to have isInterpreter as an argument, without them if
+  * we compile against vanilla Spark 1.6.0 we'll get runtime AbstractMethodException exceptions.
+  */
+trait H2OLogging extends Logging {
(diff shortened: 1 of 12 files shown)


repl/src/main/scala/org/apache/spark/h2o/H2OLogging.scala
```diff
@@ -0,0 +1,94 @@
...
+* limitations under the License.
+*/
+package org.apache.spark.h2o
+
+import org.apache.log4j.{Level, LogManager, PropertyConfigurator}
+import org.apache.spark.Logging
+import org.apache.spark.util.Utils
+import org.slf4j.impl.StaticLoggerBinder
+
+trait H2OLogging extends Logging {
```
jakubhava: It seems like this class is based on https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/internal/Logging.scala. I would mention it in the code why it needs to be copy-pasted and also would mention the origin of the code, for the future tracking.

mdymczyk: @jakubhava to be more precise this class is "borrowed" from https://github.com/cloudera/spark/blob/master/core/src/main/scala/org/apache/spark/Logging.scala :-) I'm simply adding the methods they are using here so the JVM doesn't complain at runtime.

jakubhava: yup, make sense :) I would just mention from where it is and why it's used as the comment in the code

mmalohlava:
Looks good, lets put it on the top of update rel-1.6 branch [PR-178]

mmalohlava:
Just thinking more - are we able to make a Spark independent release? A code which would configure itself based on detected Spark version (1.6 + CDHx.y.z)

mdymczyk:
@mmalohlava that's an interesting idea, especially if we could do it for all spark version, different 1.6 and 2.0 in the same repo, but in that case the codebase might get a bit hectic? Depending on how we do it I guess. Got already any ideas how the implementation would look?

Doing it only "per branch" - i.e. different code for different 1.6 versions, I don't know - might not be efficient if there would be only two versions 1.6 and CDH 1.6.

jakubhava:
@mmalohlava @mdymczyk maybe we could at least start by creating somethling like internal API - only for our purposes separating specific spark implementations by more general pluggable API. Not sure if this would be possible everywhere, but for example with the `Logging` it would be easy. Not sure if this would be achieavable in easu way with the repl, but I guess it would be doable too. Especially with the repl, it would safe us a lot of time.

Porting REPL stuff from 2.0 to 1.6 and between different cdh versions is hell and just easy to introduce new bugs. Having one REPL api doing this inside would hide this complexity and we could manage at only one place. 

Just ideas though. Once I have after the exams I could have a look on it, especially with the repl if that would be something we want.

mmalohlava:
Yup, both ways they have different pros/cons - my motivation is coming from Zeppelin Spark support which is doing that. They have an independent layer like Kuba is proposing. But right now, i am not sure if it is scalable idea. Maybe the current approach is good enough....

mdymczyk:
@mmalohlava ok I'll have a look at how Zeppelin is doing it, if they came up with a decent idea how to have code for both Spark 1.6 and 2.0 in the same branch that would be amazing since it would require (and make it easier) for people implementing new features to add them to both branches at the same time instead of implementing for 2.0 and then trying to cherrypick and adjust for 1.6.

mmalohlava:
Good point @mdymczyk - 1.6 is not progressing, so probably we do not need to care a lot about it in our code base. On the other hand, we could need to back-port new changes from SW upstream (like support to schema transformations).... but supporting multiple Spark version in one single tree seems like too much work.

At least current approach is developer friendly (clean code targeting a single Spark version), we do not need to care about supporting backward compatibility.

PR #176, opened 2017-01-24
[SW-314] Adds SSL documentation

mdymczyk:


DEVEL.md
```diff
@@ -304,6 +306,37 @@ H2O resides in the same executor JVM as Spark. The memory provided for H2O is co
...
+
+Both Spark and H2O support basic node authentication and data encryption. In H2O's case we encrypt all the data sent between server nodes and between client
+and server nodes. This feature does not support H2O's UDP feature, only data sent via TCP is encrypted.
+
+Currently only encryption based on Java's key pair is supported (more in-depth explanation can be found in H2O's documentation linked below).
+
+To enable security for Spark methods please check [their documentation](http://spark.apache.org/docs/latest/security.html).
+
+Security for data exchanged between H2O instances can be enabled manually by generating all necessary files and distributing them to all worker nodes as
+described in [H2O's documentation](https://github.com/h2oai/h2o-3/blob/master/h2o-docs/src/product/security.rst#ssl-internode-security) and passing the "spark.ext.h2o
```
mmalohlava: Is there better way to do that? Can we use Spark to do that?

mdymczyk: Yes, there is an automatic way described below you just call `Security.enableSSL(sc)` and all the files will be generated automatically and distributed via spark/yarn methods. This part is for people who, for some reason, want to do it manually.

mmalohlava: :+1:

PR #177, opened 2017-01-25
[SW-316] Encryption support to newly created H2O external clusters

mdymczyk:
This this is the only change needed in SW as the actual encryption/decryption is done on the H2O core side. This will need an H2O core version bump to work properly (with actual encryption, as it is now it should still work but the data between Spark nodes and H2O nodes won't be encrypted).

EDIT: this PR needs changes from https://github.com/h2oai/h2o-3/pull/727 to be fully functional, without them it will be working fine except for external SSL mode.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
@@ -40,20 +40,20 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
 
   def launchH2OOnYarn(conf: H2OConf): String = {
 
-    var cmdToLaunch = Seq[String]("hadoop")
+    var cmdToLaunch = Seq[String]("hadoop",
+      "jar", conf.h2oDriverPath.get)
 
     conf.sslConf match {
       case Some(ssl) =>
-        cmdToLaunch ++ Array("-internal_security", ssl)
         val sslConfig = new Properties()
         sslConfig.load(new FileInputStream(ssl))
         cmdToLaunch = cmdToLaunch ++ Array("-files", sslConfig.get("h2o_ssl_jks_internal") + "," + sslConfig.get("h2o_ssl_jts"))
+        cmdToLaunch = cmdToLaunch ++ Array("-internal_security", ssl)
+        logInfo(s"Running external cluster in encrypted mode with config $ssl")
       case _ =>
     }
 
     cmdToLaunch = cmdToLaunch ++ Seq[String](
-      "jar",
-      conf.h2oDriverPath.get,
       conf.YARNQueue.map("-Dmapreduce.job.queuename=" + _ ).getOrElse(""),
       "-nodes", conf.numOfExternalH2ONodes.get,
       "-notify", conf.clusterInfoFile.get,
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -51,6 +52,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
       "-disown"
     )
 
+    conf.sslConf match {
+      case Some(ssl) => {
+        cmdToLaunch ++ Array("-internal_security_config", ssl)
```
jakubhava: This would ensure that just communication is secured between h2o cluster which we are starting on yarn and sparkling water is secured.

We need to also handle the case where the h2o cluster is already started ( locally, anywhere ) and we are connection to it from sparkling water. But it seems like any h2o cluster started with the configuration bellow should use secure communication, right ? I'm just making sure here all cases are covered

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -51,6 +52,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
+        val sslConfig = new Properties()
+        sslConfig.load(new FileInputStream(ssl))
+        cmdToLaunch ++ Array("-files", sslConfig.get("h2o_ssl_jks_internal") + "," + sslConfig.get("h2o_ssl_jts"))
```
jakubhava: Can we add hadoop argument ( `-files`) after bunch of application specific arguments ?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
mdymczyk: Yes, when connecting to an already running H2O cluster the user will have to pass `conf --spark.ext.h2o.internal_security_config=<conf_file>` option pointing to the conf file, it should work out of the box after we merge the change I pushed to H2O core. 

The problem is that in such case the user can't use our utility class which generates the config file and key stores as it has to be the same keystore as the one used in H2O.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
mdymczyk: ah good point

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: Ah I see. So basically users need to copy the config file to node with sparkling water ?

I think we can just comment this behaviour for now and explain what needs to be done if user wants to connect to established h2o cluster. I expect anyway that most of the users will be using external sparkling water with h2o started on yarn automatically, as it is just way easier than having to configure and start your own h2o cluster 👍

mdymczyk: Yes, basically if the H2O cluster is already running in secure mode then it's using some key stores. Those key stores have to be used also in SW instances (the External Reader and Writer) so the only option is for the user to manually provide them. We could make an endpoint in H2O fetching them but then again fetching key stores over unsecured channels misses the point.

So it's either you go auto or copy the files yourself. This is also how encryption in H2O core works if we're not running on top of hadoop.

jakubhava: I would keep it simple and wouldn't create another extra endpoint for this. As you pointed out, it would be unsecured anyway. We can just mention that this is expected behaviour for this deployment strategy and that the solution with h2o on yarn can handle in in more native way

mdymczyk:
I have tested this feature with the appropriate H2O changes and all seemed to work fine.

PR #181, opened 2017-01-30
[SW-317] Upgrade of H2O dependency to 3.10.3.2

mmalohlava:


mdymczyk:
I see some SVM tests are failing, want me to have a look @mmalohlava ?

PR #183, opened 2017-02-04
[SW-325] A generic announcement subsystem

mmalohlava:
The fix includes:
  - announcement service to notify about SW events
  - announcement provider API which should be implemented to provide notification mechanism
  - REST API based notification


mmalohlava:
We need this for Azure integration.

PR #185, opened 2017-02-07
[SW-327] Attach a Sparkling Water tab into Spark UI

mmalohlava:
The tab is attached after start of H2OContext.
The tab contains runtime info about Sparkling Water (for example, IP of Flow UI),
but also H2O build information.


mmalohlava:
image

PR #198, opened 2017-03-06
[PUBDEV-3808] Mention the recommendation in the documentation

jakubhava:


README.md
```diff
@@ -404,3 +404,9 @@ Follow our [H2O Stream](https://groups.google.com/forum/#!forum/h2ostream).
     import _root_.hex.tree.gbm.GBM
     ```
   
+* Getting non-deterministic H2O Frames after the Spark Data Frame to H2O Frame conversion.
```
mmalohlava: Can we explain it little bit based on the comments in https://0xdata.atlassian.net/browse/PUBDEV-3808

jakubhava: Sure!

jakubhava:
There is 206 files since 204 out of those are part of the parquet file required for the test

jakubhava:
Tests work similarly to the code snippet you posted, but also in that case we need to have access to that parquet file - `val df = sqlContext.read.parquet("/tmp/reproducible_dataframe_h2o_bug.parquet/").repartition(1).select("id", "strfeat0")`

README.md
mmalohlava: One more thing, could we also mention Spark bug here? Just for tracking

jakubhava: I would mention there this one https://issues.apache.org/jira/browse/SPARK-17806, but just  just as related issue since  since I was still able to reproduce on Spark's 2.1.1 snapshot.

jakubhava:
Regarding the whole parquet file - I agree it would be good to find even smaller repro where we don't need to use any external file. I could spend some time on it, but I would rather file up JIRA for it and continue on other issues.

jakubhava:
All suggestions should be in, can I merge ? @mmalohlava

jakubhava:
Passed tests in Jenkins

PR #200, opened 2017-03-07
[SW-345] Document Sparkling Water external backend mode

jakubhava:


jakubhava:
@mmalohlava good idea, done!

mmalohlava:
My goal is to have a documentation similar to H2O.ai documentation or this one: http://text-io.beryx.org/releases/latest/
(look into github - they are using asciidoctor to generate the docs)

jakubhava:
That's good idea. There is still the hanging PR from @angela0xdata which should improve Sparkling Water docs, but it's few months old.

PR #209, opened 2017-03-16
[SW-350] Fixes Windows shell scripts problems

mmalohlava:
Fixes: 
  SW-350
  SW-349

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,13 +121,6 @@ There are several ways of using Sparkling Water:
 
 ---
 
-## Sparkling Water cluster backends
-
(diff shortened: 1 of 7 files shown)


jakubhava:
Otherwise looks good to me, I'll try it locally in a bit

jakubhava:
@mmalohlava tested, can confirm that run-example, run-sparkling and sparkling-shell all works on windows, good work!

mmalohlava:
@jakubhava regarding `run-sparkling.cmd` - i keep parity with *UNIX `run-sparkling.sh`

PR #210, opened 2017-03-16
[SW-321] Ensure h2oframe without reference is not deleted too early. …

jakubhava:
…Solves the cbind problem the other problems of similar type


mmalohlava:
Let's wait for tests and merge on green.

jakubhava:
@mmalohlava please wait before merging - I will attach test case for this behaviour as well

PR #212, opened 2017-03-17
[SW-321][HOTFIX] Attach the frame to dataframe to avoid early garbage…

jakubhava:
… collection of the frame

### DIFF
--- a/py/pysparkling/context.py
+++ b/py/pysparkling/context.py
@@ -143,7 +143,6 @@ class H2OContext(object):
         return self._conf
 
     def as_spark_frame(self, h2o_frame, copy_metadata=True):
-        fr = h2o_frame # create artificial reference so python doesn't delete the frame on the input. Solves SW-321
         """
         Transforms given H2OFrame to Spark DataFrame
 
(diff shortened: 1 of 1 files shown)


py/pysparkling/context.py
```diff
@@ -156,10 +155,17 @@ def as_spark_frame(self, h2o_frame, copy_metadata=True):
...
+            j_h2o_frame = h2o_frame.get_java_h2o_frame()
             jdf = self._jhc.asDataFrame(j_h2o_frame, copy_metadata, self._jsql_context)
-            return DataFrame(jdf, self._sql_context)
+            df = DataFrame(jdf, self._sql_context)
+            # Attach h2o_frame to dataframe which forces python not to delete the frame when we leave the scope of this
+            # method.
+            # Without this, after leaving this method python would garbage collect the frame since it's not used
+            # anywhere and spark. when executing any action on this dataframe, will fail since the frame
+            # would be missing.
+            df._h2o_frame = h2o_frame
```
mmalohlava: One note: in theory we should redefine `_del_` method on `DataFrame` to kill also `_h2o_frame` reference. But for now, it is good enough to unblock cust.

PR #213, opened 2017-03-17
[SW-357] Find the right ClassLoader to register SW jar

mmalohlava:
Zeppelin and other platforms bring additional ClassLoader on top of
Spark MutableURLClassLoader. That causes PySparkling to fail on missing `addURL` method.

The patch introduces code which finds the right Spark classloader in active thread classloader
chain. It does not do any error recovery if the classloader is not found.

### DIFF
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -32,7 +32,8 @@ class Initializer(object):
         if cl.getClass().getName()=='com.databricks.backend.daemon.driver.DriverLocal$DriverLocalClassLoader':
             cl.getParent().getParent().addURL(url)
         else:
-            cl.addURL(url)
+            spark_cl = Initializer.__find_spark_cl(cl, 'org.apache.spark.util.MutableURLClassLoader')
+            spark_cl.addURL(url)
 
         # Add Sparkling Water Assembly JAR to Spark's file server so executors can fetch it when they need to use the dependency.
         jsc.addJar(sw_jar_file)
@@ -53,3 +54,12 @@ class Initializer(object):
     def __get_sw_jar():
         return os.path.abspath(resource_filename("sparkling_water", 'sparkling_water_assembly.jar'))
 
+    @staticmethod
+    def __find_spark_cl(start_cl, cl_name):
+        cl = start_cl
+        while cl:
+            name = cl.getClass().getName()
+            if (name == cl_name):
+                return cl
+            cl = cl.getParent()
+        return None


ramkrishnan8994:
@mmalohlava @jakubhava 
We ran the following command on Zeppelin - 
`pip install h2o_pysparkling_2.3`

After which we are running - 
```
from pysparkling import *

hc = H2OContext.getOrCreate(sc)
```

We are still facing this issue - 
```
Py4JError: An error occurred while calling o77.addURL. Trace:
py4j.Py4JException: Method addURL([class java.net.URL]) does not exist
```

The path that has been merged is also available in the codes. But we are installing 2.3 version of Pysparkling.

jakubhava:
@ramkrishnan8994 what is the exact version of `h2o_pysparkling_2.3` you are using? Can you verify it is the latest one available on Pipy?

ramkrishnan8994:
> @ramkrishnan8994 what is the exact version of `h2o_pysparkling_2.3` you are using? Can you verify it is the latest one available on Pipy?

This is the version -  **h2o-pysparkling-2.3==2.3.31**

I solved the issue by commenting out the `cl.addUrl(url)` snippet from the code. 
I looked at versions of the code and understood that the change came in because of something related to DataBricks.

I ran the snippet of code that is actually running on 'initializer.py' and found the type of methods inside the classLoader. I do not find **'addUrl'** anywhere. Somehow it goes into the true case of the if block and executes `cl.addUrl(url)`. I commented that line of code and it worked perfectly

image


Also, we ran a small pysparkling code in a Spark Standalone cluster and it had no issues. Didn't have to touch the code or anything

jakubhava:
In Which environment are you running Sparkling Water? Are you building Sparkling Water yourself or using official distribution?

ramkrishnan8994:
> In Which environment are you running Sparkling Water? Are you building Sparkling Water yourself or using official distribution?

We are not building anything by ourself.
We run pip install `h2o_pysparkling_2.3` on Zeppelin notebooks, thats it.

jakubhava:
We still need information about the environment. Databrics/Azure/AWS ? Or YARN directly (HDP, CDH)? We need more details to be able to reason about the possible issue.

Thank you!

ramkrishnan8994:
We run it in AWS on top of a Kubernetized Environment. We use Spark Standalone cluster. We dont use HDP or CDH. Its our own cluster

ramkrishnan8994:
Also @jakubhava ,
We run Zeppelin on 'Per User' and 'isolated' mode. In this the SparkContext is created for each user and JVM is also separate for each user.

Now, when a user creates a H2OContext, it gets created. 
When a new user creates a new H2OContext, one of the following errors:
`1. IllegalArgumentException: u'\nH2O context hasn\'t been started successfully in the previous attempt and H2O client with previous configuration is already running.\nBecause of the current H2O limitation that it can\'t be restarted within a running JVM,\nplease restart your job or spark session and create new H2O context with new configuration.")\n  `

```
2. Py4JJavaError: An error occurred while calling z:org.apache.spark.h2o.JavaH2OContext.getOrCreate.
: java.lang.RuntimeException: Cloud size under 10
```

We assumed that the error is because the H2OContext JVM is already created on the Spark Workers and when a new user creates a H2OContext, resources are not available or there is a conflict. 

Please let me know what the issue might be. This is URGENT

PR #215, opened 2017-03-20
[SW-361] Setup H2O dependencies explicitly

mmalohlava:


mmalohlava:
@jakubhava @mdymczyk may you please verify this patch also manually by building SW and running it (JVM/Python)

PR #216, opened 2017-03-20
[SW-362] Fix Pysparkling by fixing initialization of spark session in…

jakubhava:
…side repl

See brief explanation here https://0xdata.atlassian.net/browse/SW-362

This commit also brings  tiny stylistic changes and removes `sqlc` from `H2OContext` since it's not used there anywhere


repl/src/main/scala/org/apache/spark/repl/h2o/BaseH2OInterpreter.scala
```diff
@@ -101,7 +100,7 @@ private[repl] abstract class BaseH2OInterpreter(val sparkContext: SparkContext,
   private def initializeInterpreter(): Unit = {
     settings = createSettings()
     intp = createInterpreter()
-    val spark = SparkSession.builder().config(sparkConf).getOrCreate()
```
michalkurka: Nice detective work! I didn't notice that myself.

PR #221, opened 2017-03-22
[SW-353] Fix loading of JAAS module

jakubhava:
Since the web was disabled, the jetty wasn't even started and thus JAAS module wasn't loaded


jakubhava:
@mmalohlava good point, reverting this as it needs change it h2o

mmalohlava:
Just idea since you are already touching configuration options: spark is using this kind of stuff: src/main/scala/org/apache/spark/internal/config/package.scala

(we can use object style as well - each parameter is defined by name, default value, type and help)

jakubhava:
Cool! Good tip @mmalohlava . I had something like that in my mind and already created some similar draft for this in separated local branch but then I didn't want to pollute this change with other refactoring as well so I just kept it there.

I created this helper method just not to copy bunch of code around at least for this fix. 

I can either do the proposed change in here, but better would be to do in in another PR and have separated JIRA for it.

mmalohlava:
Why this change contains `kerb.conf` file?

jakubhava:
@mmalohlava probably accident, I will remove it in the another commit

PR #223, opened 2017-03-23
[SW-265] Use Spark Session in public API instead of Spark Context

jakubhava:
<s>TODO: Documentation changes<s>

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The Sparkling Shell supports creation of an H<sub>2</sub>O cloud and execution o
 4. Initialize H2OContext 
   ```scala
   import org.apache.spark.h2o._
-  val hc = H2OContext.getOrCreate(sc)
+  val hc = H2OContext.getOrCreate(sparkSession)
   ```
 
   > H2OContext start H2O services on top of Spark cluster and provides primitives for transformations between H2O and Spark datastructures.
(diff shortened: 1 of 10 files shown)


mmalohlava:
Also the tests are failing now:

```
org.apache.spark.h2o.H2OConfTestSuite > test H2OConf parameters FAILED
    java.lang.NoSuchMethodException at H2OConfTestSuite.scala:29```

jakubhava:
@mmalohlava Yup, I saw the failure but had to leave for train. Working on it now 👍

jakubhava:
Should be passing now, waiting for tests to finish

mmalohlava:
Thinking about it little bit more:
  - where is boundary between using `SparkContext` and `SparkSession`?

Would it make sense to have part of SW which depends on SparkContext only?

jakubhava:
Since SparkSession is wrapping also `sqlContext` I was thinking of removing implicit `SQLContext` argument to `AsDataFrame` methods wince spark session contains it. 

Regarding the boundary - I think that on `H2OContext` and REPL level we should use SparkSession since it has the sqlContext and sparkContext and we could apply the simplification above. However the rest of the code can use SparkContext as we do now.

Since this would be API change it's good to discuss this properly, thanks for new ideas @mmalohlava !

mmalohlava:
One more comment - this is changing API, so all people using SparkContext will need to rewrite code. Is it possible to preserve the old API as well, and we will remove it in 3.x?

jakubhava:
@mmalohlava good point. I think I already implemented that - both `sparkContext` and `sparkSession` are accepted but deprecation warning is printed in case of `sparkContext`

PR #231, opened 2017-03-28
[SW-377] Implement stopping of external H2O cluster in external backend

jakubhava:
Plus some stylistic changes to pysparkling conf


jakubhava:
I already tested this out manually on our hadoop cluster together with the https://github.com/h2oai/h2o-3/pull/901 and works as expected

jakubhava:
I think there is no need - I tested it in both scala and pysparkling and it works. Currently the Python stop method will stop the external h2o cluster when running in Auto cluster mode, but does not shutdown the current h2o client session in the python since the general problem with stopping h2o in one JVM

mmalohlava:
oki, going to merge! thx kuba!

PR #233, opened 2017-03-29
[SW-380] Initial jenkinsfile for sparkling water

nikhilshekhar:


mmalohlava:
@jakubhava @mdymczyk this is initial commit mainly for testing.

jakubhava:
Awesome, thanks @nikhilshekhar ! . Will go through it as the first thing during my morning

jakubhava:
Just quick note. I think we should test the external cluster by default as well - so each test should automatically run tests for internal and external backend. More and more people are starting to use the external backend and since tests on the external backend are disabled by default, they are literally never run.

What do you think ? @mmalohlava @mdymczyk

mmalohlava:
@jakubhava yes, agree, we should test all possible configuration. This is initial attempt to setup proper Jenkins pipeline for testing PRs.

PR #235, opened 2017-03-30
[SW-376] Fix DataType conversion. The DateType is internally handled …

jakubhava:
…as Long and the corresponsing Java class has to be marked as Long as well


mmalohlava:
Ups, lot of test errors.

jakubhava:
Yup, I haven't pushed a fixed version yet, I have it in my local repo and will push it later tonight

jakubhava:
This is the working fix. We need to explicitly overrides the internal java class which is used during the conversion for these 2 types in external cluster mode.

jakubhava:
still need one small fix on the external backend

jakubhava:
The DeepLearning failing on the external cluster is intermittent issue which happens pretty often, however not related to this change. We can merge and proceed with the release ( at least of 1.6 ) 👍

PR #237, opened 2017-03-31
[SW-385] Small fix to residualPlotRCode

meganjkurka:
* Updated R code in `residualPlotRCode` function.
* Added h2oContext to the `residualPlotRCode` arguments in the readme example.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -177,7 +177,7 @@ You can configure Sparkling Water using the following variables:
 13. Generate an R-code producing residual plot:
   ```scala
   import org.apache.spark.examples.h2o.AirlinesWithWeatherDemo2.residualPlotRCode
-  residualPlotRCode(predictionH2OFrame, 'predict, bigTable, 'ArrDelay)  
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -177,7 +177,7 @@ You can configure Sparkling Water using the following variables:
 13. Generate an R-code producing residual plot:
   ```scala
   import org.apache.spark.examples.h2o.AirlinesWithWeatherDemo2.residualPlotRCode
-  residualPlotRCode(predictionH2OFrame, 'predict, bigTable, 'ArrDelay)  
+  residualPlotRCode(predictionH2OFrame, 'predict, bigTable, 'ArrDelay, h2oContext)
```
mmalohlava: @meganjkurka why do you need h2oContext here?

meganjkurka: Because if someone is running regular h2o at the same time they are trying this sparkling water example, then the residual plot r code won't execute.

mmalohlava: 👍

PR #240, opened 2017-04-05
[SW-389] Print extended JAR  task every time when extendJar is called

jakubhava:
This change also removes the part where we parse version in format x.y from hadoop distro as the parsing is no longer needed. The distributions have always name like that

### DIFF
--- a/assembly-h2o/build.gradle
+++ b/assembly-h2o/build.gradle
@@ -97,18 +97,18 @@ dependencies {
 task printExtendedJarLocation(){
     doLast {
         def fileName = h2oJarFileName(originalJar)
-        println(new File(projectDir, "private${File.separator}extended${File.separator}${fileName}.jar"))
+        println(file("private/extended/${fileName}.jar"))
     }
 }
 
 clean.doFirst{
-   delete "${projectDir}${File.separator}private${File.separator}extended"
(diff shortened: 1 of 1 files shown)


assembly-h2o/build.gradle
```diff
@@ -94,6 +94,13 @@ dependencies {
...
 }
 
+task printExtendedJarLocation(){
+    doLast {
+        def fileName = h2oJarFileName(originalJar)
+        println(new File(projectDir, "private${File.separator}${fileName}.jar"))
+    }
+}
+
 task extendJar(type: ShadowJar, dependsOn:[":sparkling-water-ml:jar", ":sparkling-water-core:jar"]) {
```
mmalohlava: This one should depend also on extendJar cleanup.

mmalohlava:
Minor comment: the files are downloaded outside build folder, so they are not deleted by regular gradle cleanup (`./gradlew clean`). What is motivation behind that? Caching?

jakubhava:
@mmalohlava yup, the downloaded files are not cleaned due to caching

jakubhava:
Regarding the location, good catch, fixed!

mmalohlava:
regarding caching - then we should have a task to clean it.

PR #241, opened 2017-04-05
[SW-388] Stop h2o when used in standalone pysparkling application

jakubhava:


py/pysparkling/context.py
```diff
@@ -119,12 +119,20 @@ def getOrCreate(spark_context, conf = None):
         # Create H2O REST API client
         h2o.connect(ip=h2o_context._client_ip, port=h2o_context._client_port)
         h2o_context.is_initialized = True
+        # Stop h2o when running standalone pysparkling scripts and the user does not explicitly close h2o
+        atexit.register(lambda: h2o_context.stop())
         return h2o_context
 
     def stop(self):
-        warnings.warn("H2OContext stopping is not yet fully supported...")
+        h2o.cluster().shutdown()
```
mmalohlava: This will shutdown all JVMs. why you cannot use `_jhc.stop(False)` directly?

py/pysparkling/context.py
```diff
@@ -119,12 +119,20 @@ def getOrCreate(spark_context, conf = None):
...
         self._jhc.stop(False)
 
+    def __del__(self):
+        self.stop()
+
+    def __exit__(self):
+        self.stop()
```
mmalohlava: Maybe, you should separate `atexit` functionality from regular `H2OContext.stop`. 

Something like:

```python
import atexit

@atexit.register
def goodbye():
    h2o.cluster().shutdown()
    self._jhc.stop(False)
```

py/pysparkling/context.py
```diff
@@ -119,12 +119,20 @@ def getOrCreate(spark_context, conf = None):
...
 
     def stop(self):
-        warnings.warn("H2OContext stopping is not yet fully supported...")
+        h2o.cluster().shutdown()
         self._jhc.stop(False)
 
+    def __del__(self):
+        self.stop()
+
+    def __exit__(self):
```
mmalohlava: `__exit__` does not make sense without overriding `__enter__`

py/pysparkling/context.py
jakubhava: I can, but  wanted to ensure proper stopping of python client as well which seems like we do not handle from pysparkling

py/pysparkling/context.py
jakubhava: I probably don't see the difference - we run the same code so it seemed like a good code reause

py/pysparkling/context.py
jakubhava: thanks, good to know 👍

jakubhava: I just copied this code from h2o code, but now when I read more about it, `__exit__` and `__enter__` doesn't make much sense in this context as h2o can't be reused. So I'm removing this

py/pysparkling/context.py
jakubhava: What do you please mean by closing all JVMs - closing all h2o nodes ? Shouldn't that be correct stopping behaviour ?

mmalohlava: It will shutdown JVM directly.

My point is that client (=PySparkling) should be to some extend stupid. Backend should be clever one to know how to properly shutdown cluster.

py/pysparkling/context.py
```diff
@@ -119,12 +119,17 @@ def getOrCreate(spark_context, conf = None):
...
+        # Stop h2o when running standalone pysparkling scripts and the user does not explicitly close h2o
+        atexit.register(lambda: h2o_context.stop())
         return h2o_context
 
     def stop(self):
         warnings.warn("H2OContext stopping is not yet fully supported...")
         self._jhc.stop(False)
 
+    def __del__(self):
+        self.stop()
```
mmalohlava: One more comment: if somebody will do:
```
def fu():
   h2oContext = ...
   ...
   h2oContext.stop()

func()
```

Then `stop` will be called twice - one explicitly by user, the second time by `__del__` - is it a problem?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -218,9 +218,14 @@ class H2OContext private (val sparkContext: SparkContext, conf: H2OConf) extends
     * @param stopSparkContext  stop also spark context
     */
   def stop(stopSparkContext: Boolean = false): Unit = {
-    announcementService.shutdown
-    backend.stop(stopSparkContext)
-    H2OContext.stop(this)
+    if(!stopped) {
```
mmalohlava: This can be in theory invoked by multiple threads -  it should be synchronized on h2oContext.

jakubhava: Good catch, will work on that

PR #244, opened 2017-04-06
[SW-393] Allow to configure timeout for clouding up in external mode …

jakubhava:
…from sparkling water


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -129,6 +135,9 @@ trait ExternalBackendConf extends SharedBackendConf {
 
 object ExternalBackendConf {
 
+  /** Timeout for starting h2o external cluster */
+  val PROP_EXTERNAL_CLUSTER_START_TIMEOUT = ("spark.ext.h2o.cluster.start.timeout", 120)
```
mmalohlava: What is unit ?

jakubhava: Seconds, good catch! I will comment it properly

mmalohlava:
@jakubhava Ups, there is a conflict. Pls could you pls update this PR.

jakubhava:
@mmalohlava should be good now

PR #248, opened 2017-04-07
[SW-397] Better error message when nofify file is not created

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -91,6 +91,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
     logError(processErr.toString)
 
 
+    if(!new File(hc.getConf.clusterInfoFile.get).exists()){
```
mmalohlava: Btw: you can use Gradle `file` call.

vpatryshev: One small nitpick.
It makes sense to stick to the coding style that IntelliJ supports as default. Meaning, 
if (!new File(hc.getConf.clusterInfoFile.get).exists()) {
Of course, since we are here, could also check that the file is readable, and that hc.getConf.clusterInfoFile.get does not throw an NPE and does not return a null. :) (Just commenting)

jakubhava: @vpatryshev regarding style - good catch, forgot about that
regarding checking whether `hc.getConf.clusterInfoFile.get` returns `null` -> not necessary, it's set with default when initiating context if not set

PR #252, opened 2017-04-09
[SW-398] Specify timeout for reading/writing confirmation

jakubhava:
When timeout is reached, the exception is thrown. It is not catched on purpose since the spark will stop the job on the current executor and restart it. That should fix several intermittent issues together with  the problem of a few jobs never finishing ( the problem appeared and the confirmation never appeared)

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
@@ -19,23 +19,26 @@ package org.apache.spark.h2o.backends.external
 
 import org.apache.spark.h2o.converters.WriteConverterCtx
 import org.apache.spark.h2o.converters.WriteConverterCtxUtils.UploadPlan
-import org.apache.spark.h2o.utils.{NodeDesc, ReflectionUtils}
 import org.apache.spark.h2o.utils.SupportedTypes._
+import org.apache.spark.h2o.utils.{NodeDesc, ReflectionUtils}
 import org.apache.spark.sql.types._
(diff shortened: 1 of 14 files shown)


mmalohlava:
Test error:
```
/home/travis/build/h2oai/sparkling-water/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala:40: too many arguments for method waitUntilAllWritten: ()Unit
      externalFrameWriter.waitUntilAllWritten(10)```

File: core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala [hunk 1]
```diff
@@ -31,11 +31,14 @@ class ExternalWriteConverterCtx(nodeDesc: NodeDesc, totalNumOfRows: Int) extends
...
     * This method closes the communication after the chunks have been closed
+    * @throws ExternalFrameConfirmationException in case of confirmation failure.
+    *         This has also effect of Spark stopping the current job and
+    *         rescheduling it
     */
   override def closeChunks(): Unit = {
     try{
-      externalFrameWriter.waitUntilAllWritten()
-    }finally {
+      externalFrameWriter.waitUntilAllWritten(10)
```
mmalohlava: this is dangerous? is it 10 sec, 10min?

jakubhava: Secods, but it is a still a draft. I need to make it configurable, but didn't have enough time at PyData

jakubhava:
Regarding the test error - it's expected since it depends on a change in h2o branch

PR #264, opened 2017-04-19
[SW-411] Migrate PySparkling examples, notebooks and tests to SparkSe…

jakubhava:
…ssion ( from SparkContext and SQLContext )


jakubhava:
Jenkins test passing

PR #267, opened 2017-04-25
[SW-421] Skip yarn files delete

mmalohlava:
The delete was marking Yarn job as failed, since we deleted yarn temporary files.


jakubhava:
We don't need to delete the notify file, when the user specifies the same file, it gets overridden by H2O

PR #271, opened 2017-05-09
[SW-427] Upgrade H2O version to 3.10.4.7

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "Upgrade H2O version to 3.10.4.7" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #273, opened 2017-05-11
[SW-430] Improve internal cache for Java references

mmalohlava:


py/pysparkling/context.py
```diff
@@ -28,22 +28,29 @@ def determine_java_vec_type(vec):
...
-        if hasattr(self, '_java_frame'):
+        # Can we use cached H2O frame?
+        # Only if we cached it before and cache was not invalidated by rapids expression
+        if hasattr(self, '_java_frame') and self._java_frame is not None \
+           and self._ex._cache._id is not None and not self._ex._cache.is_empty() \
+           and self._ex._cache._id == self._java_frame_sid:
             return self._java_frame
         else:
+            # Note: self.frame_id will trigger frame evaluation
             return hc._jhc.asH2OFrame(self.frame_id)
```
jakubhava: should we cache it here it here as well?

mmalohlava: ohh, that's interesting idea - i think it sounds good

PR #275, opened 2017-05-12
[SW-431] Allow to pass additional arguments to run-python-script.sh

jakubhava:
This seems like the easiest and the cleanest option. Spark will fail if the file doesn't exist, so we don't need to check for that.

### DIFF
--- a/bin/run-python-script.sh
+++ b/bin/run-python-script.sh
@@ -8,14 +8,6 @@ checkSparkHome
 # Verify if correct Spark version is used
 checkSparkVersion
 
-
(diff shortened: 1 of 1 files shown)


jakubhava:
@mmalohlava good catch, nope

jakubhava:
@mmalohlava had a look, there is no corresponding python script for this functionality. So it's fine

jakubhava:
Created Jira for creating corresponding Windows script  https://0xdata.atlassian.net/browse/SW-432

PR #280, opened 2017-05-18
[SW-436] Remove cache for extracted sparkling water jar.

jakubhava:
Also remove the if condition handling the old egg files.

### DIFF
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -61,16 +61,19 @@ class Initializer(object):
         import sparkling_water
         sw_pkg_file = sparkling_water.__file__
         # Extract jar file from zip
-        from pkg_resources import get_cache_path
-        zip_file = sw_pkg_file[:-len('/sparkling_water/__init__.py')]
-        zip_filename = os.path.basename(zip_file)
-        jar_path = get_cache_path(zip_filename)
-        sw_jar = os.path.abspath("{}/sparkling_water/sparkling_water_assembly.jar".format(jar_path))
-        import zipfile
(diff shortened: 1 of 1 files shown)


py/pysparkling/initializer.py
```diff
@@ -61,26 +61,15 @@ def __get_sw_jar():
...
-            return os.path.abspath(resource_filename("sparkling_water", 'sparkling_water_assembly.jar'))
-
-
+        from pkg_resources import get_cache_path
+        zip_file = sw_pkg_file[:-len('/sparkling_water/__init__.py')]
+        zip_filename = os.path.basename(zip_file)
+        jar_path = get_cache_path(zip_filename)
+        sw_jar = os.path.abspath("{}/sparkling_water/sparkling_water_assembly.jar".format(jar_path))
+        import zipfile
+        with zipfile.ZipFile(zip_file) as fzip:
```
mmalohlava: How do you know that it will be zip file? 

The path above was handling situations when `pip` unzipped installed package (however, i have to say the semantics of pip is unclear for me in this context).

jakubhava: that's good comment, I will revert this part, thanks! However I wouldn't cache the files. Caching old artefacts caused some weird issues in my python client because of class inconsistencies

jakubhava: oh, so to understand it correctly - the non-zip part was there for `pip` purposes ?

jakubhava:
@mmalohlava what do you think about this change, can I go ahead and merge ? I just hit the problem with cache again. I restored the zip handling so we should handle pip properly now

PR #281, opened 2017-05-22
[SW-426] Provide driver logs and link to download the logs in the spa…

jakubhava:
…rk UI and history server


jakubhava:
This is a final version. The driver logs can be seen at runtime and can be downloaded from the history server and the spark UI as well.

PR #282, opened 2017-05-23
[SW-437] Don't call atexit in case of pysparkling in cluster deploy mode

jakubhava:


jakubhava:
Oki! It's related to stopping H2O. Once we can do that properlythis can all be cleaned up a bit

PR #283, opened 2017-05-23
[SW-442] Get array size from metadata

petro-rudenko:
Running sparkling water on large datasets (100+gb) noticed it spends a lot of time on figuring out feature vector size: [](https://postimg.org/image/5dggcpw75/). If feature vector was created by Spark's VectorAssembler it would have array size [inside metadata](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/VectorAssembler.scala#L88). So what it does - it iterates for all vector/array indicies in metadata and if for all of them there's size in metadata - it can return result in a constant time.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
@@ -18,9 +18,11 @@
 package org.apache.spark.h2o.utils
 
 import org.apache.spark.h2o._
+import org.apache.spark.ml.attribute.AttributeGroup
 import org.apache.spark.sql._
 import org.apache.spark.sql.types._
 import org.apache.spark.{SparkContext, ml, mllib}
+import scala.annotation.tailrec
 
 /**
  * Utilities for working with Spark SQL component.
@@ -102,7 +104,7 @@ object H2OSchemaUtils {
     val vecColIdxs    = collectVectorLikeTypes(schema.fields)
     val numOfArrayCols = arrayColIdxs.length
     // Collect max arrays for this RDD, it is distributed operation
-    val fmaxLens = collectMaxArrays(sc, srdd.rdd, arrayColIdxs, vecColIdxs)
+    val fmaxLens = collectMaxArrays(sc, srdd, arrayColIdxs, vecColIdxs)
     // Flattens RDD's schema
     val flatRddSchema = flatSchema(schema)
     val typeIndx = collectTypeIndx(schema.fields)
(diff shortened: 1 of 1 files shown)


h2o-ops:
Hi @petro-rudenko, your PR title "Get array size from metadata" is missing JIRA issue number! Please specify it in the following form [SW-]

jakubhava:
Hi @petro-rudenko, thanks for the PR! Let me have a look on it as soon as possible

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -176,12 +177,27 @@ object H2OSchemaUtils {
     * @return list of max sizes for array types, followed by max sizes for vector types. */
   private[h2o]
   def collectMaxArrays(sc: SparkContext,
-                       rdd: RDD[Row],
+                       dataframe: DataFrame,
                        arrayTypesIndx: Seq[Seq[Int]],
                        vectorTypesIndx: Seq[Seq[Int]]): Array[Int] = {
     val allTypesIndx = arrayTypesIndx ++ vectorTypesIndx
+    val metadataInfo = (field: StructField) => AttributeGroup.fromStructField(field).numAttributes
```
jakubhava: I would rename  `metadataInfo` to something like `attributesNum`

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -176,12 +177,27 @@ object H2OSchemaUtils {
...
+
+    def getBasicType (columns: StructType, indexes: Seq[Int]): StructField = {
+      val structField = columns.fields(indexes.head)
+      if (indexes.tail.isEmpty) structField
```
jakubhava: I would use `if(..){
..
}` also here to be consistent, but just my opinion :)

petro-rudenko:
Hey guys, thanks for review. This trick works only with `ml.lilang.VectorUDT` type. But another option is to have some magic key in metadata so user would tell SW the size of arrays type:
`structField.metadata.contains("h2o.array.size")`. 
Recursion is needed to handle nested schema, since `collectVectorLikeTypes` returns `Seq[Seq[Int]]`.

jakubhava:
@petro-rudenko looks good! I will run our internal tests to check if they pass as well. I created JIRA for this issue https://0xdata.atlassian.net/browse/SW-442

In the meanwhile, could you squash to commits and prefix the single commit with the change with `[SW-442]` prefix ? Thanks!!

jakubhava:
All test passed, thanks @petro-rudenko. This change will be in the next release.

PR #284, opened 2017-05-23
[SW-439] handle interrupted exception in H2ORuntimeInfoUIThread

jakubhava:


core/src/main/scala/org/apache/spark/h2o/ui/H2ORuntimeInfoUIThread.scala
```diff
@@ -26,9 +26,13 @@ import water.H2O
   */
 class H2ORuntimeInfoUIThread(sc: SparkContext, conf : H2OConf) extends Thread{
   override def run(): Unit = {
-    while(!Thread.interrupted()){
+    while(!Thread.currentThread().isInterrupted){
```
mmalohlava: I do not fully understand the motivation here. Why you need to re-interrupt current thread and reread status again? A simple flag should be enough, no?

jakubhava: good point. I didn't know that `interrupted()` clears the flag but `isInterrupted` keeps the flag.
In that case no change is required here. 👍

PR #293, opened 2017-06-05
[SW-448] Add missing deepwater jar

jakubhava:


assembly/build.gradle
```diff
@@ -61,6 +61,9 @@ shadowJar {
     include(dependency("com.amazonaws:aws-java-sdk-kms"))
     include(dependency("com.amazonaws:aws-java-sdk-core"))
 
+    //deepwater jar
+    include(dependency("ai.h2o:deepwater-backend-api"))
```
mmalohlava: Should be in "ai.h2o" block

jakubhava: Fixed

PR #295, opened 2017-06-08
[SW-24] Add tab to download page explaining using sparkling water as …

jakubhava:
…spark package


jakubhava:
manually tested after running make-dist.sh locally

PR #297, opened 2017-06-12
[SW-369] Use only spark nodes with H2O nodes in internal backend

jakubhava:
This change ensures that only executors discovered during the spread rdd phase will be used during the computations. Therefore, we are no longer affected by a new spark executor joining the cluster

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContextImplicits.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContextImplicits.scala
@@ -21,6 +21,7 @@ import org.apache.spark.mllib.regression.LabeledPoint
 import org.apache.spark.sql.DataFrame
 import water.Key
 import org.apache.spark._
+import scala.reflect.ClassTag
 import scala.reflect.runtime.universe._
 
 /**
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContextImplicits.scala
```diff
@@ -29,7 +31,7 @@ abstract class H2OContextImplicits {
 
   protected def _h2oContext: H2OContext
   /** Implicit conversion from RDD[Supported type] to H2OFrame */
-  implicit def asH2OFrameFromRDDProduct[A <: Product : TypeTag](rdd : RDD[A]): H2OFrame = _h2oContext.asH2OFrame(rdd, None)
+  implicit def asH2OFrameFromRDDProduct[A <: Product : ClassTag : TypeTag](rdd : RDD[A]): H2OFrame = _h2oContext.asH2OFrame(rdd, None)
```
mmalohlava: ?

jakubhava: We need `ClassTags` and `TypeTags` in order to create `H2OAwareRDD`. It failed on compile error without this change since the class tag wasn't available for `asH2OFrame` method

jakubhava:
Jenkins tests passing

jakubhava:
Do you plan to put this change it to the next release as well @mmalohlava ? Just asking :)

jakubhava:
One more thought - Maybe we could wait and try to create test to it - enforce joining new executor. Not sure how to it, but right now the test for it is missing.

mmalohlava:
Nop, i need to test it extensively.

jakubhava:
Yup, I realized that right after the first message that I haven't included any tests. I will write tests simulating joining new spark executors to cluster when I return. This is not urgent issue anyway.

petro-rudenko:
Having another issue on largish datasets: spark/h2o executor goes OOM and got killed by Yarn. H2O then hang forever w/o any progress. Was thinking to add new listener `onExecutorRemoved` to stop h2o if one of it's executor died.

jakubhava:
Tests need to be written for this change

jakubhava:
Closing this issue, project Hydrogen should help us with solving this elegantly

PR #300, opened 2017-06-13
[SW-453] [Minor] Use size method to get attr num

petro-rudenko:
Minor fix to my previous PR. If fields had some attribute VectorAssembler would squash them to Array[Attribute] rather than numAttribute. Need to use size method: https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/attribute/AttributeGroup.scala#L80 - that would handle all cases to get num of attributes.


jakubhava:
Thanks @petro-rudenko ! The change looks good. Just minor thing - can you please use this jira number -  https://0xdata.atlassian.net/browse/SW-453 - in the commit number instead SW-442 ? I created new jira for it even though it's bug fix of a previous change. This allows us to monitor the changes better. Thanks!

jakubhava:
Looks good, I'm just waiting for the internal tests to finish

jakubhava:
@petro-rudenko thanks for contribution!

PR #298, opened 2017-06-12
[SW-450] Edits to the Installation instructions

angela0xdata:
Edited the installation instructions on the DOWNLOAD, RUN ON HADOOP, and RUN ON STANDALONE CLUSTER tabs. Changed
`val h2oContext = H2OContext.getOrCreate(sparkSession)`
to
`val h2oContext = H2OContext.getOrCreate(spark)`

PR #301, opened 2017-06-13
[SW-454] Replace sparkSession with spark in backends documentation

jakubhava:


jakubhava:
This is done in order to be consistent with https://github.com/h2oai/sparkling-water/pull/298

PR #303, opened 2017-06-14
[SW-456] Make shell scripts safe

jakubhava:
Wanted to do this for a long time already

There is no semantic change except in the `run-example` where I removed the unnecessary condition. The deploy mode should be specified in all cases. It has effect only when running on yarn and spark properly handles the MASTER and DEPLOY_MODE configurations since some are not valid.

### DIFF
--- a/bin/get-extended-h2o.sh
+++ b/bin/get-extended-h2o.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Current dir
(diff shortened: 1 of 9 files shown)


h2o-ops:
Hi @jakubhava, your PR title "Make shell scripts safe" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #304, opened 2017-06-14
[SW-457] Clean up windows scripts

jakubhava:
Clean up some parts of windows scripts. Removing the condition inside run-example as it's not necessary.
Also, no need to put jar files on class-path when running pysparkling since it's already handled by --py-files ( the same is done in sparkling-shell )

### DIFF
--- a/bin/pysparkling2.cmd
+++ b/bin/pysparkling2.cmd
@@ -11,12 +11,9 @@ rem end of checking Sparkling environment
 call %LIBSW% :banner
 
 rem setup pysparkling command line
-rem Because of SPARK-18648 we need to put assembly also on driver/executor class paths
-SET SPARK_OPT_JARS=--jars !FAT_JAR_FILE! --conf spark.driver.extraClassPath=!FAT_JAR_FILE! --conf spark.executor.extraClassPath=!FAT_JAR_FILE!
 SET SPARK_OPT_PYFILES=--py-files !PY_ZIP_FILE!
-cd %TOPDIR%
 SET PYTHONPATH=%PY_ZIP_FILE%:%PYTHONPATH%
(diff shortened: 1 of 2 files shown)


bin/pysparkling2.cmd
```diff
@@ -11,12 +11,9 @@ rem end of checking Sparkling environment
 call %LIBSW% :banner
 
 rem setup pysparkling command line
-rem Because of SPARK-18648 we need to put assembly also on driver/executor class paths
-SET SPARK_OPT_JARS=--jars !FAT_JAR_FILE! --conf spark.driver.extraClassPath=!FAT_JAR_FILE! --conf spark.executor.extraClassPath=!FAT_JAR_FILE!
 SET SPARK_OPT_PYFILES=--py-files !PY_ZIP_FILE!
-cd %TOPDIR%
 SET PYTHONPATH=%PY_ZIP_FILE%:%PYTHONPATH%
-call !SPARK_HOME!/bin/pyspark2.cmd !SPARK_OPT_PYFILES! %SPARK_OPT_JARS% %*
```
jakubhava: @mmalohlava Do we need need to put jars explicitly like this ? I read https://issues.apache.org/jira/browse/SPARK-18648 and it says that `--jars` on windows doesn't work correctly.

However we add the jars on the driver path and executors path in the code explicitly in the `Initializer.py`

```        # Add Sparkling Water Assembly JAR to Spark's file server so executors can fetch it
        # when they need to use the dependency.
        jsc.addJar(sw_jar_file)
```

It's not a big deal, it would just be clear code and less options used.

jakubhava: I will remove this part of the change if you think it's dangerous.

I will keep the rest because then it will be consistent with shell scripts where we don't need to handle deploy mode explicitly as well

jakubhava:
So small update after testing - without configuring extraClassPath it doesn't work. I'm therefore reverting this part of this change and merging the reset

PR #306, opened 2017-06-14
[SW-459] Update run-time dependencies for PySparkling

jakubhava:
The list of run-time dependencies was deprecated. Verified by running tests.

### DIFF
--- a/py/setup.py
+++ b/py/setup.py
@@ -47,7 +47,7 @@ setup(
     packages = find_packages(),
 
     # run-time dependencies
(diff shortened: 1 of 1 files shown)


py/setup.py
```diff
@@ -47,7 +47,7 @@
     packages = find_packages(),
 
     # run-time dependencies
-    install_requires=['six', 'future', 'requests', 'tabulate'],
+    install_requires=['requests', 'tabulate', 'future', 'colorama'],
```
jakubhava: yup, it's on purpose as the current h2o does not require six as well

PR #316, opened 2017-06-19
[SW-466] Separate DEVEL.md into multiple doc files

jakubhava:
- convert to rst
- make up-to-date with current spark
- also make up-to-date with 2 backends types


jakubhava:
Already done. All docs made up-to-date, converted to rst and separated. Should be a good step for the next documentation steps.

mmalohlava:
Great! One minor thing - can we put into `doc` folder one file as entry point (probably it will look like `DEVEL.md`, but we can design a new one.)

jakubhava:
I think it's already there. There is DEVEL.rst inside the doc folder which acts as the table of content.

mmalohlava:
Great! SO i think we can merge! 🎉

PR #318, opened 2017-06-21
[SW-XXX] Mm 4release

mmalohlava:


h2o-ops:
Hi @mmalohlava, your PR title "Mm 4release" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #319, opened 2017-06-22
[SW-471] Use mojo in pipelines if possible, remove H2OPipeline and On…

jakubhava:
…eTimeTransformers

This PR does 3 things:
 - removes H2OPipeline
 - removes DataSet splitter as it was one time transformer ( ugly ). It's better to do the splitting as part of the algo call. This way it's also easier for the user since he/se does not need to care about configuring keys from previous steps
- introduces H2OMOJOModel. This is very similar to H2OModel except that it should work without h2o framework running using MOJO and easy predict wrapper. 
- predictions always return with the original data as it's done in Spark


jakubhava:
This is a final version, ready to be merged

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -0,0 +1,222 @@
...
+        case BinaryType =>
+          row.getAs[Array[Byte]](idxRow).zipWithIndex.foreach { case (v, idx) =>
+            dt.put(entry.name + idx, v.toString)
+          }
+        case ByteType => dt.put(entry.name, row.getByte(idxRow).toString)
+        case ShortType => dt.put(entry.name, row.getShort(idxRow).toString)
+        case IntegerType => dt.put(entry.name, row.getInt(idxRow).toString)
+        case LongType => dt.put(entry.name, row.getLong(idxRow).toString)
+        case FloatType => dt.put(entry.name, row.getFloat(idxRow).toString)
+        case _: DecimalType => dt.put(entry.name, row.getDecimal(idxRow).doubleValue().toString)
```
mmalohlava: Why this fallback?

jakubhava: Not sure at the moment, maybe it's not necessary. However we need this if the user is allowed to create objects of type DecimalType.

 I would just firstly check if all the types above fill the domain of decimal type ( i.e. That we cover all decimal) and that the DecimalType can't be instantiated. Then we can remove it, otherwise we need to leave it there.

PR #327, opened 2017-06-29
[SW-479] Fix CHANGELOG location in make-dist.sh

jakubhava:
It was moved to doc/CHANGELOG.rst which is already collected by: `$(find doc -type f)`


h2o-ops:
Hi @jakubhava, your PR title "Fix CHANGELOG location in make-dist.sh" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #328, opened 2017-06-29
[SW-480] Bring back publishToMavenLocal task

jakubhava:


mmalohlava:
Ups?

jakubhava:
Sorry, I thought it's small change to merge right away.

I gradle 4.0 the `publishToMavenLocal` task is not available by default. The `maven-publish` plugin needs to be imported and the code in this change has to be used explicitly

mmalohlava:
Ahh oki! Good to know!

PR #334, opened 2017-07-11
[SW-449] Support for Sparse Vectors and Upgrade to H2O 3.14.0.2

jakubhava:
SparkDataFrameConverter refactored. Now we first create flat dataframe in spark to simplify handling on sw side. 

The refactoring was needed in order to separate handling of vectors.

An improvement of this solution would be to drop the flag whether the vector is sparse or dense and make it part of the expected types

This change requires https://github.com/h2oai/h2o-3/pull/1348

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
@@ -20,12 +20,12 @@ package org.apache.spark.h2o.backends.external
 import org.apache.spark.h2o.H2OConf
 import org.apache.spark.h2o.backends.SharedBackendUtils
 import org.apache.spark.h2o.utils.NodeDesc
-import water.H2O
+import water.{ExternalFrameUtils, H2O}
 
 /**
   * Various helper methods used in the external backend
   */
-private[external] trait ExternalBackendUtils extends SharedBackendUtils{
+private[external] trait ExternalBackendUtils extends SharedBackendUtils {
 
   /**
     * Get arguments for H2O client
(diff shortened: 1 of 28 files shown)


jakubhava:
Still some tests are failing on external cluster, but getting to the final state. Benchmarks shows this is a good step

jakubhava:
Ready for review, this is final state

jakubhava:
@mmalohlava `continuous-integration/jenkins/pr-head` shouldn't succeed as this change depends on different H2O version, even during compilation

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -42,4 +42,37 @@ private[external] trait ExternalBackendUtils extends SharedBackendUtils{
...
+        ExternalFrameUtils.EXPECTED_LONG
+      } else if (clazz == classOf[java.lang.Float]) {
+        ExternalFrameUtils.EXPECTED_FLOAT
+      } else if (clazz == classOf[java.lang.Double]) {
+        ExternalFrameUtils.EXPECTED_DOUBLE
+      } else if (clazz == classOf[java.lang.String]) {
+        ExternalFrameUtils.EXPECTED_STRING
+      } else if (clazz == classOf[java.sql.Timestamp]) {
+        ExternalFrameUtils.EXPECTED_TIMESTAMP
+      } else if (clazz == classOf[org.apache.spark.mllib.linalg.Vector]) {
```
mmalohlava: For future: what about about `org.apache.spark.ml.linalg.Vector` ?

mmalohlava: Oki, found it, you are doing Vectors.toML(..) coversion before.

mmalohlava:
Right now blocked by:
SW-516, SW-515

PR #337, opened 2017-07-12
[SW-355] Put latest rsparkling into the distribution package

jakubhava:
Also remove dead code in `./make-dist` script

### DIFF
--- a/assembly/build.gradle
+++ b/assembly/build.gradle
@@ -73,6 +73,18 @@ shadowJar {
   }
 }
 
+task printRSparklingCranLink(){
+  doLast {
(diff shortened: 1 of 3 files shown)


mmalohlava:
This needs update of download page to promote a new package.

jakubhava:
Do you mean instructions on how to install package from the distribution zip ? We already have RSparkling tab so we can put the instructions there.

I would put there 2 kind of information - install from cran ( the simplest one) and install from distribution jar locally.

What do you think ?

mmalohlava:
Yes pls

jakubhava:
@mmalohlava the download page is updated. Looks as:

screen shot 2017-07-14 at 11 06 24

PR #339, opened 2017-07-13
[SW-481] SVM MOJO

mdymczyk:


jakubhava:
Once the corresponding H2O change ( https://github.com/h2oai/h2o-3/pull/1361 ) is in the release and SW is upgraded to that version, this can be merged

jakubhava:
Please rebase on top of current master and fix possible scala style checks.

We are starting to enforcing some style checks as part of the build process.

jakubhava:
Rebased on current master, just need to wait for https://github.com/h2oai/h2o-3/pull/1361 to be in

PR #346, opened 2017-07-17
[SW-497] Spark 2.2 integration

jakubhava:
Merge to master and then create rel-2.2 branch
- Fix Java Documentation
- Enable testing in trevis on OpenjDK8 and OracleJDK8
- Fix failing test. This is caused by a bug fix in Spark - https://issues.apache.org/jira/browse/SPARK-14584
- Remove usage of PermGem since on Java8 it's not used anymore
- Upgrade mininal python version to 2.7


toddpi314:
(only 8 days behind the release is pretty good)
bitmoji

toddpi314:
Did you forget to update this? https://github.com/h2oai/sparkling-water/blob/jh/feature/rel-2.2/gradle.properties#L3

toddpi314:
👍 Confirmed this build is looking good on Dataproc with the Spark 2.2 preview image

jakubhava:
@toddpi314 thanks for looking in to this! Good catch with the `gradle.properties`. Fixing that now 👍

jakubhava:
Some tests are failing due to collisions between different guava versions. This will be fixed by https://github.com/h2oai/deepwater/pull/48 and https://github.com/h2oai/h2o-3/pull/1411 in the next h2o release on which this branch will be based.

PR #350, opened 2017-07-20
[SW-500] Warehouse dir does not have to be set in tests on Spark from…

jakubhava:
… 2.1+

This is because https://issues.apache.org/jira/browse/SPARK-17810 was fixed in 2.1


jakubhava:
Yup, thanks for merging!

PR #351, opened 2017-07-23
[SW-501] Propagate all security related files.

mdymczyk:
Propagate ssl.properties and ssl key files through Spark mechanisms instead of YARN.

To try:

1) run master branch to make sure it fails, `sparkling-shell --master yarn`

```
import org.apache.spark.network.Security
import org.apache.spark.h2o._
Security.enableSSL(spark)
val hc = H2OContext.getOrCreate(spark)
```

Should fail complaining about no `ssl.properties` file.

2) Run the above with this branch. H2O cluster should start, there should be mentions of `ssl.properties` in H2O logs and `ssl.properties` and `h2o-internal.jks` should be in temp files of each container (location can be found at the beginning of H2O logs).

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
@@ -46,8 +46,8 @@ trait InternalBackendConf extends SharedBackendConf {
 
   /** Setters */
 
-  def setFlatFileEnabled() = set(PROP_USE_FLATFILE._1, true.toString)
-  def setFlatFileDisabled() = set(PROP_USE_FLATFILE._1, false.toString)
+  def setFlatFileEnabled() = set(PROP_USE_FLATFILE._1, true)
+  def setFlatFileDisabled() = set(PROP_USE_FLATFILE._1, false)
 
   def setNumH2OWorkers(numWorkers: Int) = set(PROP_CLUSTER_SIZE._1, numWorkers.toString)
   def setDrddMulFactor(factor: Int) = set(PROP_DUMMY_RDD_MUL_FACTOR._1, factor.toString)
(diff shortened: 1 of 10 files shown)


mmalohlava:
@mdymczyk we need also an option to enabled the automatic mode.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
```diff
@@ -41,6 +41,9 @@ trait InternalBackendConf extends SharedBackendConf {
   def nodeIcedDir = sparkConf.getOption(PROP_NODE_ICED_DIR._1)
   def nodeNetworkMask = sparkConf.getOption(PROP_NODE_NETWORK_MASK._1)
 
+  def isInternalSecureConnectionsEnabled = sparkConf.getBoolean(PROP_INTERNAL_SECURE_CONNECTIONS._1,
```
jakubhava: Corresponding method `isInternalSecureConnectionsDisabled` is missing. All methods working with booleans have these 2 methods in the current api, so it's good for consistency

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendConf.scala
```diff
@@ -55,6 +58,7 @@ trait InternalBackendConf extends SharedBackendConf {
   def setNodeBasePort(port: Int) = set(PROP_NODE_PORT_BASE._1, port.toString)
   def setNodeIcedDir(dir: String) = set(PROP_NODE_ICED_DIR._1, dir)
   def setNodeNetworkMask(mask: String) = set(PROP_NODE_NETWORK_MASK._1, mask)
+  def setInternalSecureConnectionsEnabled(flag: Boolean) = set(PROP_INTERNAL_SECURE_CONNECTIONS._1, flag)
```
jakubhava: to be consistent with all the methods in the current api of conf, this method should be
`setInternalSecureConnectionsEnabled()` ( no arguments) and there should be corresponding method `setInternalSecureConnectionsDisabled()`

mdymczyk:
@mmalohlava ah you mean like a H2O config property for this? @jakubhava ok will add those

mmalohlava:
@mdymczyk i updated this PR, but it is missing methods kuba is referencing to

PR #352, opened 2017-07-25
[SW-504] Provides uber Sparkling Water Spark package

mmalohlava:
The motivation for this is to improve user experience. Right now, SW provides 4 modules with complex dependencies which need to be often hot-fixed (like  SW-475).
The goal of this improvement is to provide a single uber Sparkling Water package which can be used as input for Spark submit option `--packages`.

This package is being published into maven central and can be used with Spark submit:
```
$SPARK_HOME/bin/spark-submit --packages ai.h2o:sparkling-water-package_2.11:2.1.12 --class org.apache.spark.examples.h2o.CraigslistJobTitlesStreamingApp /dev/null
```

Note: right now it also provides a fixed for [SW-475]

### DIFF
--- a/package/build.gradle
+++ b/package/build.gradle
@@ -1,0 +1,27 @@
+description = 'Sparkling Water Spark Package'
+
+apply plugin: 'com.github.johnrengelman.shadow'
+
+dependencies {
+    // hack to go around Spark Ivy resolver
+    compileOnly "no.priv.garshol.duke:duke:1.2"
+
+    compile project(':sparkling-water-core')
+    compile project(":sparkling-water-examples")
+    compile project(":sparkling-water-ml")
+    compile project(":sparkling-water-repl")
+}
+
+jar {
+    enabled = true
+}
+
+shadowJar {
+    archiveName = jar.archiveName
+    //classifier = ''
+    dependencies {
+        configurations = [project.configurations.compileOnly]
+    }
+}
+
+jar.finalizedBy shadowJar
--- a/package/build.gradle
+++ b/package/build.gradle
@@ -4,7 +4,7 @@ apply plugin: 'com.github.johnrengelman.shadow'
 
 dependencies {
     // hack to go around Spark Ivy resolver
-    compileOnly "no.priv.garshol.duke:duke:1.2"
+    compile "no.priv.garshol.duke:duke:1.2"
 
     compile project(':sparkling-water-core')
     compile project(":sparkling-water-examples")
--- a/settings.gradle
+++ b/settings.gradle
@@ -10,6 +10,7 @@ include 'py'
 include 'repl'
 include 'docker'
 include 'app-streaming'
+include 'package'
 
 // Prefix all projects with sparkling-water name
 rootProject.children.each { project ->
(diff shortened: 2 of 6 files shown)


mmalohlava:
? i thought pysparkling already contains uber jar. This is just a package to use via `--packages` option.

jakubhava:
You're right, PySparkling already contains the assembly jar. For some reason I thought  that this jar does not contain the duke library. Sorry for confusion. 

This is a good change, looks good to me! Thanks!

package/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+    compile project(":sparkling-water-ml")
+    compile project(":sparkling-water-repl")
+}
+
+jar {
+    enabled = true
+}
+
+shadowJar {
+    archiveName = jar.archiveName
```
jakubhava: @mmalohlava Should we also shadow the jetty used in H2O as we do in regular assembly ?

```  relocate 'org.eclipse.jetty', 'ai.h2o.org.eclipse.jetty'```. This would resolve the jetty issues on Databribrix [SW-236] and when using sw as spark package

Comment (tomkraljevic) [inline package/build.gradle]: relocating jetty would mean taking a look at the following documentation as well

h2o-3/h2o-docs/src/product/security.rst

here the user is exposed to jetty classes.

Comment (mmalohlava) [inline package/build.gradle]: @jakubhava The problem here is that we cannot shadow jetty without processing H2O dependencies. That would mean to include h2o dependencies and i would like to avoid that. 

But worth experimenting - we can include only shaddowed jetty and processed h2o-core, but we need to make sure, the we publish dependencies in POM correctly.

@tomkraljevic the problem is more complex. In the context of Spark package, there is no actual binary which you can put on `-cp` as it is mentioned in documentation.  There is only a reference to a package. However, users can do that with binary distribution from download page (we need to put it into SW documentation).

Furthermore, so far it looks like people mostly using automatic mode (equivalent of `-internal_security_enabled` in Hadoop).

Comment (jakubhava) [inline package/build.gradle]: @mmalohlava thanks for explanation. I thought that initial ida for doing this was to solve the problems similar to the Jetty one by including problematic shadowed dependencies.

Even though it's not a nice solution it would improve the user experience on Databrix and when using `--packages`

Comment (mmalohlava) [inline package/build.gradle]: agree, my motivation came from Duke library dependency. 

Let's iterate on this one, since i think this would be great deployemnt scenario.

**Comment by mmalohlava** [comment]
This needs to include automl!

**Comment by mmalohlava** [comment]
So far rejected by DBC (CC: @jakubhava )

**Comment by jakubhava** [comment]
Finished testing using `--package`. I published `sparkling-water-package` to local maven repo and tested it as 
`./bin/sparkling-shell --repositories file:///Users/kuba/.m2/repository/ --packages "ai.h2o:sparkling-water-package:2.2.99999-SNAPSHOT"`

Works fine. We can't test it on DBC until this is published into maven central but I suppose there is high change that it will work as well.

**Comment by jakubhava** [comment]
I would mark this PR as finished and merge it. Relocation of jetty can go to another PR just to move forward with this PR. What do you think @mmalohlava ?

File: settings.gradle [hunk 2]
```diff
@@ -10,6 +10,7 @@ include 'py'
 include 'repl'
 include 'docker'
 include 'app-streaming'
+include 'package'
```
mmalohlava: Just make sure, it is published.

jakubhava: yup, I checked ant the artefact was fine. Also was testing a lot using maven central and also using Jitpack ( It can be used as maven repository as well ) using `--packages`, all was good.

Will verify once again after release.

PR #320, opened 2017-06-22
[SW-380] Minor test changes

mmalohlava:

PR #360, opened 2017-08-04
[SW-512] Make JenkinsFile up-to-date with sparkling_yarn_branch

jakubhava:


jakubhava:
As of this point, this change is more up-to-date version of https://github.com/h2oai/sparkling-water/pull/320 ( Sorry for JIRAs duplication, the initial goal was a bit different )

mmalohlava:
Does it include: https://github.com/h2oai/sparkling-water/pull/320/files ?

jakubhava:
Yup -> Except the `linux` label as the environment is not consistent across nodes at the moment

mmalohlava:
yup, that's fine - i am using it as a proxy to show that we have messy environment. 

Nevertheless, i would prefer to dockerize all steps. We have also internal docker registry - so we can even use it.

mmalohlava:
But good change, if you think that my PR is not needed anymore, pls, close it

jakubhava:
Cool, didn't know about the internal docker hub! If you like, we can merge this whenever you want. Or feel free to update it :)

jakubhava:
All Jenkins job can be removed at this point in favor of this pipeline

PR #362, opened 2017-08-07
[SW-513] Upgrade to Gradle 4.1

jakubhava:
Here's the Gradle Change log https://docs.gradle.org/current/release-notes.html

It contains speed improvements for Scala builds


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Fri Jul 28 17:46:00 CEST 2017
+#Tue Aug 08 00:31:49 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.0.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip
```
mmalohlava: Should be `-all`

jakubhava: hm, this is what you get by `./gradlew wrapper --gradle-version=4.1` so should be correct. 

Here's the Gradle Change log, feel free to merge https://docs.gradle.org/current/release-notes.html

mmalohlava: `-all` contains source code - it is handy if you are debugging something. Try to import the project in idea, it will recommend it.

jakubhava:
Cool, didn't know that, will fix it. Thanks for the tip!

gradle/wrapper/gradle-wrapper.properties
jakubhava: Cool, didn't know that, will fix it. Thanks for the tip!

PR #364, opened 2017-08-11
[SW-514] Upgrade H2O to 3.14.0.1

jakubhava:


jakubhava:
@mmalohlava closing this PR and making H2O upgrade as part of https://github.com/h2oai/sparkling-water/pull/334 since the new H2O api requires this PR

mmalohlava:
It is better to separate functionality - i would say single responsibility per PR is often better.

jakubhava:
Totally agree. I wanted to split the change into 2 however making sparse vectors and h2o upgrade is tight together and I didn't have enough time before leaving for it.

PR #367, opened 2017-08-17
[SW-518] Update to Pysparkling Chicago Crime demo

meganjkurka:
Changes made to notebook: 
- don't specify Date as string (H2O automatically detects it as Date column)
- remove hack: Force evaluation before ifelse and cut. See PUBDEV-1425.
- fixes to crime function
- formatting changes


h2o-ops:
Hi @meganjkurka, your PR title "Update to Pysparkling Chicago Crime demo" is missing JIRA issue number! Please specify it in the following form [SW-]

jakubhava:
This PR and commit need JIRA. Also the corresponding non-notebook demo needs to be updated with the same changes

meganjkurka:
@jakubhava Thanks for the feedback.  I've added a new commit with the same changes done to the non-notebook demo.

py/examples/scripts/ChicagoCrimeDemo.py
```diff
@@ -222,12 +215,12 @@ def score_event(crime, model, censusTable):
     probOfArrest = predictTable["true"][0,0]
     return probOfArrest
 
-for crime in crime_examples:
-    arrestProbGBM = 100*score_event(crime, gbm_model, df_census)
-    arrestProbDLM = 100*score_event(crime, dl_model, df_census)
+for i in crime_examples:
```
jakubhava: Not so important, but 'crime' seems explaining what is the object we are iterating over better than 'i'. Or we can use anything else for the name

meganjkurka: I changed that because `crime` is defined above this as a function.  This is not an issue if you run the notebook as is. If you want to, for example, create more new `crime_examples` later on, then you will get an error because now `crime` is no longer a function.

PR #370, opened 2017-08-24
[SW-521] spelling check

satya23k:


jakubhava:
Thanks @satya23k for the fix!

PR #381, opened 2017-09-12
[SW-423] Tests of External Cluster mode fails when starting H2O locally

jakubhava:
This change fixes the environment for external tests.


mmalohlava:
One observation: it is good to setup timeouts in more fine-grained way - let's say we know that script tests should finish under 20min, so we put the tests into 
```
timeout(20, minutes) {
}```
block

jakubhava:
Good catch! Will apply it as well in the next step

PR #393, opened 2017-09-21
[SW-539] Fix bug when pysparkling is executed in parallel on the same node

jakubhava:
This bug fix introduce back the cache which @mmalohlava wrote a while ago. However it also changes the python egg cache path for tests so the tests always uses the correct latest artefact ( this was source of issues and also reason why the cache was removed ).

Also when testing SNAPSHOT versions locally on different sparkling-water we make sure to use temporary cache for python eggs, again to be sure we run on the latest code.

The cache is fine if it's used by the users on the released versions.

### DIFF
--- a/py/pysparkling/context.py
+++ b/py/pysparkling/context.py
@@ -139,12 +139,14 @@ class H2OContext(object):
 
 
     def stop_with_jvm(self):
+        Initializer.clean_temp_dir()
         h2o.cluster().shutdown()
         self.stop()
 
 
     def stop(self):
         warnings.warn("Stopping H2OContext. (Restarting H2O is not yet fully supported...) ")
+        Initializer.clean_temp_dir()
         self._jhc.stop(False)
 
     def __del__(self):
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -11,6 +11,7 @@ class Initializer(object):
     # Flag to inform us whether sparkling jar has been already loaded or not.
     # We don't want to load it more than once.
     __sparkling_jar_loaded = False
+    __extracted_jar_dir = None
 
     @staticmethod
     def load_sparkling_jar(spark_context):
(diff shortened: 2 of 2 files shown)


py/pysparkling/initializer.py
```diff
@@ -65,15 +65,21 @@ def __get_sw_jar():
...
             zip_filename = os.path.basename(zip_file)
-            jar_path = get_cache_path(zip_filename)
-            sw_jar = os.path.abspath("{}/sparkling_water/sparkling_water_assembly.jar".format(jar_path))
-            import zipfile
-            with zipfile.ZipFile(zip_file) as fzip:
-                fzip.extract('sparkling_water/sparkling_water_assembly.jar', path = jar_path)
-            return sw_jar
+            cache_path = get_cache_path(zip_filename)
+            cached_jar = os.path.abspath("{}/sparkling_water/sparkling_water_assembly.jar".format(cache_path))
+            if os.path.exists(cached_jar) and os.path.isfile(cached_jar):
```
mmalohlava: Thinking about it more, can we disable caching totally? It could be dangerous, if we have cached old version of jar and upgraded sw.

jakubhava:
I'm all for not using cache as the cache was also cause of problems before. We can disable it in this case since we need to extract the JAR any way, but we can extract it to temporary directory which will be cleaned at the end of H2OContext.

py/pysparkling/context.py
```diff
@@ -139,12 +139,14 @@ def getOrCreate(spark, conf=None, **kwargs):
 
 
     def stop_with_jvm(self):
+        Initializer.clean_temp_dir()
```
mmalohlava: Update: This seems causing print of stack-traces during spark shutdown. It cannot expect that it will be fully executed. The better solution is simply skip cleanup.

jakubhava: good point! Will work on this tomorrow

PR #401, opened 2017-10-04
[SW-534] Fix data location so the data are available for external tests as well

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -143,6 +143,7 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
     }
     // Register web API for client
     RestAPIManager(hc).registerAll()
+    Thread.sleep(2000)
```
michalkurka: I suggest to make it a configurable parameter (perhaps with a default value 2000)
This way you will be able to disable it if needed.

Jenkinsfile
```diff
@@ -100,19 +100,7 @@ pipeline{
 
                     echo "-Dhdp.version="${params.hdpVersion}"" >> ${env.SPARK_HOME}/conf/java-opts
                    """
-
```
jakubhava: No longer needed, we get the data from the shared directory

PR #408, opened 2017-10-13
[SW-557] Create 2 jenkins files ( for internal and external backend ) backed by configurable pipeline

jakubhava:


jenkins/Jenkinsfile-external
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/groovy
+@Library('h2o3-shared-lib') _
+// Job parameters
+properties(
+        [
+                parameters(
+                        [
```
mmalohlava: The jobs properties are duplicated - that makes maintenance harder.

jenkins/Jenkinsfile-external
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/groovy
+@Library('h2o3-shared-lib') _
```
mmalohlava: I though @michal-raska  removed that library, it should not be alive.

jenkins/Jenkinsfile-external
```diff
@@ -0,0 +1,67 @@
...
+
+                                booleanParam(name: 'buildAgainstH2OBranch', defaultValue: false, description: 'By default, Sparkling Water is built with the included H2O. This option may be used to force build to use specific H2O branch against which Sparkling Water is built.'),
+                                string(name: 'h2oBranch', defaultValue: 'master', description: 'H2O branch to build against if buildAgainstH2OBranch is set to true')
+                        ]
+                ),
+                buildDiscarder(logRotator(numToKeepStr: '10'))
+        ]
+)
+
+if (env.CHANGE_BRANCH != null && env.CHANGE_BRANCH != '') {
```
mmalohlava: You should use shared jenkins library to test this stuff

mmalohlava: @Library('test-shared-library') _

jenkins/Jenkinsfile-external
```diff
@@ -0,0 +1,67 @@
...
+  cancelPreviousBuilds()
```
mmalohlava: This should be in shared jenkins library (=@Library('test-shared-library'))...CC: @michal-raska

jenkins/Jenkinsfile-external
jakubhava: I know.., they are however a bit different - we don't need to set properties related to external cluster in internal jenkinsfile and vice versa.

We could extract the shared configuration to some shared repo with Jenkins functions, however that seemed  too much for this ( at least to me )

jenkins/Jenkinsfile-external
jakubhava: still works fine, if +@Library('test-shared-lib') _ is on, I will reuse it.

jenkins/Jenkinsfile-external
jakubhava: currently this functionality is available in @Library('h2o3-shared-lib') _ . If `@library('test-shared-lib')` is on, I will change it

jenkins/Jenkinsfile-external
jakubhava: same as above

michal-raska: @mmalohlava I'll create the PR with this in `test-shared-library`

jenkins/Jenkinsfile-external
michal-raska: The h2o3-shared lib will be removed in a few days, since it looks like the H2O-Pipeline is OK without it. I'll check with @jakubhava before removal

jakubhava: `h2o3-shared-lib` is used just for the `cancelPreviousBuilds` method.  I could duplicate it in SW repo, however then I would need to first check-out the SW in order to get this method, but that means that this build would already need to wait for the previous builds. Hence I think that it is a good idea to have this method in the `test-shared-lib`

jakubhava: The other solution is to have this `cancelPreviousBuilds` method in each jenkins file, however I would like to avoid that as this method is universal and can be shared amongst different repos as well.

jakubhava:
@michal-raska @mmalohlava Oh, I didn't know that the `test-shared-library` already exists. I talked with @michal-raska and will use that library once some minor changes are merged there.

Thanks!

PR #410, opened 2017-10-16
[SW-553] Benchmark suite to measure performance of sparse transfers.

mmalohlava:
The benchmark measures time of transfering ultra wide Spark sparse
vectors to H2O Frame representation.

Results for 3k x 50k (rows x cols), sparse vector, 0.2 sparsity (20% of
non-zeros), single Spark partition transfered to a single H2O chunk:

```
Mean           Stdev        Min           Max            Unit
39452.800781 ± 1380.934204 (37205.000000, 40707.000000) (msec)
```

Results for 10 x 11 matrics, sparse vector with a single non-zero
element (matrix has a single zero column):

```
Mean        Stdev     Min        Max         Unit
64.500000 ± 8.669871 (56.000000, 80.000000)  (msec)
```

### DIFF
--- a/core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterBenchTest.scala
+++ b/core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterBenchTest.scala
@@ -1,0 +1,77 @@
+/*
+* Licensed to the Apache Software Foundation (ASF) under one or more
+* contributor license agreements.  See the NOTICE file distributed with
+* this work for additional information regarding copyright ownership.
+* The ASF licenses this file to You under the Apache License, Version 2.0
+* (the "License"); you may not use this file except in compliance with
+* the License.  You may obtain a copy of the License at
+*
+*    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterBenchTest.scala
```diff
@@ -0,0 +1,77 @@
...
+    import org.apache.spark.h2o.utils.BenchUtils.bench
+    import TestUtils.sparseVector
+    import sqlContext.implicits._
+    val NCOL = 50*1000
+    val SPARSITY = 0.2
+    val NROW = 3*1000
+    val PARTITIONS = 1
+
+    val elementsPerRow = (SPARSITY * NCOL).toInt
+    val rowGenerator = (row: Int) => {
```
jakubhava: `row` argument doesn't seem to be used

mmalohlava:
Can i merge?

jakubhava:
@mmalohlava yup, thanks for the benchmarks!

PR #414, opened 2017-10-20
[SW-559] Test for missing 0-len chunks

mmalohlava:
Note: this will fail till the patched H2O version is used (see https://github.com/h2oai/h2o-3/pull/1661)


michalkurka:
Good idea to have it tested also on SW's side. Thanks!

PR #416, opened 2017-10-20
[SW-562] Disable web on external H2O nodes in external cluster mode

jakubhava:
The web remains opened on the client node, therefore this change should not be breaking change. We expose the ip:port of the client node


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -66,6 +66,7 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
       "-mapperXmx", conf.mapperXmx,
       "-output", conf.HDFSOutputDir.get,
       "-J", "-log_level", "-J", conf.h2oNodeLogLevel,
+      "-J", "-disable_web",
```
mmalohlava: could we have an option to handle that? Or in fact a way how to pass options to external cluster launcher. We can always stick with `-Dai.h2o.disable_web` syntax...

jakubhava: We can have option for sure 👍 I wanted to mimic the internal cluster behaviour, where we simply disable it without no options for security reasons

PR #418, opened 2017-10-20
[SW-563] Print out also yarn app id of external cluster

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/SparklingBackend.scala
```diff
@@ -34,4 +34,6 @@ trait SparklingBackend {
   def stop(stopSparkContext: Boolean)
 
   def backendUIInfo: Seq[(String, String)]
+
+  def backendFinalisedString: String
```
mmalohlava: Minor: weird name for variable - `closingString`? Or preamble antonym - `epilog`?

jakubhava: Thanks!  I really like `epilog`

PR #424, opened 2017-10-23
[SW-499] Implement handling for BinaryType

jakubhava:


jakubhava:
No particular motivation, it was just missing piece in our implementation ( Hosever it took  short time to implement this )

PR #427, opened 2017-10-24
[SW-569] Fix stopping of application from PySparkling  ( fix & less n…

jakubhava:
…oise )

The change allowing running multiple pysparkling sessions in parallel is not the only cause of some noise when stopping pysparkling app.

This change is also motivated by another additional point:
 - When ensuring all tests are passing on external backend, I made stopping of backends consistent on both clusters. However this also caused that we can't stop H2OContext on both backends from PySparkling ( we stop java app and then we get a lots of exceptions because of py4j)

So this change just informs the user that H2OContext can't be stoped when called `hc.stop()` and PySpark sessions needs to be restarted. 

When the user calls `exit()` the cluster is shutdown as before using `h2o.cluster.stutdown()` which correctly stops the cluster. This was tested on internal and external cluster. In case of external cluster, it was tested also on yarn and I verified that the exit codes are correct.

From my point and point of the user this is the cleanest solution. We can stop cluster correctly during `exit` and inform the cluster that h2o just can't be restarted when calling `stop`. No other magic.

### DIFF
--- a/py/pysparkling/context.py
+++ b/py/pysparkling/context.py
@@ -136,8 +136,14 @@ class H2OContext(object):
         # In driver mode the application would call exit which is handled by Spark AM as failure
         deploy_mode = spark_session.sparkContext._conf.get("spark.submit.deployMode")
         if deploy_mode != "cluster":
-            atexit.register(lambda: h2o.cluster().shutdown())
+            atexit.register(lambda: h2o_context.__stop())
         return h2o_context
+
+    def __stop(self):
+        try:
+            h2o.cluster().shutdown()
+        except:
+            pass
 
     def stop(self):
         warnings.warn("Stopping H2OContext from PySparkling is not fully supported. Please restart your PySpark session and create a new H2OContext.")
--- a/py/pysparkling/context.py
+++ b/py/pysparkling/context.py
@@ -132,24 +132,15 @@ class H2OContext(object):
         print(h2o_context)
 
         # Stop h2o when running standalone pysparkling scripts, only in client deploy mode
-        #, so the user does not explicitly close h2o.
+        #, so the user does not need explicitly close h2o.
         # In driver mode the application would call exit which is handled by Spark AM as failure
         deploy_mode = spark_session.sparkContext._conf.get("spark.submit.deployMode")
         if deploy_mode != "cluster":
-            atexit.register(lambda: h2o_context.stop_with_jvm())
+            atexit.register(lambda: h2o.cluster().shutdown())
         return h2o_context
 
-
-    def stop_with_jvm(self):
-        Initializer.clean_temp_dir()
-        h2o.cluster().shutdown()
-        self.stop()
-
-
     def stop(self):
-        warnings.warn("Stopping H2OContext. (Restarting H2O is not yet fully supported...) ")
-        Initializer.clean_temp_dir()
-        self._jhc.stop(False)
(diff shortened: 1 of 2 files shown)


py/pysparkling/context.py
```diff
@@ -132,24 +132,15 @@ def getOrCreate(spark, conf=None, **kwargs):
         print(h2o_context)
 
         # Stop h2o when running standalone pysparkling scripts, only in client deploy mode
-        #, so the user does not explicitly close h2o.
+        #, so the user does not need explicitly close h2o.
         # In driver mode the application would call exit which is handled by Spark AM as failure
         deploy_mode = spark_session.sparkContext._conf.get("spark.submit.deployMode")
         if deploy_mode != "cluster":
-            atexit.register(lambda: h2o_context.stop_with_jvm())
+            atexit.register(lambda: h2o.cluster().shutdown())
```
mmalohlava: Let's make it more robust - i would not expect that the method will always have time to finish and as a result it will show bunch of random exceptions. 

My proposal is to make the best effort to shutdown in quite mode.

jakubhava: I think that this is solution is already quite in all the cases. `h2o.cluster.shutdown()` tries to close the cluster using the Rest API. I tried it in several environments and tests and it closed nicely with no exceptions.

This is the code handling the closing the cluster
```if self._session_id:
            try:
                # If the server gone bad, we don't want to wait forever...
                if self._timeout is None: self._timeout = 1
                self.request("DELETE /4/sessions/%s" % self._session_id)
                self._print("H2O session %s closed." % self._session_id)
            except Exception:
                pass
            self._session_id = None
        self._stage = -1
```

So I think we're fine, what do you think ?

mmalohlava: btw code ^^ will not shutdown cluster, only delete active objects.

jakubhava: ups, sorry. I didn't share the full code. We use the `shutdown` method
```  def shutdown(self, prompt=False):
        """
        Shut down the server.
        This method checks if the H2O cluster is still running, and if it does shuts it down (via a REST API call).
        :param prompt: A logical value indicating whether to prompt the user before shutting down the H2O server.
        """
        if not self.is_running(): return
        assert_is_type(prompt, bool)
        if prompt:
            question = "Are you sure you want to shutdown the H2O instance running at %s (Y/N)? " \
                       % h2o.connection().base_url
            response = input(question)  # works in Py2 & Py3 because redefined in h2o.utils.compatibility module
        else:
            response = "Y"
        if response.lower() in {"y", "yes"}:
            h2o.api("POST /3/Shutdown")
            h2o.connection().close()
```

where the code above is the content of `h2o.connection.close()`

mmalohlava: but this can still throw an exception, no?

jakubhava: hm, only exception could be printed because of `h2o.api("POST /3/Shutdown)`, `h2o.connection().close()` is wrapped in try catch.

So what I can do is to wrap the lambda into `atexit.register(lambda: stop())`
```def stop():
  try: 
      h2o.cluster().shutdown()
  except:
     #ignore everything
```

jakubhava: The possible exception is ignored in the latest commit. What do you think about it know pls?

jakubhava:
Thanks @mmalohlava !!

PR #429, opened 2017-10-24
[SW-535] asH2OFrame gives error if column names have DOT in it

jakubhava:


core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -52,9 +52,11 @@ private[h2o] object SparkDataFrameConverter extends Logging {
   /** Transform Spark's DataFrame into H2O Frame */
   def toH2OFrame(hc: H2OContext, dataFrame: DataFrame, frameKeyName: Option[String]): H2OFrame = {
     import H2OSchemaUtils._
-
+    val substPattern = "_____SUBST_DOT____"
```
mmalohlava: Why not `_` ?

core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterTest.scala
```diff
@@ -729,6 +729,13 @@ class DataFrameConverterTest extends FunSuite with SharedH2OTestContext {
 
   }
 
+  test("Convert DataFrame to H2OFrame with dot in column name"){
+    import sqlContext.implicits._
```
mmalohlava: Try to test nested structure:

```
field1
field2.1
   |- field2.1.1
   |- field2.1.2
```

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
jakubhava: I thought that `_` might be used in names quite a lot

core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterTest.scala
jakubhava: good point 👍

jakubhava: Implemented necessary changes together with test for this case

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
mmalohlava: i prepared the following code to unblock a cust:

```scala
import org.apache.spark.sql.Column
import org.apache.spark.sql.types._
import org.apache.spark.sql.functions._

def flattenSchema(schema: StructType, prefix: String = null) : Array[Column] = {
  schema.fields.flatMap(f => {
    val colPath = if (prefix == null) s"`${f.name}`" else s"${prefix}.`${f.name}`"

    f.dataType match {
      case st: StructType => flattenSchema(st, colPath)
      // Skip user defined types like array or vectors
      case x if x.isInstanceOf[ArrayType] => Array.empty[Column]
      case _ => Array(col(colPath).alias(colPath.replaceAll("[.`]", "_")))
    }
  })
}

val df = spark.read.parquet("/tmp/blahala.csv")
df.select(flattenSchema(df.schema):_*).write.csv("/tmp/xxx.csv")```

jakubhava: thanks! It seems like this code is doing very similar thing to what is implemented in this change with just one difference `_` vs ` `_____SUBST_DOT____`

I think that `_` can be used very often, any ideas how to overcome this without using static replace pattern ?

jakubhava:
@mmalohlava do you think this change is fine or do you see any other improvements to be done on it pls ?

PR #431, opened 2017-10-25
[SW-574] Process steam handle and use it for connection to external h…

jakubhava:
…2o cluster

This code still needs to be tested in steam environment.
It allows starting external h2o cluster using steam and then creating H2OContext on it using the following API.
```
import h2osteam
from pysparkling import *

# Get steam reference
l = h2osteam.login('https://steam', 'kuba', 'kuba', verify_ssl=False)
# Create cluster and get the handle
external_cluster_handle = l.start_h2o_cluster('cluster_name', 4)
# Create H2O Context
hc = H2OContext.getOrCreate(spark, external_cluster_handle=external_cluster_handle)
```

In the implementation, we need to get IP address and port of one node in the cluster. The handle contains the ip and port of the proxy so I'm not sure if that will do. We need to test it.

### DIFF
--- a/py/pysparkling/context.py
+++ b/py/pysparkling/context.py
@@ -102,42 +102,59 @@ class H2OContext(object):
 
         self.is_initialized = False
 
+    def __default_h2o_connect(h2o_context, **kwargs):
+        return h2o.connect(ip=h2o_context._client_ip, port=h2o_context._client_port, **kwargs)
+
+
     @staticmethod
-    def getOrCreate(spark, conf=None, verbose=True, **kwargs):
+    def getOrCreate(spark, conf=None, verbose=True, pre_create_hook=None, h2o_connect_hook=__default_h2o_connect, **kwargs):
         """
-         Get existing or create new H2OContext based on provided H2O configuration. If the conf parameter is set then
-         configuration from it is used. Otherwise the configuration properties passed to Sparkling Water are used.
-         If the values are not found the default values are used in most of the cases. The default cluster mode
-         is internal, ie. spark.ext.h2o.external.cluster.mode=false
+        Get existing or create new H2OContext based on provided H2O configuration. If the conf parameter is set then
+        configuration from it is used. Otherwise the configuration properties passed to Sparkling Water are used.
+        If the values are not found the default values are used in most of the cases. The default cluster mode
+        is internal, ie. spark.ext.h2o.external.cluster.mode=false
 
-         param - Spark Context or Spark Session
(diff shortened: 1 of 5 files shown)


py/pysparkling/context.py
```diff
@@ -103,6 +104,30 @@ def getOrCreate(spark, conf=None, **kwargs):
          returns H2O Context
         """
 
+        # When external cluster handle is specified then we need to make sure the configuration
+        # is corresponding to this situation -> We already have H2O cluster running and need to connect to it.
+        # For this we need to use manual cluster start mode, need to know the cloud name and ip address and port of a
+        # single node in this cluster
+        if external_cluster_handle is not None:
```
mmalohlava: Would be great to test this path in cooperation with @CreatureDev . A possible problem i can see, is that we are forwarding comm via steam proxy. Probably not big deal, however, in theory proxy port does not need to be visible on edge node side - we need to just test it.

jakubhava: If I understand correctly, only communication with REST API should be transmitted using steam proxy. The rest of internal communication is done via internal protocol ( h2o client <-> h2o worker nodes) and should not be affected by this.

Just to understand this properly, why the proxy port might not be visible on the edge node ?

py/pysparkling/context.py
```diff
@@ -92,41 +92,58 @@ def __do_init(self, spark_session):
...
             spark_session = SparkSession.builder.getOrCreate()
+        # Get H2OConf
+        if conf is not None:
+            selected_conf = conf
+        else:
+            selected_conf = H2OConf(spark_session)
+
+        # Call pre_create hook
+        if pre_create_hook:
+            pre_create_hook(spark_session, selected_conf)
```
jakubhava: We are passing H2OConf to steam/different service here as an argument - that probably means that steam will need pysparkling dependency

mmalohlava: yes

mmalohlava: https://github.com/h2oai/h2o-secure/commit/fde1bb711328a12d85de1c2e6948bcf406d27f50

jakubhava: cool!

jakubhava:
I brought this on top of current master plus fixed a issue preventing this to build

mmalohlava:
@jakubhava can we merge this one?

jakubhava:
Sure, let me just catch up on this change. Can it wait till the today's steam meeting?

PR #434, opened 2017-10-26
[SW-519] Introduce SW Models into Spark python pipelines

jakubhava:
ColumnPruner, H2OGBM and H2ODeepLearning ( H2OGBM backed by mojo, H2ODeepLearning backed by binary model ) can now be used in PySpark pipelines.

### DIFF
--- a/py/pysparkling/ml/algo.py
+++ b/py/pysparkling/ml/algo.py
@@ -1,0 +1,76 @@
+from pyspark import since, keyword_only
+from pyspark.ml.param.shared import *
+from pyspark.ml.util import JavaMLReadable, JavaMLWritable
+from pyspark.ml.wrapper import JavaEstimator, JavaModel, JavaTransformer, _jvm
+from pyspark.sql import SparkSession
+from pysparkling import *
+from .params import H2OAlgorithmParams
+
+
+class H2OGBM(JavaEstimator, H2OAlgorithmParams, JavaMLReadable, JavaMLWritable):
+
+    @keyword_only
+    def __init__(self, ratio=1.0, featuresCols=[], predictionCol=None):
+        super(H2OGBM, self).__init__()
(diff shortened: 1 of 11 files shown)


jakubhava:
@mmalohlava when you have some free time please have a look.

I still need to make all available arguments to models in scala in python as well but after that it's final.

This makes mojo predictions available from pyspark

py/pysparkling/ml/algo.py
```diff
@@ -0,0 +1,76 @@
...
+    def setParams(self, ratio=1.0, featuresCols=[], predictionCol=None):
+        kwargs = self._input_kwargs
+        return self._set(**kwargs)
+
+    def _create_model(self, java_model):
+        return H2OGBMModel(java_model)
+
+
+class H2OGBMModel(JavaModel, JavaMLWritable, JavaMLReadable):
+    pass
```
mmalohlava: How do i create an instance of `H2OGBMModel` if i have MOJO in hand?

mmalohlava:
+1 for merge

py/pysparkling/ml/algo.py
jakubhava: This I still need to figure out. But I think that we just need to create a helper method on python side similar to ```def createFromMojo(is: InputStream, uid: String = Identifiable.randomUID("mojoModel")): H2OMOJOModel = {```

which would call this method using py4j and then use this as JavaModel

This is jira for it, marked for the release after this one https://0xdata.atlassian.net/browse/SW-611

jakubhava:
https://0xdata.atlassian.net/browse/SW-612 jira for making all args available in python API, also marked for the next release

jakubhava:
thx @mmalohlava for review!

py/pysparkling/ml/algo.py
jakubhava: Already resolved in the latest code

PR #435, opened 2017-10-26
[SW-420] Check java availability first before checking spark version

jakubhava:
Implemented in both python and unix scripts

### DIFF
--- a/bin/libsw.cmd
+++ b/bin/libsw.cmd
@@ -21,7 +21,19 @@ if not exist "%SPARK_HOME%/bin/spark-submit.cmd" (
 )
 exit /b 0
 
+
+:checkJava
+
+set "JV="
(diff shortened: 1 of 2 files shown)


bin/libsw.cmd
```diff
@@ -21,7 +21,19 @@ if not exist "%SPARK_HOME%/bin/spark-submit.cmd" (
 )
 exit /b 0
 
+
+:checkJava
+
+set "JV="
+For /F "Tokens=3" %%A In ('java -version 2^>^&1') Do If Not Defined JV Set "JV=%%~A"
```
mmalohlava: `For` or `for` ?

jakubhava: good catch! Cmd is not case sensitive, but I will turn these all to lower case to be consistent

PR #436, opened 2017-10-26
[SW-320] Unblock H2OConf test

jakubhava:
It's workaround on the Gradle level

### DIFF
--- a/py/build.gradle
+++ b/py/build.gradle
@@ -45,6 +45,21 @@ def preparePythonEnv(environment) {
 }
 
 //
+// Prepare Python environment to run python tests
+//
+def preparePythonTestEnv(environment){
+    // Add PySparkling zip on PYTHONPATH so unittests can access all the PySparkling packages
+    environment['PYTHONPATH'] = "${distPython.archivePath}${FPS}${environment['PYTHONPATH']}"
+    environment['spark.ext.h2o.backend.cluster.mode'] = detectBackendClusterMode()
+    environment["spark.ext.h2o.external.start.mode"] = detectExternalBackendStartMode()
+    // sparkling assembly jar is here because of external h2o tests
+    environment["sparkling.assembly.jar"] = configurations.compile.join(',')
+
+    environment['SPARK_WORKER_DIR'] = file('build/h2ologs-pyunit/nodes')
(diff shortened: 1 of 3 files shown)


py/build.gradle
```diff
@@ -170,27 +185,24 @@ artifacts {
...
-
+    preparePythonTestEnv(environment)
+    commandLine getOsSpecificCommandLine([pythonexec, "tests/tests_unit_conf.py"])
+}
 
+task testPythonConversions(type: Exec, dependsOn: distPython){
+    doFirst {
+        preparePythonEnv(environment)
+    }
+    preparePythonTestEnv(environment)
```
mmalohlava: Why one preparation is at configuration stage and second one at execution?

py/tests/tests_unit_conf.py
```diff
@@ -0,0 +1,52 @@
...
+        cls._hc = H2OContext.getOrCreate(cls._spark, h2o_conf)
+
+    @classmethod
+    def tearDownClass(cls):
+        h2o.cluster().shutdown()
+        test_utils.tear_down_class(cls)
+
+    # test passing h2o_conf to H2OContext
+    def test_h2o_conf(self):
+        self.assertEquals(self._hc.get_conf().cloud_name(),"test-cloud", "Configuration property cloud_name should match")
```
mmalohlava: is it possible to test the rest properties of config? Futhermore, the test should fail if we add a new config property.

py/build.gradle
jakubhava: The `preparePythonTestEnv(environment)` depends on the JAR archive of SW, that's why it is in the execution phase

jakubhava:
The original problem was that when having multiple test classes in a single python file, the second tests were blocking the gradle as the underlying JVM was shared and H2O can't be restarted. By moving the H2OConf test into a new file and executing it separately we create a workaround for this issue as the tests are executed independently

py/tests/tests_unit_conf.py
jakubhava: I created these 2 PRs ( just to keep this one simple ) https://0xdata.atlassian.net/browse/SW-578 - for testing all properties

https://0xdata.atlassian.net/browse/SW-579 - for testing consistency of properties.

We might actually automatically generate the python configuration based on the Scala one.

PR #437, opened 2017-10-26
[SW-576] Major cleanup of pysparkling test infrastructure

jakubhava:


jakubhava:
Before the configuration we passes to different kinds of unit tests might be applied to all the tests. Currently we have only single integ tests so it doesn't mater but I added more integ tests in one of the previous PR where weird behaviour could occur

PR #458, opened 2017-11-08
[SW-593] H2OConf initialization exception on windows

RobinTPotter:
change to file:/// to initialize the H2OConf. /// is understood by all according to https://en.wikipedia.org/wiki/File_URI_scheme (including windows shell)

### DIFF
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -26,7 +26,7 @@ class Initializer(object):
         jvm = sc._jvm
         # Add Sparkling water assembly JAR to driver
         sw_jar_file = Initializer.__get_sw_jar()
-        url = jvm.java.net.URL("file://{}".format(sw_jar_file))
+        url = jvm.java.net.URL("file:///{0}".format(sw_jar_file))
         # Assuming that context class loader is always instance of URLClassLoader
         # ( which should be always true)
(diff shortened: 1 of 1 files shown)


h2o-ops:
Hi @RobinTPotter, your PR title "Update initializer.py" is missing JIRA issue number! Please specify it in the following form [SW-]

RobinTPotter:
hi @h2o-ops - I wouldn't know a JIRA number if it gave me a haircut :)

RobinTPotter:
updated

jakubhava:
Great, thanks a lot! I just did some testing on my Unix machine and it works as well. Let just wait for the tests to finish and then we can merge 👍

jakubhava:
On unix the created path will be actually `file:////xx/yy` because of the prepended `file:///` and the absolute path `/xx/yy`. I tried it and worked fine on Mac and out test machines, however seems like this is not standardised.

What we can do yo avoid this is to check if `sw_jar_file` starts with `/` and if it doesn't, we can add it. What do you think @RobinTPotter ? can you please change it ?

jakubhava:
Perfect, looks good! Thanks for contribution!

RobinTPotter:
thanks, added a check

PR #461, opened 2017-11-09
[SW-594] Gradle ml submodule reports success even though tests fail

jakubhava:


core/src/test/scala/org/apache/spark/h2o/utils/SharedH2OTestContext.scala
```diff
@@ -39,10 +41,38 @@ trait SharedH2OTestContext extends SparkTestContext {
...
-    hc = null
-    resetSparkContext()
-    super.afterAll()
+    // The method H2O.exit calls System.exit which confuses Gradle and marks the build
+    // as successful even though some tests failed.
+    // We can solve this by using security manager which forbids System.exit call.
+    // It is safe to use as all the methods closing H2O cloud and stopping operations have been
+    // already called and we just need to ensure that JVM with the client/driver doesn't call the System.exit method
+    try {
+      val securityManager = new NoExitCheckSecurityManager
```
mmalohlava: Wow, good trick!

jakubhava: This can be scary Gradle behaviour ( false-positive), I couldn't find why it happened only in ml submodule, but this made it deterministic

PR #462, opened 2017-11-09
[SW-597] Daily integration tests with H2O Master in Jenkins

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "Jh/jira/sw 597" is missing JIRA issue number! Please specify it in the following form [SW-]

mmalohlava:
Does it publish results into S3?

jakubhava:
nope, right now it's that just testing using composite build, no artefacts are published. Well, we archive tests results on jenkins level. Should these be reported to s3 as well together with the h2o and sw artefacts ?

mmalohlava:
i was thinking we can have a regular nightly bleeding edge release.

jakubhava:
Good idea - I think that we can do it even when we use the composite build. We just put call at the end `./make-dist.sh` and publish whole result to s3

mmalohlava:
can you use H2O nightly - the reason for that is that we will have exact connection between builds.

jakubhava:
oki, this makes sense. Is H2O nightly in some internal repo already and do we have way how to discover the current latest nightly version ?

In that case we could just change gradle.properties to point to that nigtly and build against it

mmalohlava:
There is latest link: `s3://h2o-release/h2o/master/latest` - it is available via HTTP:

```
curl http://h2o-release.s3.amazonaws.com/h2o/master/latest
```

Then you can get project version

```
curl http://h2o-release.s3.amazonaws.com/h2o/master/$(curl http://h2o-release.s3.amazonaws.com/h2o/master/latest)/project_version
```

mmalohlava:
And maven repo is on the latest:
```
s3cmd ls s3://h2o-release/h2o/master/4080/maven
```

jakubhava:
Thanks for great info! Working on it in different branch, will share soon

PR #465, opened 2017-11-15
[SW-600] Proper Sparkling Water Nightly

jakubhava:
Testing against h2o nightly.

We build and test sw against latest H2O nightly and upload the results into S3.

I think that we don't have to generate also all extended jars for external cluster as if we need to get them, we can just download the distribution and call script `./get-extended-h2o`. What do you think @mmalohlava ?

### DIFF
--- a/jenkins/Jenkinsfile-external
+++ b/jenkins/Jenkinsfile-external
@@ -67,6 +67,9 @@ withCustomCommitStates(scm, 'h2o-ops-personal-auth-token', 'continuous-integrati
             backendMode = "external"
             hdpVersion = "${p.hdpVersion}"
             driverHadoopVersion = "${p.driverHadoopVersion}"
+            buildNightly = "false"
+            uploadNightly = "false"
+
          }
      }
  }
--- a/jenkins/sparklingWaterPipeline.groovy
+++ b/jenkins/sparklingWaterPipeline.groovy
@@ -3,6 +3,7 @@
 def call(params, body) {
     def config = [:]
     body.resolveStrategy = Closure.DELEGATE_FIRST
+
     body.delegate = config
     body(params)
 
@@ -35,12 +36,23 @@ def call(params, body) {
                         scriptsTest()(config)
                         integTest()(config)
                         pysparklingIntegTest()(config)
+                        publishNightly()(config)
                     }
                 }
             }
         }
     }
 }
+
+
+def getGradleCommand(config) {
+    if (config.buildAgainstH2OBranch.toBoolean()) {
+        "H2O_HOME=${env.WORKSPACE}/h2o-3 ${env.WORKSPACE}/gradlew --include-build ${env.WORKSPACE}/h2o-3"
+    } else {
+        "${env.WORKSPACE}/gradlew"
+    }
+}
+
 
 def prepareSparkEnvironment() {
     return { config ->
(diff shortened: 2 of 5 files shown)


jenkins/sparklingWaterPipeline.groovy
```diff
@@ -70,6 +82,33 @@ def prepareSparklingWaterEnvironment() {
...
+                        script: "curl http://h2o-release.s3.amazonaws.com/h2o/master/latest",
+                        returnStdout: true
+                ).trim()
+
+                def h2oNightlyMajorVersion = sh(
+                        script: "curl http://h2o-release.s3.amazonaws.com/h2o/master/${h2oNightlyBuildVersion}/project_version | cut -f1-3 -d\".\"",
+                        returnStdout: true
+                ).trim()
+
+                sh  """
```
mmalohlava: Minor note: you do not need shell to process properties in this sense, maybe it could be easier to generate new one...or read them with help of groovy.

jakubhava:
I can read the url with groovy without shell, good point.

jenkins/sparklingWaterPipeline.groovy
```diff
@@ -220,4 +251,36 @@ def pysparklingIntegTest() {
...
+                    """
+                def fullSWVersion = sh(
+                        script: "echo \$( cat gradle.properties | grep version | grep -v '#' | sed -e \"s/.*=//\" )",
+                        returnStdout: true).trim()
+
+                def majorSWVersion = fullSWVersion.toString().substring(0, fullSWVersion.toString().lastIndexOf('_'))
+                def buildSWVersion = fullSWVersion.toString().substring(fullSWVersion.toString().lastIndexOf('_') + 1)
+                def artifactPath = "${env.WORKSPACE}/dist/build/sparkling-water-${fullSWVersion}.zip"
+
+                s3up {
```
mmalohlava: This is good, but main place to upload Sparkling Water artifacts should
be the same place where releases are uploaded including web page.

mmalohlava: s3up is for new projects mainly....

jenkins/sparklingWaterPipeline.groovy
jakubhava: good point, I can use groovy to read the content of the urls

jenkins/sparklingWaterPipeline.groovy
jakubhava: Ah, I see - the goals is to have nightly released in the same way as regular release except that we don't the the maven publish. I'll work on it.

I'm however a bit confused on `s3up` - was pointed to this by @michal-raska . Any alternative I could use ? :) ?

mmalohlava: `s3up` we wrote for dtbl, DAI. It does not fit into h2o-3 structure. for now i would use s3cmd directly.

mmalohlava: the plan is to slowly unify infrastructure tho share these pieces of code.

jakubhava:
oki, thanks! I didn't want to replicate the code so that was the main reason for going for s3up.

I will reuse the s3cmd from buildsparklingwater release script

jenkins/sparklingWaterPipeline.groovy
jakubhava: Oki, thanks! I didn't want to duplicate code so that's why I went for s3up.

I will reuse the code for sparkling water release script as we want to have the same page and release for nightly.

jenkins/Jenkinsfile-external
```diff
@@ -67,6 +67,7 @@ withCustomCommitStates(scm, 'h2o-ops-personal-auth-token', 'continuous-integrati
             backendMode = "external"
             hdpVersion = "${p.hdpVersion}"
             driverHadoopVersion = "${p.driverHadoopVersion}"
+            buildNightly = "true"
```
jakubhava: needs to be` false` here

PR #471, opened 2017-11-22
[SW-608] Measure time of conversions to H2OFrame in debug mode

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -202,7 +203,7 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
   /** Transforms Dataset[Supported type] to H2OFrame */
   def asH2OFrame[T<: Product : TypeTag](ds: Dataset[T]): H2OFrame = asH2OFrame(ds, None)
   def asH2OFrame[T<: Product : TypeTag](ds: Dataset[T], frameName: Option[String]): H2OFrame =
-    DatasetConverter.toH2OFrame(this, ds, frameName)
+    withConversionDebugPrints(sparkContext, "Dataset", DatasetConverter.toH2OFrame(this, ds, frameName))
```
mmalohlava: Curious question, could we use Spark to collect these stats? They have metric system in place already.

jakubhava: We could probably measure time of the conversion job and also the rest of related tasks. In Spark we can measure jobs, tasks at better granularity then we do now.

This measures the conversion and also all the related tasks in a quick easy solutionI can play with it and try to re-use spark for it, it would be nicer solution for sure.

This was just to have something quickly for the release :)

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -60,5 +62,18 @@ private[spark] trait H2OContextUtils extends Logging{
   def isRunningOnCorrectSpark(sc: SparkContext) = sc.version.startsWith(BuildInfo.buildSparkMajorVersion)
 
 
+  def withConversionDebugPrints[R <: Frame](sc: SparkContext, conversionName: String, block: => R): R = {
+    val sparkTesting = sc.getConf.getOption("spark.testing").orElse(sys.props.get("spark.testing"))
+
+    if (sparkTesting.nonEmpty && sparkTesting.get.toBoolean) {
```
mmalohlava: one more note: i would use different spark property then spark.testing - just to avoid different code path - i would say `spark.h2o.measurements.timing`

jakubhava: good point!

jakubhava:
@mmalohlava changed the property to better name.

I think that we do not need to rush with this change for this release and I can have a look if we can really reuse spark for obtaining this data. What do you think ?

jakubhava:
This should be in future handled by [SW-622]

PR #478, opened 2017-11-29
[SW-617] Support for exporting mojo to hdfs

miuma2:
I added a new feature to export the mojo model to HDFS. As discussed in the support ticket #91495 I am opening a pull request.


h2o-ops:
Hi @miuma2, your PR title "new feature/ exportMOJOModelToHDFS" is missing JIRA issue number! Please specify it in the following form [SW-]

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -61,6 +63,18 @@ trait ModelSerializationSupport {
...
   }
 
+  def exportMOJOModelToHDFS(model: Model[_, _, _], destination: URI, sc: SparkContext): URI = {
+    try {
+    val fs = FileSystem.get(sc.hadoopConfiguration)
+    val output = fs.create(new Path(destination))
+    val os = new BufferedOutputStream(output)
+    model.getMojo.writeTo(os)
+    } catch {
+      case e: IllegalArgumentException => e.printStackTrace
```
jakubhava: I would remove try catch and let the eventual exception propagate right away

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -61,6 +63,18 @@ trait ModelSerializationSupport {
     destination
   }
 
+  def exportMOJOModelToHDFS(model: Model[_, _, _], destination: URI, sc: SparkContext): URI = {
```
jakubhava: I would delete this method and integrate this with `exportMojoModel` method to have clean API.

The `exportMojoModel` can check the uri schema and execute hdfs specific code, otherwise run as normal

jakubhava:
Thanks for the PR. We should merge the 2 methods for exporting mojo

core/src/main/scala/water/support/ModelSerializationSupport.scala
miuma2: thanks for the feedback, that makes sense. I will update accordingly

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -56,22 +56,21 @@ trait ModelSerializationSupport {
...
-    try {
-    val fs = FileSystem.get(sc.hadoopConfiguration)
-    val output = fs.create(new Path(destination))
-    val os = new BufferedOutputStream(output)
-    model.getMojo.writeTo(os)
-    } catch {
-      case e: IllegalArgumentException =>
-        throw new IllegalArgumentException (s"Error while writing to HDFS (most likely no HDFS is available) in destination $destination : ${e.getMessage}")
+  def exportMOJOModel(model: Model[_, _, _], destination: URI,sc: SparkContext): URI = {
+    if (destination.getScheme()=="hdfs") {
```
jakubhava: missing spaces before and after `==`

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -56,22 +56,21 @@ trait ModelSerializationSupport {
...
+      try {
+        val fs = FileSystem.get(sc.hadoopConfiguration)
+        val output = fs.create(new Path(destination))
+        val os = new BufferedOutputStream(output)
+        model.getMojo.writeTo(os)
+      } catch {
+        case e: IllegalArgumentException =>
+          throw new IllegalArgumentException (s"Error while writing to HDFS (most likely no HDFS is available) in destination $destination : ${e.getMessage}")
+      }
+    }else{
```
jakubhava: missing spaces before and after `else`

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -56,22 +56,21 @@ trait ModelSerializationSupport {
...
-  def exportMOJOModelToHDFS(model: Model[_, _, _], destination: URI, sc: SparkContext): URI = {
-    try {
-    val fs = FileSystem.get(sc.hadoopConfiguration)
-    val output = fs.create(new Path(destination))
-    val os = new BufferedOutputStream(output)
-    model.getMojo.writeTo(os)
-    } catch {
-      case e: IllegalArgumentException =>
-        throw new IllegalArgumentException (s"Error while writing to HDFS (most likely no HDFS is available) in destination $destination : ${e.getMessage}")
+  def exportMOJOModel(model: Model[_, _, _], destination: URI,sc: SparkContext): URI = {
```
jakubhava: missing space after `URI`

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -54,10 +56,22 @@ trait ModelSerializationSupport {
...
+    if (destination.getScheme()=="hdfs") {
+      try {
+        val fs = FileSystem.get(sc.hadoopConfiguration)
+        val output = fs.create(new Path(destination))
+        val os = new BufferedOutputStream(output)
+        model.getMojo.writeTo(os)
+      } catch {
+        case e: IllegalArgumentException =>
+          throw new IllegalArgumentException (s"Error while writing to HDFS (most likely no HDFS is available) in destination $destination : ${e.getMessage}")
```
jakubhava: Can you pls wrap this message to 2 lines ? I think now it would fail and line too long during scalaCheck

core/src/main/scala/water/support/ModelSerializationSupport.scala
```diff
@@ -54,10 +56,23 @@ trait ModelSerializationSupport {
...
+        model.getMojo.writeTo(os)
+      } catch {
+        case e: IllegalArgumentException =>
+          throw new IllegalArgumentException (s"Error while writing to HDFS " +
+            s"(most likely no HDFS is available) in destination $destination : ${e.getMessage}")
+      }
+    } else {
+      val destFile = new File(destination)
+      val fos = new FileOutputStream(destFile)
+      model.getMojo.writeTo(fos)
```
mmalohlava: One more comment: we should probably have one more abstraction:
```
def exportModel(model, destination: OutputStream) {
...
}
```

The method `exportModel` would just created OS based on schema and pass to function ^.

Furthermore, the output stream handling should be handled by our Persist layer.

miuma2: Hi @mmalohlava,

Thanks for your feedback. 

do you mean a function that will return an OutputStream? something like that?

def exportModel(model, destination: URI):  OutputStream{
...
}

Regarding the second point, do you mean the exception handling? should I remove try/catch part?

miuma2:
@jakubhava , @mmalohlava ,

I realized that some of my commits were not mapped to my correct username. I updated the git config now but I would like to correct the author info for my previous commits (instead of TAAUSMI1 should be miuma2). They consider this steps as bad practice, but I would like to know if you have experience with it.
https://help.github.com/articles/changing-author-info/

Would that be ok for you, if I proceed to change that?

Thanks for your feedback.

core/src/main/scala/water/support/ModelSerializationSupport.scala
jakubhava: By Persistent storage we mean this code
```
    val p: Persist = H2O.getPM.getPersistForURI(destinationURI)
    val os: OutputStream = p.create(destination.toString, true)
```

It's used also in the `exportH2OModel` method, so you can have a look for inspiration.

jakubhava: @mmalohlava I'm a bit lost what you mean by `exportModel` method. We need to distinguish the cases for exporting MOJO, POJO and binary model.

jakubhava:
@miuma2 Seems like some kind of bad rebase/merge. Can you please have a look and ensure that only the your changes are here ?

miuma2:
Hi @jakubhava, yes the best will be to close this pull request and I will create a new one with a clean version.

PR #479, opened 2017-11-29
[SW-615] pysparkling.__version__ returns incorrectly 'SUBST_PROJECT_V…

jakubhava:
…ERSION'

### DIFF
--- a/py/pysparkling/__init__.py
+++ b/py/pysparkling/__init__.py
@@ -5,7 +5,19 @@
 pySparkling - The Sparkling-Water Python Package
 =====================
 """
-__version__ = "SUBST_PROJECT_VERSION"
+import zipfile
+from codecs import open
+from os import path
+
+here = path.abspath(path.dirname(__file__))
+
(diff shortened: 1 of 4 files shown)


py/pysparkling/__init__.py
```diff
@@ -5,7 +5,18 @@
 pySparkling - The Sparkling-Water Python Package
 =====================
 """
-__version__ = "SUBST_PROJECT_VERSION"
+import zipfile
+from codecs import open
+from os import path
+
+here = path.dirname(path.abspath(path.dirname(__file__)))
+if '.zip' in here:
```
mmalohlava: Did you test this path?

jakubhava: I test the zip path but right now I discovered few additional issues which I need to fix. 

Will let you know when this PR is completely ready

jakubhava: Tested both paths, also tested via testpypi to ensure that pip works

PR #482, opened 2017-12-01
[SW-XXX] H2o world hands on

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "H2o world hands on" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #492, opened 2017-12-11
[SW-629] REPL Jar list Spark 2.2.1 compatible

goosalex:
SPARK-24174 changes the way the REPL looks for Jars. 
This PR first tries the Spark >= v2.2.1 - way. if the respective config key is not present, the previous way.

### DIFF
--- a/repl/src/main/scala_2.10/org/apache/spark/repl/h2o/H2OInterpreter.scala
+++ b/repl/src/main/scala_2.10/org/apache/spark/repl/h2o/H2OInterpreter.scala
@@ -50,7 +50,15 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
       logWarning("ADD_JARS environment variable is deprecated, use --jar spark submit argument instead")
     }
     val jars = {
-      val userJars = Utils.getUserJars(conf, isShell = true)
+      val userJars = {
+        conf.getOption("spark.repl.local.jars") match { /* starting Spark 2.2.1. See SPARK-21714 */
+          case Some(s:String) => s.split(",").filter(_.nonEmpty).toSeq
+          case None => Utils.unionFileLists(
+            conf.getOption("spark.yarn.dist.jars"),
+            conf.getOption("spark.jars")
+          ).toSeq
+        }
+      }
       if (userJars.isEmpty) {
         envJars.getOrElse("")
       } else {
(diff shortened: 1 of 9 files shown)


h2o-ops:
Hi @goosalex, your PR title "Sw 629" is missing JIRA issue number! Please specify it in the following form [SW-]

goosalex:
Actually, I did not intend to merge Jakub's Commit too. Shall I close and open another PR?

jakubhava:
Ho @goosalex, thanks for the PR! What commits you have in mind ? I see only your commit, which should be fine.

However we need to do the same change here https://github.com/h2oai/sparkling-water/blob/master/repl/src/main/scala_2.10/org/apache/spark/repl/h2o/H2OInterpreter.scala - `H2OInterpreter` for Scala 2.10.

Can you please make sure this change is propagated there as well ?

Thx!

repl/src/main/scala_2.10/org/apache/spark/repl/h2o/H2OInterpreter.scala
```diff
@@ -50,7 +50,15 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
       logWarning("ADD_JARS environment variable is deprecated, use --jar spark submit argument instead")
     }
     val jars = {
-      val userJars = Utils.getUserJars(conf, isShell = true)
+      val userJars = {
```
jakubhava: @goosalex I think it would be cleaner to use reflection at this stage to call corresponding methods then specifying some list of dependencies (this might change without failing), also this does not mimic the exact Spark behaviour.

What I think would be the best to try calling `getLocalUserJarsForShell(conf)` if the method exists we can call it and can continue, if it doesn't we can fallback to Utils.getUserJars. This way, the internal implementation of these methods remains hidden, but we can work on both Spark 2.1.1 and 2.1.0 at the same time.

What do you think ?

repl/src/main/scala_2.11/org/apache/spark/repl/h2o/H2OInterpreter.scala
```diff
@@ -63,7 +63,16 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
     }
 
     val conf = sparkContext.getConf
-    val jars = Utils.getUserJars(conf, isShell = true).mkString(File.pathSeparator)
+    val localJars = {
+      conf.getOption("spark.repl.local.jars") match { /* starting Spark 2.2.1. See SPARK-21714 */
```
jakubhava: same comment as above

jakubhava:
@goosalex Thanks a lot for working on this. I was inspired by your change, however needed to put some additional code to the PR for Spark 2.2.1. So if you don't mind, I will close this PR and continue on https://github.com/h2oai/sparkling-water/pull/495.

Thanks for the contribution!

PR #497, opened 2017-12-15
[SW-637] Create new DataFrame with new schema when it actually contains dot

jakubhava:
…n any dot in names


Tagar:
Thank you @jakubhava . Would it be possible to push another Sparkling Water version out, with this patch in as a hot gix? Our data scientists reported that this issue is an "h2o down" severity for them as our company deals with wide datasets very often.

jakubhava:
Yup, we have a few more issues we need to put into the hot-fix first. I'll discuss it with the team and let you know as soon as possible.

PR #503, opened 2017-12-18
[SW-640] Make H2OMOJOModel.create_from_mojo available to PySparkling

jakubhava:


jakubhava:
Can do, I didn't want to duplicate the tests, but it's probably good to test it in the PySparkling env as well. thanks!

jakubhava:
Here's the JIRA https://0xdata.atlassian.net/browse/SW-642

PR #504, opened 2017-12-19
[SW-632] Dump full spark configuration during H2OContext.getOrCreate …

jakubhava:
…into DEBUG


mmalohlava:
Tiny detail, DEBUG is not default - we need to see that in our logs coming from users.

jakubhava:
Good point! But it can also annoy users if it's printed all the time, what do you think ? I can make it print into INFO level, then it would be printed any time we call `H2OContext.getOrCreate()`

jakubhava:
More I think about it, it probably won't hurt if you just print it every-time we start H2OContext.getOrCreate (as part of Log.INFO )

PR #510, opened 2017-12-20
[SW-645] Fix H2OInterpreter on Scala 2.10

jakubhava:


repl/src/main/scala_2.10/org/apache/spark/repl/h2o/H2OInterpreter.scala
```diff
@@ -101,4 +101,24 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
 
 object H2OInterpreter {
   def classOutputDirectory = H2OIMain.classOutputDirectory
+
+
+  def getUserJars(conf: SparkConf): Seq[String] = {
```
mmalohlava: WHy this method is duplicated by a method below?

jakubhava: It differs in few lines because of different APIs if Scala 2.10 and Scala 2.11

PR #514, opened 2017-12-22
[SW-647] The implementation fallback for Spark2.2.0 was not triggered

mmalohlava:


mmalohlava:
@jakubhava pls check the change - i expect it affects only Spark2.2 - am i right?

jakubhava:
Yup, this change needs to go only to rel-2.2

PR #526, opened 2018-01-09
[SW-666] Kill the cluster when a new executor joins in the internal …

jakubhava:
…backend

Still WIP, do not merge, tests are failing


core/build.gradle
```diff
@@ -75,6 +75,7 @@ dependencies {
   integTestCompile "org.scalatest:scalatest_${scalaBaseVersion}:2.2.1"
   integTestCompile "junit:junit:4.11"
 
+    testCompile "com.github.stefanbirkner:system-rules:1.17.0"
```
mmalohlava: Just curious, what is that?

jakubhava: this library is required for https://github.com/h2oai/sparkling-water/blob/77fe4aab020d1ebc96e991ceaeee4b1f4720137f/core/src/test/scala/org/apache/spark/h2o/NewExecutorJoinsTest.scala, especially the rule for checking system exit.

PR #529, opened 2018-01-11
[SW-273] Remove the workaround introduced by SW-272 as it's already f…

jakubhava:
…ixed in Spark 2.1.0 and later


jakubhava:
Manually tested on YARN cluster mode

PR #537, opened 2018-01-18
[SW-677] Release simplification and make more transparent

jakubhava:


build.gradle
```diff
@@ -21,9 +21,7 @@ buildscript {
     dependencies {
         classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.2'
         classpath 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.10:0.8.2'
-        classpath 'com.bmuschko:gradle-nexus-plugin:2.3.1'
         classpath 'net.researchgate:gradle-release:2.6.0'
-        classpath 'com.adaptc.gradle:nexus-workflow:0.6'
```
jakubhava: we haven't used this plugin at all

build.gradle
```diff
@@ -21,9 +21,7 @@ buildscript {
     dependencies {
         classpath 'com.github.jengelman.gradle.plugins:shadow:2.0.2'
         classpath 'org.github.ngbinh.scalastyle:gradle-scalastyle-plugin_2.10:0.8.2'
-        classpath 'com.bmuschko:gradle-nexus-plugin:2.3.1'
```
jakubhava: we used this just for creating additional artefacts for nexus upload.

Currently, we create the artefacts as well in the same way but for all the repositories -> i.e -> releasing to local maven or nexus will be always consistent

gradle/publish.gradle
```diff
@@ -1,67 +1,137 @@
...
 
-      // H2O licensing
-      licenses {
-        license {
-          name 'The Apache Software License, Version 2.0'
-          url  'http://www.apache.org/licenses/LICENSE-2.0.txt'
-          distribution 'repo'
+      /** Define the POM file */
+      pom.withXml { provider ->
+        def sparklingWaterPom = {
```
jakubhava: POM file remained the same as before

gradle/release.gradle
```diff
@@ -1,30 +1,21 @@
...
+    // Safe point - do releases only from release branch - can be deleted in future
+    failOnUnversionedFiles = false
+    failOnCommitNeeded = false
+    preCommitText = ":tada: "
+    git {
+        // Note: the releases are only allowed from master branch
+        requireBranch = '<<FILL ME>>'
+        // DEBUG: disable dry run in production/CI
+        // pushOptions = ["--dry-run"]
+    }
```
jakubhava: Separate the stages as there was some cradle bug not probably launching all `publish` tasks

gradle/publish.gradle
```diff
@@ -1,67 +1,137 @@
...
-
-task mappings {
-  doLast {
-    println conf2ScopeMappings.mappings
-    poms.each { pom ->
-      pom.writeTo("/tmp/x" + project.name)
-    }
-  }
-}
-// for publishing to maven local
```
mmalohlava: this is used for debugging

gradle/release.gradle
mmalohlava: ? follow documentation of publish plugin - it needs to be invoked per subproject

jakubhava: This PR didn’t go in, but publishing and releasing can be separated. Currently we get different artefacts when publishing to nexus and locally which is hard for debugging.

I know that in order to publish all the sub projects you need to income tasks on all of them, we just don’t do it automatically.

The purpose of this change was to clarify usage of the release plugins and make the release and creation of Pom more transparent and consistent. Currently we have some weird Pom files preventing us to load h2o and sw artefacts in dbc and this was attempt to debug/fix it

PR #539, opened 2018-01-18
[SW-681] Separate releasing on Github and Publishing artefacts

jakubhava:


mmalohlava:
?

jakubhava:
I just released using this version. 

The motivation was to have more fine grained release script as sometimes we do not want to do these two steps together.

I do not understand the second part - the release plugin updates the version to the new one, creates tag on GitHub and hen creates the new snapshot version. For all other tasks such us publish to s3 we change the version in gradle.properties so it is using correct release version.

jakubhava:
I mean publish to Nexus in the second paragraph. I hope it makes sense..

Let me please know if I should explain it more

mmalohlava:
Sorry, confusion on my side - makes sense now

PR #563, opened 2018-02-06
[SW-702] Initial implementation of H2OAutoML pipeline step

jakubhava:
Still work in progress, this is really just init phase

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
@@ -20,21 +20,24 @@ import java.io._
 import java.util.Date
 
 import ai.h2o.automl.{AutoML, AutoMLBuildSpec}
-import hex.deeplearning.DeepLearningModel.DeepLearningParameters
-import hex.schemas.DeepLearningV3.DeepLearningParametersV3
+import hex.{FrameSplitter, ScoreKeeper}
 import org.apache.hadoop.fs.Path
 import org.apache.spark.annotation.{DeveloperApi, Since}
 import org.apache.spark.h2o._
 import org.apache.spark.ml.h2o.models.H2OMOJOModel
-import org.apache.spark.ml.h2o.param.H2OAlgoParams
-import org.apache.spark.ml.param.{Param, ParamMap, Params}
+import org.apache.spark.ml.param._
 import org.apache.spark.ml.util._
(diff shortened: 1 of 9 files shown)


jakubhava:
Almost ready, Scala is passing, in Python, getters and setters are to be done

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -0,0 +1,386 @@
...
+    val aml = new AutoML(Key.make(uid), new Date(), spec)
+    AutoML.startAutoML(aml)
+    // Block until AutoML finishes
+    aml.get()
+    val model = new H2OMOJOModel(ModelSerializationSupport.getMojoData(aml.leader()))
+    model.setConvertUnknownCategoricalLevelsToNa(true)
+    model
+  }
+
+  private def split(fr: H2OFrame, hc: H2OContext): Array[Key[Frame]] = {
```
mmalohlava: Do we need to have split duplicated here? It is already implemented in core/src/main/scala/water/support/H2OFrameSupport.scala

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -0,0 +1,386 @@
...
+    val trainKey = Key.make[Frame]("train")
```
mmalohlava: What if `train` key already exist?

jakubhava: Good, point, thanks. I will fix this before the release

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: I will see if I can reuse the code, thanks!

jakubhava:
Yup, I just wanted to quickly test it and finally have it out. I can parametrize the ham_or_spam scripts in both Scala and Python to be parametrized. Will do this before the release as well. Thanks for the review!

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: Replaced also as part of https://github.com/h2oai/sparkling-water/pull/586

PR #579, opened 2018-02-15
[SW-725] create an an equvivalent of h2o.download_all_logs in scala

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -329,6 +329,17 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
     basic ++ sparkYarnAppId ++ backend.epilog
   }
 
+  /**
+    * @param path directory where the logs will be downloaded
+    */
+  def downloadH2OLogs(path: String, logFileName: String = "logs.zip"): String = {
```
jakubhava: `downloadH2OLogs` instead of `downloadAllLogs` is used on purpose as in the case of `downloadAllLogs` the user might be confused that it would download also Spark logs

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -329,6 +329,17 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
...
+    import sys.process._
+    import java.net.URL
+    import java.io.File
+    new URL(s"http://$localClientIp:$localClientPort/3/Logs/download").#>(new File(path, logFileName)).!!
```
mmalohlava: Please, improve it:
  - there is already distributed task collecting logs, please use it
   - let user to specify the location in the same way as we expose ModelSerializationSupport trait

jakubhava:
Good points, sorry for not giving this that much thoughts. These improvements make sense, thank you!

PR #587, opened 2018-02-21
[SW-727] Improve method for downloading H2O logs

jakubhava:


core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -77,6 +84,103 @@ private[spark] trait H2OContextUtils extends Logging {
...
 
+  /**
+    * @param destination directory where the logs will be downloaded
+    */
+  def downloadH2OLogs(destination: URI): URI = {
+    val perNodeZipByteArray = H2O.CLOUD.members.zipWithIndex.map { case (node, i) =>
+      // Skip nodes that aren't healthy, since they are likely to cause the entire process to hang.
+      try {
+        if (node.isHealthy) {
+          val g = new GetLogsFromNode
```
mmalohlava: interesting i would expect `new GetLogsFromNode(i).doIt().bytes()`

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -77,6 +84,103 @@ private[spark] trait H2OContextUtils extends Logging {
...
+  def downloadH2OLogs(destination: String): String = {
+    downloadH2OLogs(new URI(destination))
+    destination
+  }
+
+  /** Zip the H2O logs and return them as byte array */
+  private def zipLogs(results: Array[Array[Byte]], clientResult: Array[Byte], topDir: String, destination: URI): URI = {
+    assert(H2O.CLOUD._memary.length == results.length, "Unexpected change in the cloud!")
+    val l = results.map(_.length).sum
+
```
mmalohlava: Minor: 2 empty lines

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -77,6 +84,103 @@ private[spark] trait H2OContextUtils extends Logging {
...
+    val outputFileStem = "h2ologs_" + new SimpleDateFormat("yyyyMMdd_hhmmss").format(new Date)
+    zipLogs(perNodeZipByteArray, clientNodeByteArray, outputFileStem, destination)
+  }
+
+  def downloadH2OLogs(destination: String): String = {
+    downloadH2OLogs(new URI(destination))
+    destination
+  }
+
+  /** Zip the H2O logs and return them as byte array */
```
jakubhava: Wrong comment, we return the path to the downloaded logs

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Me too, but current H2O API does not allow that  and I wanted to limit the scope of this change just to SW

jakubhava:
thanks @mmalohlava !

PR #589, opened 2018-02-22
[SW-736] Document how to download logs on Databricks cluster

jakubhava:


jakubhava:
Thanks @angela0xdata!

PR #591, opened 2018-02-26
[SW-738] Move spark.ext.h2o.node.network.mask setter to SharedArguments

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "Move spark.ext.h2o.node.network.mask setter to SharedArguments" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #594, opened 2018-02-28
[SW-744] Document how to use Sparkling Water with LDAP in Sparkling W…

jakubhava:
…ater docs

### DIFF
--- a/doc/src/site/sphinx/tutorials/ldap.rst
+++ b/doc/src/site/sphinx/tutorials/ldap.rst
@@ -1,0 +1,87 @@
+Enabling LDAP
+-------------
+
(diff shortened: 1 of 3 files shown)


jakubhava:
@angela0xdata H2O documentation also contains steps how to set up LDAP on Sparkling Water. I would also maybe propose removing it from there and keeping it just here. H2O doc should be for H2O and SW for SW - what do you think?

jakubhava:
great, thanks @angela0xdata!

angela0xdata:
And I agree with keeping H2O docs only for H2O and SW docs only for SW. I think there might be a few other topics that are SW specific in the H2O-3 docs.

PR #601, opened 2018-03-07
[SW-752] Collect stack traces on each h2o node as part of log collect…

jakubhava:
…ing extension


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -224,6 +231,12 @@ object SharedBackendConf {
   /** Subnet selector for H2O nodes running inside executors - if the mask is specified then Spark network setup is not discussed. */
   val PROP_NODE_NETWORK_MASK = ("spark.ext.h2o.node.network.mask", None)
 
+  /** Enable or disable stack trace collector on each h2o node */
+  val PROP_NODE_STACK_TRACE_COLLECTOR_ENABLED = ("spark.ext.h2o.stacktrace.collector.enabled", false)
+
+  /** Set how often stack traces are taken on each h2o node */
+  val PROP_NODE_STACK_TRACE_COLLECTOR_INTERVAL = ("spark.ext.h2o.stacktrace.collector.interval", 20)
```
michalkurka: What unit is used for the interval? Minutes?

jakubhava: Seconds,  should be documented. Thx!

extension-stack-trace/src/main/java/ai/h2o/extensions/stacktrace/StackTraceExtension.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private class StackTraceCollectorThread extends Thread {
+    public StackTraceCollectorThread() {
+      super("StackTraceCollectorThread");
+    }
+
+    @Override
+    public void run() {
+      while (true) {
```
michalkurka: This doesn't do anything when debug is not enabled? How should we handle that? Should we produce an error message during the extension setup?

jakubhava: Maybe the ideal solution could be to log into `INFO` level when this is enabled so the user does not need to configure anything else

michalkurka: Good idea

extension-stack-trace/src/main/java/ai/h2o/extensions/stacktrace/StackTraceExtension.java
```diff
@@ -0,0 +1,104 @@
...
+import water.AbstractH2OExtension;
+import water.H2O;
+import water.util.Log;
+
+import java.util.Date;
+import java.util.Map;
+import java.util.Set;
+
+public class StackTraceExtension extends AbstractH2OExtension {
+  private int interval = 10; // 10 seconds
```
michalkurka: Very aggressive interval - maybe we can be more conservative? It will generate a lot of output in the logs

michalkurka: Idea: do not use any default - force the user to set a reasonable value that works for them

jakubhava: I would still prefer to set some default value, again to minimize the configuration, but agree that it should be higher.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -71,11 +71,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
       "-timeout", conf.clusterStartTimeout.toString,
       "-disown",
+      "-J", "-stacktrace_collector_interval", "-J", conf.stacktraceCollectorInterval.toString,
       "-J", "-watchdog_stop_without_client",
       "-J", "-watchdog_client_connect_timeout", "-J", conf.clientConnectionTimeout.toString,
       "-J", "-watchdog_client_retry_timeout", "-J", conf.clientCheckRetryTimeout.toString
     )
 
+    if(conf.stacktraceCollectorEnabled){
+      cmdToLaunch = cmdToLaunch ++ Seq[String]("-J", "-stacktrace_collector_enabled")
```
michalkurka: Do we need 2 arguments? Perhaps one argument that also sets the interval would make it simpler for everybody. I cannot image who would be happy with the default interval

michalkurka: Also there is a generic mechanism to disable H2O extensions.

jakubhava: Do you mean something like `-stacktrace_collector_timeout=-1` to disable the extension? I still prefer to have 2 arguments for the explicity.

core/build.gradle
```diff
@@ -61,6 +61,9 @@ dependencies {
   // Sparkling Water REPL
   compile(project(':sparkling-water-repl'))
 
+  // Extension for collecting the stack traces
+  runtime project(':sparkling-water-extension-stack-trace')
```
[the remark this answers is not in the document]
jakubhava: good point, fixed

extension-stack-trace/src/main/java/ai/h2o/extensions/stacktrace/StackTraceExtension.java
```diff
@@ -0,0 +1,104 @@
...
+          Map<Thread, StackTraceElement[]> allStackTraces = Thread.getAllStackTraces();
+          for (Map.Entry<Thread, StackTraceElement[]> e : allStackTraces.entrySet()) {
+            Log.debug("Taking stacktrace for thread: " + e.getKey());
+            for (StackTraceElement st : e.getValue()) {
+              Log.debug("\t" + st.toString());
+            }
+          }
+          sleep(interval * 1000);
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
jakubhava: This is here just for the good standards, not for any specific reasons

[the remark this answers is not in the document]
michalkurka: This thread is designed to never finish, I don't think this is necessary (I am not sure it is harmful though).

For non-cancellable tasks, you should call interrupt() when the work is done: see eg.: https://www.ibm.com/developerworks/library/j-jtp05236/index.html section Noncancelable tasks

However, your work is never done in this thread. I am leaning towards not calling interrupt()

PR #639, opened 2018-03-27
[SW-778] Handle nulls properly in H2OMojoModel

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -153,41 +153,43 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
...
-                put(f.name + idx, vector(idx).toString)
-              }
-            case vector: ml.linalg.Vector =>
-              (0 until vector.size).foreach { idx =>
-                put(f.name + idx, vector(idx).toString)
-              }
-          }
-        case null => // no op
-        case _ => put(f.name, get(idxRow).toString)
+      if(row.get(idxRow) != null) {
```
jakubhava: I just add this line to not include nulls in the output -> which ensures we treat them as missing values

PR #640, opened 2018-03-28
[SW-781] Added byte export of mojo to allow easy import.

gerbenoostra:
The `MojoModel.load` expects a zip file, which is not what ModelSerializationSupport exports, as it delegates it to the specific module. While the R and python exporters export to a zip file, the scala code does not. Therefore needed to change the export and import lines of code a little bit.

This MR adds a byte format export on ModelSerializationSupport, which allows easy import from h2o-genmodel.

Part of the solution to [SW-781] Valid export from sparkling-water

### DIFF
--- a/core/src/main/scala/water/support/ModelSerializationSupport.scala
+++ b/core/src/main/scala/water/support/ModelSerializationSupport.scala
@@ -94,8 +94,72 @@ object ModelSerializationSupport extends ModelSerializationSupport {
 
   def getMojoData(model: Model[_, _, _]) = {
     val baos = new ByteArrayOutputStream()
-    model.getMojo.writeTo(baos)
+    exportMojo(model, baos)
     baos.toByteArray
+  }
+
+  /**
+    * Exports the model mojo bytes to an (local) file.
+    *
+    * The file can then be used to create the mojo using ModelSerializationSupport (thus with sparkling-water):
+    * <pre>
+    * {@code
+    * ModelSerializationSupport.exportMojo(model, file)
+    * val bytes = java.nio.file.Files.readAllBytes(file.toPath)
+    * val mojo = ModelSerializationSupport.getMojoModel(bytes)
+    * }
+    * </pre>
+    *
+    * The file can also be used to create a mojo model without sparkling water (with just h2o-genmodel)
+    * <pre>
+    * {@code
+    * ModelSerializationSupport.exportMojo(model, file)
+    * val is = new FileInputStream(file)
+    * val reader = MojoReaderBackendFactory.createReaderBackend(is, MojoReaderBackendFactory.CachingStrategy.MEMORY)
+    * val mojoModel = ModelMojoReader.readFrom(reader)
+    * }
+    * </pre>
+    *
+    * Note that this does not generate a .zip file, which is expected by the `MojoModel.load(path)`
+    * @param model The model to serialize
+    * @param file  The location to write the model to
+    */
+  def exportMojo(model: Model[_, _, _], file: File) = {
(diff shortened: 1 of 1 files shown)


h2o-ops:
Hi @gerbenoostra, your PR title "Added byte export of mojo to allow easy import." is missing JIRA issue number! Please specify it in the following form [SW-]

jakubhava:
Hi @gerbenoostra, thanks for the PR, however, there is `exportMOJOModel` method in the `ModelSerializationSupport`, have you tried that? I think that this method should be compatible with the MojoModel.load.

jakubhava:
@gerbenoostra thanks for the PR. However, I just tested it and exportMOJOModel on the `water.support.ModelSerializationSupport._` trait can export model to the zip representation as:

```
import water.support.ModelSerializationSupport._
exportMOJOModel(model, "mojo.zip", false)
```

So this PR is not necessary. Do you agree?

gerbenoostra:
Excuse me but I don't agree.

The problem i faced was that the `exportMOJOModel(model, "mojo.zip", false)` writes in a format which cannot be read by `MojoModel.load("mojo.zip")`

The only way I could make the export-import roundtrip work, was by exporting the byte array representation of the contained mojo model, and then read it back with MojoModel's specific method that reads from a  byte array (stream).

For me, the following snippet
```
      import water.support.ModelSerializationSupport._
      exportMOJOModel(gbmModel, "mojo.zip", false)

      import hex.genmodel.easy.{EasyPredictModelWrapper, RowData}
      import hex.genmodel.MojoModel
      val mojoModel = MojoModel.load("mojo.zip")
```

fails with 

```
error in opening zip file
java.util.zip.ZipException: error in opening zip file
	at java.util.zip.ZipFile.open(Native Method)
	at java.util.zip.ZipFile.<init>(ZipFile.java:220)
	at java.util.zip.ZipFile.<init>(ZipFile.java:150)
	at java.util.zip.ZipFile.<init>(ZipFile.java:121)
	at hex.genmodel.ZipfileMojoReaderBackend.<init>(ZipfileMojoReaderBackend.java:13)
	at hex.genmodel.MojoModel.load(MojoModel.java:33)
```

on the MojoModel.load call

jakubhava:
@gerbenoostra I just verified the code works, I'm against duplicating the code in the code base. 

Maybe there is a different problem - Can you share what Sparkling Water version you are using for producing this MOJO and then what H2O version you are using for re-creating the model?

I was testing this on Sparkling Water 2.2.10 which is internally using H2O 3.18.0.4

jakubhava:
Also can you try the code I shared on http://h2o-release.s3.amazonaws.com/sparkling-water/rel-2.2/10/index.html ?

gerbenoostra:
i'm running in local mode in a unittest with dependencies on sparkling-water 2.2.9, also tried 2.1.23 (as for me starting h2o in 2.2.10 or 2.1.24 both fail with "ERROR: Unknown argument (-stacktrace_collector_interval) " as described in https://0xdata.atlassian.net/browse/SW-776)

let me check your code sample anyway too

gerbenoostra:
let met also create a unittest in sparkling-water to reproduce the exception

jakubhava:
Hi @gerbenoostra, thanks for the code! Could you please write these tests as part of the Sparkling Water project in another branch/PR based on master branch? It's then easy for me to verify the functionality pretty quickly.

Ideally 2 tests suites, the exact code I'm proposing and your code so we can see the differences in behavior. We have already bunch of tests in Sparkling Water, this would fit nicely in tests in sparkling-water-ml sub-project

Thanks!

PR #660, opened 2018-04-11
[SW-793] test: reproduction of problem with columns of wide data frames in pyspark

woj-i:
test: reproduction of problem with columns of wide data frames in pyspark, that affected 
till the commit id:
f50dd728281d11f9a2ab3cdaeb994644b892d65a
(included)

This test describes the problem and helps to prevent against it in future commits.

Sponsored by fingerprints.digital

### DIFF
--- a/py/tests/tests_unit_conversions.py
+++ b/py/tests/tests_unit_conversions.py
@@ -59,16 +59,16 @@ class FrameTransformationsTest(unittest.TestCase):
 
     # test transformation from wide dataframe to h2o frame and edit it
     def test_wide_df_to_h2o_frame_and_edit(self):
-        n_col = 210
+        n_col = 110
         test_data_frame = self._spark.createDataFrame([tuple(range(n_col))])
         h2o_frame = self._hc.as_h2o_frame(test_data_frame)
         self.assertEquals(h2o_frame.dim[1], n_col, "Number of cols should match")
         self.assertEquals(h2o_frame['_107'], 107, "Content of columns should be the same")
         # h2o_frame.refresh()     # this helps to pass the test
         # in commit f50dd728281d11f9a2ab3cdaeb994644b892d65a
-        col_202 = '_202'
+        col_102 = '_102'
         # replace a column after the column 100
-        h2o_frame[col_202] = h2o_frame[col_202].asfactor()
+        h2o_frame[col_102] = h2o_frame[col_102].asfactor()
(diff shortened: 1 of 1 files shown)


h2o-ops:
Hi @woj-i, your PR title "test: reproduction of problem with columns of wide data frames in pyspark" is missing JIRA issue number! Please specify it in the following form [SW-]

jakubhava:
Thanks for the test @woj-i! I forgot for which issue it is, can you please try to find a jira for the original issue in our JIRA?

woj-i:
Hi @jakubhava! I think it wasn't an issue in Jira. I met this problem when I worked on a previous release of Sparkling. I prepared the test and now, when I am allowed to commit it, I can see it pass in the most recent master branch, because it was (probably accidentally) solved by other changes.

jakubhava:
oki oki, can you please describe the issue?

woj-i:
Issue description: 
I have a wide data frame (hundreds of columns) and transform it to h2o frame. Then I want to convert one column of the h2o frame to factor, like:
`h2o_frame[col_target] = h2o_frame[col_target].asfactor()`
It cause I have a new column (as the end of the frame) instead of replacement of the old one.

woj-i:
Yes, I can reproduce it on that commit.

Technically, pysparkling had only the first N (200, as I remember) columns in the field name. When the column col_target was after the first N, then it was interpreted as a new column untill a call of h2o_frame.refresh()- after that more columns were downloaded and the new column was added to the end with the postfix "0".

jakubhava:
Ah now I see, it was fixed by this change https://0xdata.atlassian.net/browse/PUBDEV-5179

woj-i:
Yes, good catch! So now you have also a test for that. :) I hope it helps.

jakubhava:
Yup, having this test in SW is perfect! I wold however reduce the number of columns to 110 as 100 columns is the current limit in SW(to make the tests run faster, but still test the issue).

What do you think?

woj-i:
It make sense. I'll check if 110 cols also break tests in f50dd728281d11f9a2ab3cdaeb994644b892d65a and if so, I make this change.

jakubhava:
Thank you!

woj-i:
Checked- 110 columns works. Commited.

jakubhava:
Thanks for the contribution @woj-i !

PR #672, opened 2018-04-17
[SW-XXX] Pipeline instrumentation

jakubhava:
Extremely simple example, can be further extended:

To try out:
```
cd sparkling-water

export AGENT_PATH="$(pwd)/agent/build/libs/sparkling-water-agent_2.11-2.3.99999-SNAPSHOT-all.jar"

./bin/pysparkling --conf "spark.driver.extraJavaOptions=-javaagent:$AGENT_PATH" --conf "spark.executor.extraJavaOptions=-javaagent:$AGENT_PATH"
```

```
from pysparkling import *
from pysparkling.ml import H2OMOJOModel

mojo = H2OMOJOModel.create_from_mojo("ml/src/test/resources/deep_learning_airlines_categoricals.zip")
mojo.setConvertUnknownCategoricalLevelsToNa(True)
d =[{'sepal_len':5.1, 'sepal_wid':3.5, 'petal_len':1.4, 'petal_wid':0.2, 'class':'Missing_categorical'}]
df = spark.createDataFrame(d)
data = mojo.transform(df).collect()[0]

d =[{'sepal_len':5.1, 'sepal_wid':3.5, 'petal_len':1.4, 'petal_wid':0.2, 'class':'abc'}]
df = spark.createDataFrame(d)
data = mojo.transform(df).collect()[0]
```
We should see to error messages from Spark about unseen categoricals. This code was added using the instrumentation. It will work cluster-wise if we add the instrumentation agent to each Spark executor as well.
 
Currently, the agent does not have any dependency on H2O. I added dependency on Spark so we can use Spark logging and possibly other sinks where we can report data later.

This PR is not for merging, just for demonstrating the example

### DIFF
--- a/agent/build.gradle
+++ b/agent/build.gradle
@@ -1,0 +1,31 @@
+apply plugin: 'com.github.johnrengelman.shadow'
+
+description = "Sparkling Water Agent"
+
+dependencies{
+    compile "net.bytebuddy:byte-buddy:1.8.5"
+    compile "net.bytebuddy:byte-buddy-agent:1.8.5"
+    compile "org.apache.spark:spark-core_${scalaBaseVersion}:${sparkVersion}"
+}
+
+jar {
+    manifest {
+        attributes(
+                "Premain-Class": "Agent"
+        )
+    }
+}
+
+// Shadow jar settings
+shadowJar {
+    dependencies {
+        include(dependency("net.bytebuddy:byte-buddy"))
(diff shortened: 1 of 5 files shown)


mmalohlava:
Looks good. One question - do we want to go with agent? it is not so easy to use - i was thinking about some higher-level approach - reflection based - modifying the Spark pipeline.

PR #698, opened 2018-05-04
[SW-826] Rest API calls go through H2O Jobs

jakubhava:


gradle.properties
```diff
@@ -2,11 +2,11 @@
 group=ai.h2o
 version=2.3.99999-SNAPSHOT
 # Major version of H2O release
-h2oMajorVersion=3.20.0
+#h2oMajorVersion=3.20.0
```
jakubhava: TODO: before merging, ensure the gradle.properties do not use the temporary values

jakubhava:
Requires release of H2O 3.20.0.3

michalkurka:
Should this be released without the support for job queues?

jakubhava:
All the required code we need for this is in h2o and we just need to have release 3.20.0.3. The job queues are purely Sparkling Water change and I will also implement them in the same Sparkling Water release. 

So please just release 3.20.0.3, I will take care of the rest :) Thanks!

michalkurka:
Okay, I was just making sure. Because if it turns out there needs to be a change in h2o-core to support the queues we would need another release. Thx.

jakubhava:
perfect, thanks!

PR #699, opened 2018-05-04
[SW-831] Check shape of H2O Frame after conversion from Spark frame

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "[PUBDEV-5340] Check shape of H2O Frame after conversion from Spark frame" is missing JIRA issue number! Please specify it in the following form [SW-]

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
```diff
@@ -112,6 +112,22 @@ object WriteConverterCtxUtils {
           Log.debug(node + ": " + n + " chunks.")
       }
     }
+
+    val layouts = fr.vecs().map(_.espc())
+
+    if(!layouts.isEmpty){
```
michalkurka: What if ESPC is correct but the chunks itself are broken? Eg. there is a chunk that has less data than what the ESPC states. I think that might be the true cause of the bug

jakubhava: I see, good point. I will check also real shape of data

jakubhava:
Added more checks

PR #709, opened 2018-05-15
[SW-XXX] [PUBDEV-5340] Check shape of H2O Frame after conversion from Spark frame

jakubhava:
(cherry picked from commit c1cc85d38dca17eb0c34e405aa5f945411f8d215)


h2o-ops:
Hi @jakubhava, your PR title "[PUBDEV-5340] Check shape of H2O Frame after conversion from Spark frame" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #726, opened 2018-05-25
[SW-853] Sparkling Water py should require pyspark dependency

jakubhava:


Tagar:
@jakubhava what was the reason to add `pyspark` as a dependency?
We have `pyspark` as part of CDH Spark distro - it has its own `pyspark` version there, which might have some fixes on top of mainstream `pyspark`.. 
Another reason not to have `pyspark` as a dependency is because it brings a lot of things that we don't need, for example, it installs `beeline` binary that doesn't work for us - we have to use what's in our Cloudera distro again.. 
Any way we can remove this dependency?
Or as a workaround, when we install pysparkling using pip, I guess we could try playing with `--no-deps`, but again reverting this option would be ideal for us.
Thank you.

jakubhava:
Pip is easy way to install PySparkling. Semantically it makes sense to depend on Spark, as we really depend on it. It's required dependency.  In your case I would suggest installing PySparkling from the distribution zip file which can be downloaded from the web page.

Sorry, no current plans of reverting this change

Tagar:
Something like this should do 
```pip install --no-deps h2o-pysparkling-2.3 requests tabulate future colorama>=0.3.8 six```

So I took all the requirements as in 
https://github.com/h2oai/sparkling-water/blob/5a3042443ec4249b3ec798495e04f462d5ba0639/py/setup.py#L49
and just removed `pyspark`.

PR #727, opened 2018-05-25
[SW-850] Expose methods to get input/output names in H2OMOJOPipelineModel

jakubhava:


jakubhava:
TODO: tests

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
```diff
@@ -110,6 +110,39 @@ class H2OMOJOPipelineModel(val mojoData: Array[Byte], override val uid: String)
   override def transformSchema(schema: StructType): StructType = {
     StructType(schema ++ predictionSchema())
   }
+
+  def getInputNames(): Seq[String] = {
+    val inputMeta = getOrCreateModel().getInputMeta
+    (0 until inputMeta.size()).map(idx => inputMeta.getColumnName(idx))
```
mmalohlava: Btw: we should change MOJO api first to get column list directly.

jakubhava: agree, will do

PR #774, opened 2018-06-21
[SW-964] Azure Templates

JHerna17:
The only file that needs to be modified on each release is the head_node_installv2.sh with the appropriate minor versions for Spark 2.1, 2.2 and 2.3 (placeholder for now) 

Then it needs to be uploaded to our azure blob storage 
On Azure Government Cloud the access to s3 and github is restricted, hence the need to have these files on blob storage

### DIFF
--- a/templates/src/azure/Notebooks/ChicagoCrimeDemo.ipynb
+++ b/templates/src/azure/Notebooks/ChicagoCrimeDemo.ipynb
@@ -1,0 +1,550 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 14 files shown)


jakubhava:
@JHerna17 finally having a look today

JHerna17:
I will verify with HDI 2.3 before Monday

jakubhava:
thank you @JHerna17!

jakubhava:
Hi @JHerna17, I would like to go ahead and merge this. Can you please have a look and let me know if this is the final state? Thanks!

jakubhava:
Closing for now. We should get back to it

PR #796, opened 2018-07-02
[SW-898] Allow to use S3A import on Spark 2.3.x

jakubhava:


jakubhava:
This will unblock use of S3 when using the distribution downloaded from our page. The jar has less then 1mb so we are not making the final zip distribution significantly larger.

TODO: update windows files as well

mmalohlava:
I would revisit this PR.

What is core problem?

jakubhava:
@mmalohlava It's described on the associated Jira

PR #854, opened 2018-07-19
[SW-957] RSparkling should run tests in both, external and internal mode

jakubhava:


jakubhava:
R Unit Tests are running now in both modes, failing on several tests in external mode

PR #864, opened 2018-07-31
[SW-968] Remove confusing message about stopping H2OContext in PySparkling, proper stopping

jakubhava:


py/pysparkling/context.py
```diff
@@ -178,14 +178,12 @@ def __stop(self):
...
 
     def stop(self):
-        warnings.warn("Stopping H2OContext from PySparkling is not fully supported. Please restart your PySpark session and create a new H2OContext.")
+        self.__stop()
+        exit()
 
     def download_h2o_logs(self, destination):
         return self._jhc.h2oContext().downloadH2OLogs(destination)
     
-    def __del__(self):
```
jakubhava: use of ``__del__`` method is not recommended, it is more deterministic without it

PR #901, opened 2018-08-16
[SW-1000] Setting context path in pysparkling fails to launch h2o client

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "Setting context path in pysparkling fails to launch h2o client" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #921, opened 2018-09-13
[SW-1017] Add Parquet to list of supported data formats in the docs

angela0xdata:


angela0xdata:
@jakubhava I'll wait to merge this until after Orc is tested, in case we need to add that too.

jakubhava:
@angela0xdata let's merge this one now, I will need to test orc.

Thanks a lot for help!

PR #922, opened 2018-09-13
[SW-1018] Updates to R documentation files

angela0xdata:
- In the R Readme, fixed the link for Troubleshooting RSparkling on Windows. This topic was moved from Tutorials to Deployment.
- Fixed a link to the Main RSparkling Documentation. This pointed to a local rst file rather than the HTML file, resulting in a broken link.

### DIFF
--- a/doc/src/site/sphinx/deployment/rsparkling_on_windows.rst
+++ b/doc/src/site/sphinx/deployment/rsparkling_on_windows.rst
@@ -135,6 +135,6 @@ Troubleshooting
 References
 ~~~~~~~~~~
 
-- `Main RSparkling Documentation <docs.h2o.ai/sparkling-water/master/bleeding-edge/doc/rsparkling.html>`__
+- :ref:`rsparkling`
 - `H2O.ai website <http://h2o.ai>`__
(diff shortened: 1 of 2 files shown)


doc/src/site/sphinx/deployment/rsparkling_on_windows.rst
```diff
@@ -135,6 +135,6 @@ Troubleshooting
 References
 ~~~~~~~~~~
 
-- `Main RSparkling Documentation <../rsparkling.rst>`__
+- `Main RSparkling Documentation <docs.h2o.ai/sparkling-water/master/bleeding-edge/doc/rsparkling.html>`__
```
jakubhava: It would be easier to do it using the references in Spinx such as:

Main RSparkling Documentation :ref:`rsparkling_main` and definene the reference as `.. _rsparkling_main:`

This way we don't need to worry about changing the link for master and release branches. Could you please change it?

angela0xdata: Sure, that makes sense.

PR #925, opened 2018-09-14
[SW-1024] remove call to ./gradlew --help in jenkins pipeline

jakubhava:


jakubhava:
Just verified the call is still required. The call `.gradlew -q printH2OWheelPackage` fails as it also contains info from the Gradle changelog`

PR #977, opened 2018-10-18
[SW-1071] Fallback to original IP discovery in case we can't find the same network

jakubhava:


Tagar:
Would it be possible to release a hotfix SW 2.3.17 release with this fix sometime soon?
2.3.16 is broken for us. 
Thanks!

jakubhava:
@Tagar sorry, we plan to release according to our release schedule. If you need to use Sparkling Water with this fix you either need to build it yourself or use the nightly which can be downloaded from our page. The last nightly built was out on Wednesday so this fix is not in at the time I'm writing this, next nightly build will run this night

Tagar:
Ok, we will have to wait for next release then. 
Hopefully the reconnect issue will be fixed in that release too.
We will stay on 2.3.13 for now.
Thanks @jakubhava .

PR #995, opened 2018-12-04
[SW-995] Don't require implicit sqlContext parameter as part of asDataFrame as we can get it in spark session internally

jakubhava:


derSascha:
Is there a reason why this was cherry-picked on to the 2.1, 2.2 and 2.3 branch? This brakes the JavaH2OContext API :/ Any chance to revert this on versions below 2.4?

jakubhava:
Hi @derSascha, this was a planned change and we are not planning reverting the changes, please update your code to the new version.

We tried to ensure the old code is deprecated, but maybe we missed a spot. We will try to add deprecated method for this but just until next major release where it will be removed. 

Thanks for understanding

derSascha:
Hi @jakubhava, thanks for the fast answer. We need to support H2O 3.22.0.2 and 3.24.0.4 with Spark 2.3 and Spark 2.4. It's OK to change this between Sparkling Water 2.3 and 2.4, but not between 2.3.27 and 2.3.30... This makes it impossible to support different H2O versions with the same Spark 2.3 code base. Any idea how to build a workaround to handle the change from `asDataFrame(String key, SQLContext sqlContext)` to `asDataFrame(String key)`?

jakubhava:
Sparkling Water Major releases are driven by H2O major releases, not by Spark ones. 

As I mentioned, we will see if we car create a temporary deprecated method, but please be aware, that method will be removed in the next major release - around July. If you would like to use newer versions, you need to update your codebase.

jakubhava:
https://0xdata.atlassian.net/browse/SW-1318 - however only permanent solution is to update the codebase Thanks for pointing this out!

jakubhava:
@derSascha this will be fixed in the next minor release of Sparkling Water, PR is already done. However as I mentioned, I suggest upgrading to the newer API if you want to use newer versions after the major release. Thank you

derSascha:
@jakubhava thanks for fixing this.

PR #1005, opened 2018-12-13
[SW-1099] Fix GCS support on external backend

jakubhava:


jakubhava:
Closing as https://github.com/h2oai/h2o-3/pull/3137 will provide fix directly in H2O

PR #1054, opened 2019-01-22
[SW-369] Use only Spark nodes with H2O for H2O-related computations

jakubhava:


jakubhava:
Just reopened the old PR to see if it is still relevant

jakubhava:
Attempting to write a test

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
```diff
@@ -59,7 +62,7 @@ object WriteConverterCtxUtils {
     * Converts the RDD to H2O Frame using specified conversion function
     *
     * @param hc            H2O context
-    * @param rdd           rdd to convert
+    * @param rddInput           rdd to convert
```
mn-mikke: nit: indentation

core/src/main/scala/org/apache/spark/rdd/h2o/H2OAwareRDD.scala
```diff
@@ -14,20 +14,21 @@
...
-    Log.err(s"Executor ${executorAdded.executorId} without H2O instance discovered, killing the cloud!")
-    H2O.shutdown(-1)
-  }
+class H2OAwareRDD[U: ClassTag](nodes: Seq[String], prev: RDD[U]) extends RDD[U](prev: RDD[U]) {
+
+  override def getPreferredLocations(split: Partition): Seq[String] = nodes
 
+  override def compute(split: Partition, context: TaskContext): Iterator[U] = prev.compute(split, context)
+
+  override protected def getPartitions: Array[Partition] = {
```
mn-mikke: nit: I would prefer one line without curly brackets

core/src/main/scala/org/apache/spark/rdd/h2o/H2OAwareRDD.scala
```diff
@@ -14,20 +14,21 @@
...
 
-class ExecutorAddNotSupportedListener extends H2OSparkListener {
 
-  override def onExecutorAdded(executorAdded: SparkListenerExecutorAdded): Unit = {
-    Log.err(s"Executor ${executorAdded.executorId} without H2O instance discovered, killing the cloud!")
-    H2O.shutdown(-1)
-  }
+class H2OAwareRDD[U: ClassTag](nodes: Seq[String], prev: RDD[U]) extends RDD[U](prev: RDD[U]) {
+
+  override def getPreferredLocations(split: Partition): Seq[String] = nodes
```
mn-mikke: Does this method ensure that the logic will be always executed on the list of specified nodes? Aren't there any edge cases?

jakubhava: Tests went fine, but still need to check the actual implementation, good questions

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendUtils.scala
```diff
@@ -134,7 +134,6 @@ private[internal] trait InternalBackendUtils extends SharedBackendUtils {
     // Create global accumulator for list of nodes IP:PORT
     val bc = sc.collectionAccumulator[NodeDesc]
     val isLocal = sc.isLocal
-    val userSpecifiedCloudSize = sc.getConf.getOption("spark.executor.instances").map(_.toInt)
```
jakubhava: totally unused variable

core/src/main/scala/org/apache/spark/rdd/h2o/H2OAwareRDD.scala
jakubhava: More context -> It's not guaranteed for 100%. Spark will try to use the preferred locations, but only for duration of `spark.locality.wait` timeout. If the node preferred note is not available till then, it will try to launch the task on different executor.

We can set `spark.locality.wait=0` to ensure we will just try to use the right nodes. This does not affect as as we truly need to run tasks on nodes with H2O( and don't support dynamic allocation) not the other ones - so the behavior will still be the same as it is now

I think we will get rid of this anyway after the big stabilization effort.

PR #1075, opened 2019-02-20
[SW-1052] Document terraform templates

jakubhava:


jakubhava:
@angela0xdata could I please ask you for a review?

angela0xdata:
@jakubhava I made some changes and built locally. Everything looks good, but let me know if there are any issues with my edits.

doc/src/site/sphinx/deployment/terraform_emr.rst
```diff
@@ -1,70 +1,69 @@
...
+   - ``sw_patch_version`` (optional) - Sparkling Water minor version. Defaults to SUBST_SW_MINOR_VERSION.
 
 
-We provide these 3 templates as we realize some users have already some network infrastructure on their
-AWS clusters available and they don't want to use template which create it again.
+We provide these 3 templates as we realize some users already have some network infrastructure on their
+AWS clusters available, and they don't want to use a template that creates it again.
 
-To start any template you want, please run
+To start any template you want, please run the following in the corresponding **/templates/build/terraform/aws/modules/** directory:
```
jakubhava: Here the path is not valid for all the 3 templates because the `default` one is inside `/templates/build/terraform/aws`. I would just mention here something like `in the corresponding module directory`

jakubhava:
@angela0xdata looks great, thanks a ton for help! Just one small thing which we need to change

doc/src/site/sphinx/deployment/terraform_emr.rst
angela0xdata: Done. Just pushed a change.

jakubhava:
Thank you @angela0xdata!

PR #1117, opened 2019-04-04
[SW-1186] No need to pass properties defined in spark-defaults.conf to cli

jakubhava:
There is no need to additionally pass this configuration to CLI as Spark reads the spark-defaults.conf file


mn-mikke:
What was the original reason for populating ```spark.driver.extraJavaOptions``` explicitly?

jakubhava:
I think it was because we needed to add some additional values to `spark.driver.extraJavaOptions` internally in our scripts. Such as `-XX:MaxPermSize`

In the current code-base we do not need to pass any additional values.

From user point of view the behavior in case someone explicitly overrides this option as:

```
./bin/sparkling-shell --conf "spark.driver.extraJavaOptions=..." remains the same.

jakubhava:
thx @mn-mikke !

PR #1118, opened 2019-04-04
[SW-1189] Fix Sparkling Water 2.1.x compile on Scala 2.10

jakubhava:
Sparkling Water 2.1 on Scala 2.10 was failing to compile as it required more strict type variables. Sparkling Water compiles well on Scala 2.11 with the current code, but it does not hurt to be consistent and provide the types also on Scala 2.11.

That is why I'm applying this change on to master and later we should cherry-pick to all rel-2.1, rel-2.2, rel-2.3 and rel-2.4 branches

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
@@ -57,7 +57,7 @@ class H2OGridSearch(val gridSearchParams: Option[H2OGridSearchParams], override 
 
   def this(uid: String, hc: H2OContext, sqlContext: SQLContext) = this(None, uid)(hc, sqlContext)
 
-  private var grid: Grid[_] = _
+  private var grid: Grid[_ <: Model.Parameters] = _
   private var gridModels: Array[Model[_, _ <: Model.Parameters, _ <: Model.Output]] = _
   private var gridMojoModels: Array[H2OMOJOModel] = _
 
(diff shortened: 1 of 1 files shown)


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -195,7 +195,7 @@ class H2OGridSearch(val gridSearchParams: Option[H2OGridSearchParams], override
     }
   }
 
-  private def sortGrid(grid: Grid[_]) = {
+  private def sortGrid(grid: Grid[ _ <: Model.Parameters]): Array[hex.Model[_, _ <: hex.Model.Parameters, _ <: hex.Model.Output]] = {
```
mn-mikke: Is the specification of the return type really needed?

jakubhava: just tried - SW 2.1 on Scala 2.10 won't compile without this output type specified.

The automatically inferred types have slightly different type bounds leading to fail during compilation

mn-mikke: WDYT about using a type alias for the result type? IMHO, it would improve readability.

jakubhava: Good point, especially for the long `hex.Model[_, _ <: hex.Model.Parameters, _ <: hex.Model.Output]`. Will do, thx!

mn-mikke: +1 , Hope there won't be any restrictions

jakubhava:
@mn-mikke added the type alias for `hex.Model[_, _ <: hex.Model.Parameters, _ <: hex.Model.Output]`

mn-mikke:
Really like the current implementation! :-)  Is ```hex.Model[_, _ <: hex.Model.Parameters, _ <: hex.Model.Output]``` used also at other places?

jakubhava:
Yup, it is. I was thinking yesterday about this and it would be great to have this alias defined in `h2o` package object and used it everywhere where this type is used. What do you if I create new JIRA for this?

jakubhava:
https://0xdata.atlassian.net/browse/SW-1191

PR #1122, opened 2019-04-08
[SW-1194] RSparkling Can't be used on Spark 2.4

jakubhava:


r/src/R/package.R
```diff
@@ -74,7 +74,7 @@ verify_h2o_version <- function(spark_version, sw_version) {
 #' @export
 check_spark_version_for_any_sw <- function(spark_version) {
   spark_major_minor <- get_spark_major_minor_version(spark_version)
-  supported_spark_versions <- c("2.3", "2.2", "2.1", "2.0", "1.6")
+  supported_spark_versions <- c("2.4", "2.3", "2.2", "2.1")
```
jakubhava: 2.4 was missing here. Also 2.0 and 1.6 are not supported for long time

PR #1123, opened 2019-04-09
[SW-1195] Disable gradle daemon

mn-mikke:
This PR disables execution of the gradle daemon for the whole project.


jakubhava:
When cherry-picking to rel branches, we might remove explicit disable of Gradle daemon in release pipelines as it is not necessary since now

PR #1131, opened 2019-04-12
[SW-1198] Implement an annotation deprecating legacy methods

mn-mikke:
The PR introduce a new annotation for deprecating legacy methods in Sparkling Water API. When a user uses an annotated method with such a class, he/she will receive a warning during compile and also during runtime when the method is called.


jakubhava:
@mn-mikke Can you pls fix the link issues? Otherwise looks great, thx!

jakubhava:
Looks good, feel free to merge 👍

PR #1133, opened 2019-04-15
[SW-1204] Fix formating in python tests

jakubhava:
No semantic changes -> just clicked on reformat code in Idea in test utilities and tests files themselfs

### DIFF
--- a/py/tests/generic_test_utils.py
+++ b/py/tests/generic_test_utils.py
@@ -23,7 +23,8 @@ from random import randrange
 
 
 def unique_cloud_name(script_name):
-    return str(script_name[:-3].replace("/", "_"))+str(randrange(65536))
+    return str(script_name[:-3].replace("/", "_")) + str(randrange(65536))
+
 
 def locate(file_name):
     if os.path.isfile("/home/0xdiag/" + file_name):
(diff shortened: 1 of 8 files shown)


py/tests/generic_test_utils.py
```diff
@@ -23,24 +23,28 @@
...
 
 def locate(file_name):
     if os.path.isfile("/home/0xdiag/" + file_name):
         return os.path.abspath("/home/0xdiag/" + file_name)
     else:
         return os.path.abspath("../examples/" + file_name)
 
-def cluster_mode(spark_conf = None):
+
+def cluster_mode(spark_conf=None):
```
mn-mikke: I would prefer the original version with spaces or is it more frequent like this in Python?

jakubhava:
Yup, PEP8 recommends to not have spaces around `=`. Citing:

```
Don't use spaces around the = sign when used to indicate a keyword argument, or when used to indicate a default value for an unannotated function parameter.
``` from https://www.python.org/dev/peps/pep-0008/#other-recommendations

**Comment by jakubhava** [comment]
Tests failing at: 

```
org.apache.spark.ml.spark.models.H2OAlgoTest > Test H2OGLM Pipeline FAILED
    java.lang.NullPointerException at H2OAlgoTest.scala:33
```

PR #1135, opened 2019-04-16
[SW-1203] Custom metric not evaluated in internal mode of Sparkling Water

jakubhava:
Add support for custom metric on the python client in Sparkling Water by including the dependencies to the assembly


jakubhava:
It will currently fail as the necessary changes will be in H2O 3.24.0.2

mn-mikke:
@jakubhava  IMHO, the test file is still too big 1MB for having it in the repo. WDYT?

jakubhava:
ok, I will make it half, but that is enough by me

PR #1142, opened 2019-04-17
[SW-1209] Rename the 'predictionCol' model parameter to 'labelCol'

mn-mikke:
The PR renames the ```'predictionCol``` parameter to ```labelCol```  for all SW algorithms.


ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParams.scala
```diff
@@ -90,16 +91,20 @@ trait H2OAlgoParams[P <: Parameters] extends H2OAlgoParamsHelper[P] with Logging
...
 
   /** @group getParam */
   def getWeightCol(): String = $(weightCol)
 
   /** @group getParam */
   def getFeaturesCols(): Array[String] = {
-    if ($(featuresCols).contains($(predictionCol))) {
-      logDebug("Prediction col '" + $(predictionCol) + "' removed from the list of features.")
-      $(featuresCols).filter(_ != $(predictionCol))
+    val lblCol = getLabelCol()
```
jakubhava: Minor: I would prefer `labelCol` instead of `lblCol`

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -241,7 +241,7 @@ trait H2OAutoMLParams extends Params with Logging {
   // Default values
   //
   setDefault(
-    predictionCol -> "predictionCol",
+    labelCol -> "labelCol",
```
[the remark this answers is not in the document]
mn-mikke: Good spot!

ml/src/test/resources/sms_pipeline_deployment/sms_pipeline.model/stages/4_mojoModel_4886dd97b3a0/metadata/part-00000
```diff
@@ -1 +1 @@
-{"class":"org.apache.spark.ml.h2o.models.H2OMOJOModel","timestamp":1513945002816,"sparkVersion":"2.2.1","uid":"mojoModel_4886dd97b3a0","paramMap":{"predictionCol":"label","outputCol":"prediction_output","featuresCols":["tf_idf"]}}
+{"class":"org.apache.spark.ml.h2o.models.H2OMOJOModel","timestamp":1513945002816,"sparkVersion":"2.2.1","uid":"mojoModel_4886dd97b3a0","paramMap":{"labelCol":"label","outputCol":"prediction_output","featuresCols":["tf_idf"]}}
```
mn-mikke: @jakubhava Shouldn't we also support backward compatibility of MOJO files when we rename parameters?

jakubhava: Good point, yeah. At this point I would support the backwards compatibility between in between minor releases. New Major release should be able to break this ( and allow us to clean). WDYT?

mn-mikke: Yeah, it makes sense. Should we also remove the deprecated properties at the same time?

mn-mikke: I'm thinking about solving the backwards in a different PR. WDYT?

jakubhava: yup, let's make a new PR

mn-mikke:
Rebased to the master branch, waiting for tests

mn-mikke:
It seems that  this is failing because of missing backwards compatibility of MOJO files

PR #1143, opened 2019-04-17
[SW-1215] Automatically generate PySparkling README

jakubhava:


jakubhava:
When new Spark release is out, there is need of manually updating versions in PySparkling readme.

This PR avoids doing that as during each build we update the documentation according to changes.

When cherry-picking to `rel-x` branches, I will also edit release pipeline to commit the updated Readme so the whole process is automated

py/setup.py
```diff
@@ -6,7 +6,7 @@
 here = path.abspath(path.dirname(__file__))
 
 # Get the long description from the relevant file
-with open(path.join(here, 'README.rst'), encoding='utf-8') as f:
+with open(path.join(here, 'README_a.rst'), encoding='utf-8') as f:
```
mn-mikke: Why it's changed?

py/README_TEMPLATE.rst
```diff
@@ -0,0 +1,171 @@
...
+
+The driver program in Sparkling Water, creates a ``SparkSession`` (spark) which in turn is used to create an ``H2OContext`` (hc) that is used to start H2O services on the spark executors. An H2O Context is a connection to H2O cluster and  also facilitates communication between H2O and Spark. When an H2O cluster starts, it has the same topology as the Spark cluster and H2O nodes shares the same JVMs as the Spark Executors.
+
+To leverage H2O's algorithms, data in Spark cluster, stored as an RDD, needs to be converted to an H2OFrame (H2O's distributed data frame). This requires a data copy because of the difference in data layout in Spark (blocks/rows) and H2O (columns). But as data is stored in H2O in a highly compressed format, the overhead of making a data copy is low. When converting an H2OFrame to RDD, Sparkling Water creates a wrapper around the H2OFrame to provide an RDD-like API. In this case, no data is duplicated and data is served directly from the underlying H2OFrame. As H2O runs in the same JVMs as the Spark Executors, moving data from Spark to H2O or vice-versa requires a simple in memory, in process call.
```
mn-mikke: Shouldn't we rather talk about dataframes/datasets than RDDs?

py/README_TEMPLATE.rst
```diff
@@ -0,0 +1,171 @@
...
+
+
+ or download and unpack sparkling water release from  `here <https://www.h2o.ai/download/>`_.
+
+2. Configure the location of Spark distribution and cluster:
+
+  .. code-block:: bash
+
+    export SPARK_HOME="/path/to/spark/installation"
+    export MASTER='local[*]'
```
mn-mikke: Isn't this configuration just for developers running Spark on laptops?

py/setup.py
jakubhava: Debugging change leaked in, should be `README`, thx

py/README_TEMPLATE.rst
jakubhava: Yup, agree, however I would create a new JIRA to improve the PySparkling doc. What do you think?

py/setup.py
jakubhava: Fixed

py/README_TEMPLATE.rst
jakubhava: https://0xdata.atlassian.net/browse/SW-1216

py/README_TEMPLATE.rst
jakubhava: Yup, this doc should be for a quick start. We can mention also yarn etc. I would also do that in another JIRA as it's not target of this one, what do you think? :)

jakubhava:
Let's have a look on the issues you pointed out in a new jira. Great feedback!

PR #1146, opened 2019-04-18
[SW-1219] Remove suport for testing external cluster in manual mode

jakubhava:
Minimal changes required to get right of unused call


core/src/test/scala/org/apache/spark/h2o/utils/SharedH2OTestContext.scala
```diff
@@ -49,7 +49,7 @@ trait SharedH2OTestContext extends SparkTestContext {
     try {
```
mn-mikke: What line could throw a ```SecurityException```?

jakubhava: I believe `System.setSecurityManager(securityManager)`, would need to check, but it's not target of this PR :)

mn-mikke: Shouldn't be:
```
hc.stop()
hc = null
```
a part of finally statement then? Or is ```System.setSecurityManager(securityManager)``` a prerequisite for that?

jakubhava: I think it's not target of this PR to discuss this particular part of code as nothing has been changed regarding this. Please create JIRA if something doesn't seem right and let's discuss & have a look later 👍

jakubhava: but in short, yes. You're right!

jakubhava: I'll add `hc = null` this line in this PR

mn-mikke: Yeah, np. :-) We can answer this question in a dedicate PR.

jakubhava: Cool! Step by step :) I just want to avoid changing to much in one PR. But you raise good questions, thanks for that!

PR #1152, opened 2019-04-18
[SW-1226] Mechanism for enabling backward compatibility of MOJO files

mn-mikke:


mn-mikke:
@jakubhava I played several hours with writing tests covering this problem, but didn't come up with something reasonable. Do you have any idea?

jakubhava:
Rebased on top of master. The master should fail and this PR should pass

jakubhava:
Passing. @mn-mikke . Feel free to merge

PR #1154, opened 2019-04-23
[SW-1230] Fix typo in GLM API and rename 'removeCollinearColumns' to 'removeCollinearCols'

mn-mikke:


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGLM.scala
```diff
@@ -109,7 +109,7 @@ trait H2OGLMParams extends H2OAlgoParams[GLMParameters] {
   final val gradientEpsilon = doubleParam("gradientEpsilon")
   final val objReg = doubleParam("objReg")
   final val computePValues = booleanParam("computePValues")
-  final val removeCollinearColumns = booleanParam("removeCollinearColumns")
+  final val removeCollinearCols = booleanParam("removeCollinearCols", "A flag indicating whether collinear columns should be removed or not")
```
jakubhava: We need to deprecate on the mojo level as well

jakubhava: Please ignore the comment above ^

PR #1155, opened 2019-04-23
[SW-1229][rel-2.1] Remove dependencies to Scala 2.10

mn-mikke:


mn-mikke:
Good point! Thanks a lot.

PR #1159, opened 2019-04-23
[SW-1232] Fix RUnits after upgrading to Gradle 5.3.1

jakubhava:


assembly/build.gradle
```diff
@@ -22,7 +22,7 @@ dependencies {
 
 jar {
   enabled = false // we do not need to generate jar file in this case
-  archiveBaseName = "${jar.archiveBaseName}_${scalaBaseVersion}"
+  archiveBaseName = "${jar.archiveBaseName.get()}_${scalaBaseVersion}"
```
jakubhava: crucial bit

gradle/r.gradle
```diff
@@ -1,18 +1 @@
 apply plugin: 'base'
-
```
jakubhava: clean up - no need

PR #1174, opened 2019-04-26
[SW-1249] probability calibration does not work in Sparkling Water Dataframe API

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -78,9 +79,23 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
   }
 
 
-  implicit def toBinomialPrediction(pred: AbstractPrediction) = BinomialPrediction(
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(0),
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(1))
+  implicit def toBinomialPrediction(pred: AbstractPrediction): BinomialPrediction = {
+    if(pred.asInstanceOf[BinomialModelPrediction].calibratedClassProbabilities != null) {
```
jakubhava: Calibrated probabilities are set up only by specific algorithms, otherwise they are null

py/tests/tests_unit_conversions.py
```diff
@@ -229,8 +229,12 @@ def test_load_mojo_gbm(self):
         pred_model = model.transform(prostate_frame).repartition(1).collect()
 
         self.assertEquals(len(pred_mojo), len(pred_model))
+        num_cols = len(pred_mojo[0])
         for i in range(0, len(pred_mojo)):
-            self.assertEquals(pred_mojo[i], pred_model[i])
+            for k in range(0, num_cols - 2):
+                self.assertEquals(pred_mojo[i][k], pred_model[i][k])
+            self.assertEquals(pred_mojo[i][num_cols - 1][0], pred_model[i][num_cols - 1][0])
+            self.assertEquals(pred_mojo[i][num_cols - 1][1], pred_model[i][num_cols - 1][1])
```
jakubhava: because the p0_calibrated & p1_calibrated are `nan` and `nan == nan` return `false` in Python

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -64,7 +64,8 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
 
   def predictionSchema(): Seq[StructField] = {
     val fields = getOrCreateEasyModelWrapper().getModelCategory match {
-      case ModelCategory.Binomial => StructField("p0", DoubleType) :: StructField("p1", DoubleType) :: Nil
+      case ModelCategory.Binomial => StructField("p0", DoubleType) :: StructField("p1", DoubleType) ::
```
mn-mikke: nit: WDYT about following?
```
Seq(
  StructField("p0", DoubleType) ,
  StructField("p1", DoubleType),
  StructField("p0_calibrated", DoubleType),
  StructField("p1_calibrated", DoubleType))
```

or
```Seq("p0", "p1", "p0_calibrated", "p1_calibrated").map(StructField(_, DoubleType))```

Comment (mn-mikke) [inline ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala]: Can ```p0```, ```p1``` and other columns contain ```null``` values?

Comment (jakubhava) [reply ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala]: They can, do you think that is better solution than `nan`s? I was thinking about it but no strong opinion. `null` seemed to me a bit low-level, but otherwise no preference

Comment (jakubhava) [reply ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala]: Love it

File: ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala [hunk 4]
```diff
@@ -78,9 +79,23 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
...
+    } else {
+      BinomialPrediction(
+        pred.asInstanceOf[BinomialModelPrediction].classProbabilities(0),
+        pred.asInstanceOf[BinomialModelPrediction].classProbabilities(1),
+        Double.NaN,
+        Double.NaN
+      )
+    }
+
+  }
```
deil87: I would even reduce this to
```
implicit def toBinomialPrediction(pred: AbstractPrediction): BinomialPrediction = {
    val prediction = pred.asInstanceOf[BinomialModelPrediction]
      BinomialPrediction(
        prediction.classProbabilities(0),
        prediction.classProbabilities(1),
        if(prediction.calibratedClassProbabilities !=null) prediction.calibratedClassProbabilities(0) else Double.NaN,
        if(prediction.calibratedClassProbabilities !=null) prediction.calibratedClassProbabilities(1) else Double.NaN
      )
    }
```

deil87: or even to this as scope is small:
```
implicit def toBinomialPrediction(pred: AbstractPrediction): BinomialPrediction = {
    val prediction = pred.asInstanceOf[BinomialModelPrediction]
    val ccp = prediction.calibratedClassProbabilities
    BinomialPrediction(
        prediction.classProbabilities(0),
        prediction.classProbabilities(1),
        if(ccp !=null) ccp(0) else Double.NaN,
        if(ccp !=null) ccp(1) else Double.NaN
      )
    }
```

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
mn-mikke: I'm asking because you don't specify nullability  in the constructor of ```StructField```. The default value is ```true```.  Spark optimizes generated whole-stage code if nullability is set to ```false``` (removes statements like ```if(value == null) ...``` from the generated code). It's just a nit anyway :-)

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -46,7 +46,7 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
   // Some MojoModels are not serializable ( DeepLearning ), so we are reusing the mojoData to keep information about mojo model
   @transient var easyPredictModelWrapper: EasyPredictModelWrapper = _
 
-  case class BinomialPrediction(p0: Double, p1: Double)
+  case class BinomialPrediction(p0: Double, p1: Double, p0_calibrated: Double, p1_calibrated: Double)
```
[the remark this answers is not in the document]
jakubhava: I would like to keep the structure simple for user, at least for now.

Marking as resolved as discussed in person

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
jakubhava: Cool, didn't know that, thanks!

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -78,9 +79,23 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
   }
 
 
-  implicit def toBinomialPrediction(pred: AbstractPrediction) = BinomialPrediction(
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(0),
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(1))
+  implicit def toBinomialPrediction(pred: AbstractPrediction): BinomialPrediction = {
+    if (pred.asInstanceOf[BinomialModelPrediction].calibratedClassProbabilities != null) {
```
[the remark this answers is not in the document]
jakubhava: Good question, let's thing of that later! 👍

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
jakubhava: 👍

py/tests/tests_unit_conversions.py
```diff
@@ -229,8 +229,12 @@ def test_load_mojo_gbm(self):
         pred_model = model.transform(prostate_frame).repartition(1).collect()
 
         self.assertEquals(len(pred_mojo), len(pred_model))
+        num_cols = len(pred_mojo[0])
         for i in range(0, len(pred_mojo)):
-            self.assertEquals(pred_mojo[i], pred_model[i])
+            for k in range(0, num_cols - 2):
```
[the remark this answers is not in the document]
jakubhava: It was because the last column contained the nan - with null, we can return to original codebase

jakubhava:
@mn-mikke fixed the issues, please have a second look :)

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -78,9 +79,15 @@ class H2OMOJOModel(val mojoData: Array[Byte], override val uid: String)
   }
 
 
-  implicit def toBinomialPrediction(pred: AbstractPrediction) = BinomialPrediction(
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(0),
-    pred.asInstanceOf[BinomialModelPrediction].classProbabilities(1))
+  implicit def toBinomialPrediction(pred: AbstractPrediction): BinomialPrediction = {
+    val calibratedProps = pred.asInstanceOf[BinomialModelPrediction].calibratedClassProbabilities
+    BinomialPrediction(
+      pred.asInstanceOf[BinomialModelPrediction].classProbabilities(0),
```
mn-mikke: nit: we can down-cast only once

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
mn-mikke: Apologize for the confusion with ```NaN``` vs. ```null```. ```Double``` is a value type is in Scala.
```
scala> val d = null.asInstanceOf[Double]
d: Double = 0.0
```
So I'm wondering how we gonna specify ```p0_calibrated```, ```p1_calibrated``` when are not defined... ```Option```, ```java.lang.Double```, ```Double.NaN```?

PR #1175, opened 2019-04-26
[SW-1231] WIP: Training API more similar to Spark

jakubhava:
Very very WIP, for Scala, probably not complete even for Scala.

Because the only input will be array of feaures ( doubles) we can simplify the code and throw out all the categorical conversions

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
@@ -47,7 +47,7 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag, M <: SparkModel[M]
   }
 
   override def fit(dataset: Dataset[_]): M = {
-
+    import org.apache.spark.sql.functions.col
     // Update H2O params based on provided configuration
     updateH2OParams()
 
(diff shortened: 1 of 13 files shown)


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -85,12 +80,12 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag, M <: SparkModel[M]
...
     }
     water.DKV.put(trainFrame)
-    
+
     // Train
     val model: M with H2OModelParams = trainModel(getParams)
 
     // pass some parameters set on algo to model
-    model.setFeaturesCols($(featuresCols))
+    model.setFeaturesCols(Array($(featuresCol)))
```
jakubhava: For prediction time, still keep the api same, iterate in small steps :)

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -47,18 +47,17 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag, M <: SparkModel[M]
...
-      setFeaturesCols(dataset.columns.filter(n => n.compareToIgnoreCase(getLabelCol()) != 0))
+    if(dataset.columns.contains(getFeaturesCol())){
+      throw new IllegalArgumentException(s"Can not find features column '${getFeaturesCol()}' in the dataset")
     }
 
-    val cols = getFeaturesCols().map(col) ++ Array(col(getLabelCol()))
-    val input = hc.asH2OFrame(dataset.select(cols: _*).toDF())
+    // To find the model, we just need two columns, label & features
+    // We are sure that features column is array of doubles
+    val input = hc.asH2OFrame(dataset.select(getLabelCol(), getFeaturesCol()).toDF())
```
jakubhava: Actually, we will need to convert also fold column and weight column, if set as they need to be available on the H2O frame

jakubhava:
Should go to next major release as it will unblock us with pipeline improvements

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OModelParams.scala
```diff
@@ -16,73 +16,41 @@
...
+  setDefault(predictionCol -> "prediction")
   setDefault(convertUnknownCategoricalLevelsToNa -> false)
 
   //
   // Getters
   //
-  /** @group getParam */
-  @DeprecatedMethod("getLabelCol")
-  def getPredictionsCol(): String = getLabelCol()
+  def getPredictionCol(): String = $(predictionCol)
```
mn-mikke: remove

jakubhava:
Closing this one due to number of conflicts, it will be safer to start this change over

PR #1183, opened 2019-05-02
[SW-1248][FOLLOWUP] H2OMojoModel features will be determined by H2O internals

mn-mikke:
The PR depends on the release of https://github.com/h2oai/h2o-3/commit/e80b2421ca1ad4bc808bf87f9f1847dced7895fb.


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
```diff
@@ -124,11 +124,12 @@ class H2OMOJOPipelineModel(override val uid: String)
   }
 
   override def transform(dataset: Dataset[_]): DataFrame = {
-    val flatten = H2OSchemaUtils.flattenDataFrame(dataset.toDF())
-    val names = flatten.schema.fields.map(f => flatten(f.name))
+    val flattenedDF = H2OSchemaUtils.flattenDataFrame(dataset.toDF())
+    val relevantColumnNames = flattenedDF.columns.intersect(getInputNames())
```
jakubhava: @mn-mikke Please don't use input names here.

Correct approach is to define featureCols and define them during creating of H2OMOjoModel. Here, you should use getFeaturesCols.

GetInputNames will get removed in the future.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOPipelineModelParams.scala
```diff
@@ -27,24 +27,33 @@ trait H2OMOJOPipelineModelParams extends Params {
...
+  def getFeaturesCols(): Array[String] = $(featuresCols)
+
   //
   // Setters
   //
   /** @group setParam */
   def setNamedMojoOutputColumns(value: Boolean): this.type = set(namedMojoOutputColumns, value)
 
+  /** @group setParam */
+  def setFeaturesCols(cols: Array[String]): this.type = set(featuresCols, cols)
```
mn-mikke: I'm not sure whether we should specify a setter for the property.

py/py_sparkling/ml/models.py
```diff
@@ -66,9 +67,18 @@ def predict(self, dataframe):
         return self.transform(dataframe)
 
     def get_input_names(self):
+        warnings.warn("The method 'get_input_names' is deprecated. Use 'get_features_colssss' instead!")
         return list(self._java_obj.getInputNames())
 
+    def get_features_cols(self):
```
mn-mikke: Naming conventions of this class are inconsistent with other classes since snake case used here.

jakubhava: Can we deprecated the snake case and recommend the camel case?

Such as:
```
warnings.warn("The method 'get_input_names' is deprecated. Use 'getFeaturesCols' instead!")
```

We should use camel case here as well

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOPipelineModelParams.scala
jakubhava: On the model, I would keep for `featuresCols` just getter

PR #1185, opened 2019-05-02
[SW-1264] Use default params reader in pipelines

jakubhava:
@mn-mikke so it was possible :) Tested locally, but want to wait for tests. This is the last step before I go ahead and try to unify H2OMojo & H2OMojoPipelineModel


ml/src/main/scala/py_sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -16,27 +16,11 @@
...
-import water.support.ModelSerializationSupport
+import org.apache.spark.ml.util.DefaultParamsReadable
 
 /**
   * H2O AutoML Wrapper for PySparkling
   */
-class H2OAutoML(override val uid: String) extends org.apache.spark.ml.h2o.algos.H2OAutoML(uid) {
+class H2OAutoML(override val uid: String) extends org.apache.spark.ml.h2o.algos.H2OAutoML(uid)
 
-  override def trainModel(aml: AutoML): H2OMOJOModel = {
```
jakubhava: The implementation is the same as in the super class. We can remove it. The same holds for all others PySparkling wrappers

mn-mikke: That sounds great!

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -135,43 +134,9 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
...
-    val ctor = aClass.getConstructor(classOf[String])
-    ctor.newInstance(uid).asInstanceOf[CT]
-  }
 }
 
-
-object H2OAutoMLReader {
-  def create[A <: H2OAutoML : ClassTag](defaultFileName: String) = new H2OAutoMLReader[A](defaultFileName)
-}
+object H2OAutoML extends DefaultParamsReadable[py_sparkling.ml.algos.H2OAutoML]
```
mn-mikke: Can you help me understand why types from ```py_sparkling.ml.algos._``` are passed as generic parameters to ```DefaultParamsReadable```? I would expect types that implement ```DefaultParamsReadable```.

jakubhava: This is needed as we always need to store the classes using `py_sparkling.ml.algos` classes.

Without this, it works fine if you just train and run pipelines in PySpark or just in Scala.

But if you, for example, decide to train pipeline in Scala, export it and load it in PySpark then we hid the issue as `org.apache.spark` prefix is translated to `pyspark` on Python side. So to support all cases, we need to use `py_sparkling` classes as backend.

Hope it makes sense

PR #1189, opened 2019-05-06
[SW-1187][Pre-Step] Use H2OMojoModel to load mojo & mojo pipelines as well, slowly deprecating publicly visible H2OMOJOPipelineMode

jakubhava:


gradle.properties
```diff
@@ -21,7 +21,7 @@ mojoPipelineVersion=0.13.16
 
 # Set to True of mojo pipelines integration tests should run
 # as part of the Sparkling Water testing suite
-testMojoPipeline=false
+testMojoPipeline=true
```
jakubhava: revert before merging

PR #1192, opened 2019-05-08
[SW-1207] Implement H2OTargetEncoder for Scala API

mn-mikke:
This PR requires https://github.com/h2oai/h2o-3/pull/3282 to be released for a successful build.

The implementation of pysparkling wrappers and detailed tests will come in subsequent PRs.


ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  //
+  // Default values
+  //
+  setDefault(
+    foldCol -> null,
+    labelCol -> "label",
+    inputCols -> Array[String](),
+    holdoutStrategy -> H2OTargetEncoderHoldoutStrategy.None,
+    blending -> null,
```
mn-mikke: Blending and noise parameters are grouped together. if `blending` or `noise` is disabled, the particular feature is disabled. @jakubhava, I would like to know your opinion on this.

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,83 @@
...
+  with DefaultParamsWritable {
+
+  def this() = this(Identifiable.randomUID("H2OTargetEncoder"))
+
+  override def fit(dataset: Dataset[_]): H2OTargetEncoderModel = {
+    val h2oContext = H2OContext.getOrCreate(SparkSession.builder().getOrCreate())
+    val input = h2oContext.asH2OFrame(dataset.toDF())
+    changeRelevantColumnsToCategorical(input)
+    val targetEncoderModel = trainTargetEncodingModel(input)
+    val model = new H2OTargetEncoderModel(uid, targetEncoderModel).setParent(this)
```
jakubhava: Why do we need to call `setParent(this)`

jakubhava: Why do we need to call `setParent(this)` ?

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,83 @@
...
+    copyValues(model)
+  }
+
+  private def trainTargetEncodingModel(trainingFrame: Frame) = {
+    val targetEncoderParameters = new TargetEncoderModel.TargetEncoderParameters()
+    val blending = Option(getBlending())
+    targetEncoderParameters._withBlending = blending.isDefined
+    targetEncoderParameters._blendingParams = blending.map(_.toBlendingParams()).getOrElse(null)
```
jakubhava: can we use `orNull()` ?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OTargetEncoderModel.scala
```diff
@@ -0,0 +1,69 @@
...
+      mojoModel.transform(dataset)
+    }
+  }
+
+  def transformTrainingDataset(dataset: Dataset[_]): DataFrame = {
+    val h2oContext = H2OContext.getOrCreate(SparkSession.builder().getOrCreate())
+    val input = h2oContext.asH2OFrame(dataset.toDF())
+    changeRelevantColumnsToCategorical(input)
+    val noise = getNoise()
+    val holdoutStrategyId = getHoldoutStrategy().ordinal().asInstanceOf[Byte]
```
jakubhava: I'm sure you have thought of it - is `asInstanceOf[Byte]` necessary ?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OTargetEncoderModel.scala
```diff
@@ -0,0 +1,69 @@
...
+      targetEncoderModel.transform(input, holdoutStrategyId, 0L)
+    } else {
+      targetEncoderModel.transform(input, holdoutStrategyId, noise.amount, 0L)
+    }
+    h2oContext.asDataFrame(output)
+  }
+
+  private def inTrainingMode: Boolean = {
+    val stackTrace = Thread.currentThread().getStackTrace()
+    stackTrace.exists(e => e.getMethodName == "fit" && e.getClassName == "org.apache.spark.ml.Pipeline")
```
jakubhava: hehe, you found a way at the end! 👍

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
jakubhava: Good question! Does it semantically make sense to group this feature under one new, specific configuration object?

mn-mikke: if you started to use this estimator and you would see parameters like `smoothing`, `inflectionPoint` would you implicitly understand that they are parameters of blending? If blending is disabled they don't have any impact on anything. We can pose a similar question about the `seed` parameter for `noise`.

An alternative could be to name the parameters with some prefix that groups them, but I'm not sure what is better.

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OTargetEncoderModel.scala
mn-mikke: `ordinal()` returns `Integer`, and the type of the second parameter of the `transform` function is `Byte`. The conversion `Integer` -> `Byte` is not lossless, so I'm casting explicitly.

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
mn-mikke: Yeah, that will be better. Thanks!

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
mn-mikke: Just being consistent with other Spark estimators.

mn-mikke: Just being consistent with other Spark estimators.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
jakubhava: On the other hand, when I thing about it the users setting these properties probably have to know they require bleeding to be enabled. Maybe we can keep it this is simple.

At the end, it is easier for the user to set the options directly compared to creating some additional configuration object.

I would keep it like it is now

jakubhava:
One more note: @mn-mikke could you please also add tutorial into our documentation how to use target encoding in Sparkling Water? Different PR is fine for this

mn-mikke:
Yes, will do once this gets merged :-) Thanks @jakubhava!

ml/src/test/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoderTest.scala
```diff
@@ -0,0 +1,88 @@
...
+    val model = pipeline.fit(trainingDataset)
+    val path = "build/ml/targetEncoder_save_load"
+    model.write.overwrite().save(path)
+    val loadedModel = PipelineModel.load(path)
+    val transformedTestingDataset = model.transform(testingDataset)
+
+    assertDataFramesAreIdentical(expectedTestingDataset, transformedTestingDataset)
+  }
+
+  test("The target encoder doesn't apply noise on the testing dataset") {
```
deil87: @mn-mikke Having a hard time to understand this test. Could you please help me with this. So... we call `fit` on our pipeline...it means that in H2OTargetEncoderModel we are in `inTrainingMode` mode, right? And we have following logic `val noise = Option(getNoise()).getOrElse(H2OTargetEncoderNoiseSettings(amount = 0.0))` and it feels like we should apply noise as we set it to 0.5.  And it is not `H2OTargetEncoderMojoModel`. What am I missing?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OTargetEncoderModel.scala
```diff
@@ -0,0 +1,69 @@
...
+      transformTrainingDataset(dataset)
+    } else {
+      mojoModel.transform(dataset)
+    }
+  }
+
+  def transformTrainingDataset(dataset: Dataset[_]): DataFrame = {
+    val h2oContext = H2OContext.getOrCreate(SparkSession.builder().getOrCreate())
+    val input = h2oContext.asH2OFrame(dataset.toDF())
+    changeRelevantColumnsToCategorical(input)
```
deil87: @mn-mikke minor.. maybe `convert` suffix is better?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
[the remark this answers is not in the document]
mn-mikke: These parameters are serialized by reflection and deserialized by Pyspark. I'm not sure whether there is a support for options already. Maybe @jakubhava could shed some light on that.

Quite like skipping null defaults, but if we do that, we should do the same for all SW algorithms. We need to eventually double check the consequences.

Exposed default values in H2O-3 would be nice, it least it will be always consistent if a H2O-3 default value changes in future.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+  def getOutputCols(): Array[String] = getInputCols().map(_ + "_te")
+
+  def getHoldoutStrategy(): H2OTargetEncoderHoldoutStrategy = $(holdoutStrategy)
+
+  def getBlending(): H2OTargetEncoderBlendingSettings = $(blending)
+
+  def getNoise(): H2OTargetEncoderNoiseSettings = $(noise)
+
+  //
+  // Others
```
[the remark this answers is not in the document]
mn-mikke: Yep, will change as part of the comment [here](https://github.com/h2oai/sparkling-water/pull/1192/files#r297711737). Thanks!

ml/src/test/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoderTest.scala
```diff
@@ -0,0 +1,88 @@
...
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+import water.api.TestUtils
+
+@RunWith(classOf[JUnitRunner])
+class H2OTargetEncoderTest extends FunSuite with SharedH2OTestContext {
+
+  override def createSparkContext = new SparkContext("local[*]", "H2OTargetEncoderTest", conf = defaultSparkConf)
+
+  private def loadDataFrameFromCsv(path: String): DataFrame = {
```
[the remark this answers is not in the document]
mn-mikke: Yep, something generic has already been created  [here](https://github.com/h2oai/sparkling-water/pull/1279/files#diff-cc0eea4e341c4b7ae26dc9c5028164a0R88) recently.

ml/src/test/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoderTest.scala
mn-mikke: `inTrainingMode` is valid only for [fitting](https://github.com/h2oai/sparkling-water/pull/1192/files#diff-7908001f8c9b38a0a2354f5996b51c6aR83). At that time ,noise is applied according to the parameters. When testing dataset is [transformed](https://github.com/h2oai/sparkling-water/pull/1192/files#diff-7908001f8c9b38a0a2354f5996b51c6aR84) we are not `inTrainingMode` anymore and noise won't be applied.

 Does it make sense?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.spark.h2o.Frame
+
+trait H2OTargetEncoderParams extends PipelineStage with Params {
+
+  //
+  // Override pipeline stage members
+  //
+  override def transformSchema(schema: StructType): StructType = {
+    val flatSchema = H2OSchemaUtils.flattenSchema(schema)
+    require(getLabelCol() != null, "Label column can't be null!")
```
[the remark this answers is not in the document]
deil87: @mn-mikke the point is that validation here has very little to do with transformation of the schema. It is just validation of the parameters and ideally should be done during `H2OTargetEncoderParams` initialisation. I understand that flattening of the schema helps with validation but still validation should be done long before transformation. Maybe we can reuse/pass around flattened schema. I would understand if we did validation of the Schema here.... then just separate private method for this would be enough. But in this case it is validation of parameters and it should be done somewhere else ( constructor). Just wanted to bring your attention .... feel free to ignore or consider just private method with name that highlights that it is params validation(and not a validation of newly created schema).

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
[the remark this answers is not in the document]
jakubhava: If bleeding have default params then H2O should define them and we should use them. Otherwise use `null` as is in the rest of the code

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+    inputCols -> Array[String](),
+    holdoutStrategy -> H2OTargetEncoderHoldoutStrategy.None,
+    blending -> null,
+    noise -> H2OTargetEncoderNoiseSettings(0.01, -1)
+  )
+
+  //
+  // Getters
+  //
+  def getFoldCol(): String = $(foldCol)
```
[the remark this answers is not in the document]
jakubhava: Consistency with the rest of the product

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,83 @@
...
+
+class H2OTargetEncoder(override val uid: String)
+  extends Estimator[H2OTargetEncoderModel]
+  with H2OTargetEncoderParams
+  with DefaultParamsWritable {
+
+  def this() = this(Identifiable.randomUID("H2OTargetEncoder"))
+
+  override def fit(dataset: Dataset[_]): H2OTargetEncoderModel = {
+    val h2oContext = H2OContext.getOrCreate(SparkSession.builder().getOrCreate())
```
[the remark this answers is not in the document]
jakubhava: Good idea, but not related to the core change. Let's focus on TE part please :)

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,83 @@
...
+    val input = h2oContext.asH2OFrame(dataset.toDF())
+    changeRelevantColumnsToCategorical(input)
```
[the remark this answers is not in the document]
jakubhava: Sparkling Water is trying( and still some way to reach the goal) to hide tasks which can be done automatically. I believe we should do these automatically, not ask user for explicit data prep

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+package org.apache.spark.ml.h2o.param
+
+import org.apache.spark.h2o.utils.H2OSchemaUtils
+import org.apache.spark.ml.h2o.features._
+import org.apache.spark.ml.PipelineStage
+import org.apache.spark.ml.param.{Param, Params, StringArrayParam}
+import org.apache.spark.sql.types.{DoubleType, StructField, StructType}
+import org.apache.spark.h2o.Frame
+
+trait H2OTargetEncoderParams extends PipelineStage with Params {
```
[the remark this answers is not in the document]
jakubhava: We don't want to introduce new `PipelineStage`. That might break reading and loading of spark pipelines. 

Users should work with native Spark types such as `Pipeline`, `PipelineModel` and `PipelineStage`

ml/src/test/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoderTest.scala
deil87: @mn-mikke the question is how do we switch from fitting/`inTrainingMode`  to not `inTrainingMode`?  Because it feels like StackTraceElement which satisfies `inTrainingMode` conditions could be still there.

ml/src/main/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoder.scala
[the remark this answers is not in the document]
deil87: @kuba I agree that we should try to do automatically as much as possible just what if user specified wrong column for TE and it was numerical one... algorithm will silently convert it but performance would be awful (not saying that probably it is not what user was planning to do). If user had not to specify TE columns himself then I would totally agree with full automation. When human is in the loop I would make any transformations more explicit (log warning or throw). Just my opinion.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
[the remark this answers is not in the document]
deil87: @jakubhava It would not break anything.... you can inherit from `PipelineStage` and place some shared logic there (e.g. `transformSchema`).  Liskov substitution principle.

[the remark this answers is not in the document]
jakubhava: yes, this will work, thanks. Got confused as previously we had H2OPipeline in the code which was really breaking serialization and deserialization + user was forced to use different kind of pipeline.

As long as the introduced types are not exposed, we're good

ml/src/test/scala/org/apache/spark/ml/h2o/features/H2OTargetEncoderTest.scala
mn-mikke: > Because it feels like StackTraceElement which satisfies inTrainingMode conditions could be still there.

@deil87 Please can you give me an example?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+    val ioIntersection = getInputCols().intersect(getOutputCols())
+    require(ioIntersection.isEmpty,
+      s"""The columns [${ioIntersection.map(i => s"'$i'").mkString(", ")}] are specified
+         |as input columns and also as output columns. There can't be an overlap.""".stripMargin)
+    StructType(flatSchema.fields ++ getOutputCols().map(StructField(_, DoubleType, nullable = true)))
+  }
+
+  //
+  // List of Parameters
```
mn-mikke: @deil87 Please can you also go through the parameter descriptions and check whether they make sense?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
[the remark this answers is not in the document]
mn-mikke: @deil87 I got your point but the parameters may change after the instance is created
```
val targetEncoder = new H2OTargetEncoder()
  .setInputCols(Array("RACE", "DPROS", "DCAPS"))
  .setLabelCol("CAPSULE")
```

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+  protected final val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
+  protected final val labelCol = new Param[String](this, "labelCol", "Label column name")
+  protected final val inputCols = new StringArrayParam(this, "inputCols", "Names of columns that will be transformed")
+  protected final val holdoutStrategy = new H2OTargetEncoderHoldoutStrategyParam(this,
+    "holdoutStrategy",
+    """A strategy deciding what records will be excluded when calculating the target average on the training dataset.
+      |Options:
+      | None        - All rows are considered for the calculation
+      | LeaveOneOut - All rows except the row the colculation is made for
+      | KMeans      - Only out-of-fold data is considered (The option requires foldCol to be set.)
```
deil87: @mn-mikke KMeans -> KFold

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+  //
+  protected final val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
+  protected final val labelCol = new Param[String](this, "labelCol", "Label column name")
+  protected final val inputCols = new StringArrayParam(this, "inputCols", "Names of columns that will be transformed")
+  protected final val holdoutStrategy = new H2OTargetEncoderHoldoutStrategyParam(this,
+    "holdoutStrategy",
+    """A strategy deciding what records will be excluded when calculating the target average on the training dataset.
+      |Options:
+      | None        - All rows are considered for the calculation
+      | LeaveOneOut - All rows except the row the colculation is made for
```
deil87: @mn-mikke typo -> cAlculation

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+      | KMeans      - Only out-of-fold data is considered (The option requires foldCol to be set.)
+    """.stripMargin)
+  protected final val blending = new CaseClassParam[H2OTargetEncoderBlendingSettings](this,
+    "blending",
+    """If set, the target average becomes a weighted average of the group target value and the global target value of a given row.
```
deil87: @mn-mikke  maybe like this -> "....average of the posterior average for a given categorical level and the prior average of the target."
"global target value of a given row" - does not make sense as global value is not being computed for a row.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+      |The weight is determined by the size of the given group that the row belongs to.
+      |Attributes:
+      | InflectionPoint - The bigger number it's, the bigger groups will consider the global target value as a component in the weighted average.
```
deil87: @mn-mikke better rephrase InflectoinPoint as well. We should reason from the size of the given group. It is relative. We can't just say the bigger the bigger. Maybe can say kile this about the diff. Something like this: the bigger difference between size of the given group and IP the more weight will be put for posterior average of the target.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+      | Smoothing       - Controls the rate of transition between group target values and global target values.""".stripMargin)
```
deil87: @mn-mikke maybe better to use single nouns? "...target value and global target value". Not sure.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,125 @@
...
+  protected final val noise = new CaseClassParam[H2OTargetEncoderNoiseSettings](this,
+    "noise",
+    """The settings affecting how much of noise will be added to the target average.
```
deil87: @mn-mikke actually only one setting will affect how much noise will be added. Seed does not affect this. Maybe "The setting that affects...."

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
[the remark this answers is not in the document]
deil87: @mn-mikke Got your point too :) And the order of these methods is not predefined... we probably can't add validation into these methods - too many combinations of possible settings(orders).

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OTargetEncoderModel.scala
deil87: @mn-mikke No.... no way for flag here as you are using spark's `Pipeline`. I just remember we were discussing in Confluence that SW has its own wrapper (`H2OPipeline`?) for pipelines ( which is probably what @jakubhava is referring to [here](https://github.com/h2oai/sparkling-water/pull/1192/files#r297783946))

jakubhava: We do not have H2OPipeline anymore and we definitely need to use the provided Pipeline from Spark

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
mn-mikke: Oh, good catch! Thanks a lot!

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OTargetEncoderParams.scala
mn-mikke: @deil87 Would it make sense if we rephrased the sentence  like this: "the bigger groups" -> "the groups relatively bigger to the overall data set size will ..." ?

deil87: @mn-mikke It should be relatively bigger than inflection point. "The bigger group is with respect to an inflection point the more weight we will put onto its target average and less weight onto a prior average. " . Take a look at Py/R documentation for extra inspiration.

jakubhava:
@mn-mikke I changed jenkins files a bit, please rebase on top of master so we can run tests on this PR. Thanks!

mn-mikke:
@jakubhava Thanks for letting me know! I will do the rebase together with the next set of changes.

core/src/main/scala/org/apache/spark/sql/DatasetExtensions.scala
```diff
@@ -0,0 +1,24 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package org.apache.spark.sql
+
+object DatasetExtensions{
```
jakubhava: NIT: missing space

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,92 @@
...
+  //
+  // Parameter Setters
+  //
+  def setFoldCol(value: String): this.type = set(foldCol, value)
+
+  def setLabelCol(value: String): this.type = set(labelCol, value)
+
+  def setInputCols(values: Array[String]): this.type = set(inputCols, values)
+
+  def setHoldoutStrategy(value: H2OTargetEncoderHoldoutStrategy): this.type = set(holdoutStrategy, value)
```
jakubhava: Should we use String value here as we discussed right from beginning? See the PR here https://github.com/h2oai/sparkling-water/pull/1359

I in favor of the the string variant

ml/src/main/java/ai/h2o/sparkling/ml/features/H2OTargetEncoderHoldoutStrategy.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.h2o.sparkling.ml.features;
+
+public enum H2OTargetEncoderHoldoutStrategy {
```
jakubhava: should be just internal, does not need to be exposed to public as we will use String representations on the API

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -0,0 +1,92 @@
...
+    targetEncoderParameters._teFoldColumnName = getFoldCol()
+    targetEncoderParameters._columnNamesToEncode = getInputCols()
+    targetEncoderParameters.setTrain(trainingFrame._key)
+
+    val builder = new TargetEncoderBuilder(targetEncoderParameters)
+    builder.trainModel().get() // Calling get() to wait until the model training is finished.
+    builder.getTargetEncoderModel()
+  } catch {
+    case e: IllegalStateException if e.getMessage.contains("We do not support multi-class target case") =>
+      throw new RuntimeException("The label column can not contain more than two unique values.")
```
jakubhava: Thanks for the rewording on our side!

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBase.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import ai.h2o.sparkling.ml.params.H2OTargetEncoderParams
+import org.apache.spark.h2o.Frame
+import org.apache.spark.ml.PipelineStage
+import org.apache.spark.sql.types.{DoubleType, StructField, StructType}
+
+trait H2OTargetEncoderBase extends PipelineStage with H2OTargetEncoderParams {
+  override def transformSchema(schema: StructType): StructType = {
+    validateSchema(schema)
+    StructType(schema.fields ++ getOutputCols().map(StructField(_, DoubleType, nullable = true)))
```
jakubhava: Are we sure output cols will have type of `DoubleType` ?

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBase.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  private def validateSchema(flatSchema: StructType): Unit = {
+    require(getLabelCol() != null, "Label column can't be null!")
+    require(getInputCols() != null && getInputCols().nonEmpty, "The list of input columns can't be null or empty!")
+    val fields = flatSchema.fields
+    val fieldNames = fields.map(_.name)
+    require(fieldNames.contains(getLabelCol()),
+      s"The specified label column '${getLabelCol()}' was not found in the input dataset!")
+    for(inputCol <- getInputCols()) {
```
jakubhava: NIT: What do you think about `getInputCols.foreach` ?

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderModel.scala
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+import org.apache.spark.sql.functions._
+import water.support.ModelSerializationSupport
+
+class H2OTargetEncoderModel(
+    override val uid: String,
+    targetEncoderModel: TargetEncoderModel)
+  extends Model[H2OTargetEncoderModel] with H2OTargetEncoderBase with MLWritable {
+
+  lazy val mojoModel: H2OTargetEncoderMojoModel = {
```
jakubhava: Just note: Do we have a test for the serialization?

Ie: We train Te Model, serialize it and load it again, we should be able to score again,should get the same results.

For this we need to store the mojoModel as part of the serialized structure

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMojoModel.scala
```diff
@@ -0,0 +1,84 @@
...
+import water.support.ModelSerializationSupport
+
+class H2OTargetEncoderMojoModel(override val uid: String) extends Model[H2OTargetEncoderMojoModel]
+  with H2OTargetEncoderBase with H2OMOJOWritable with H2OMOJOFlattenedInput {
+
+  override protected def inputColumnNames: Array[String] = getInputCols()
+
+  override protected def outputColumnName: String = this.getClass.getSimpleName + "_output"
+
+  def this() = this(Identifiable.randomUID("H2OTargetEncoderMojoModel"))
```
jakubhava: We could actually use `this.getClass.getSimpleName` instead of manually specifying the base for the random UID

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMojoModel.scala
```diff
@@ -0,0 +1,84 @@
...
+  * The class holds all necessary dependencies of udf that needs to be serialized.
+  */
+case class H2OTargetEncoderMojoUdfWrapper(mojoData: Array[Byte], outputCols: Array[String]) {
+  @transient private lazy val easyPredictModelWrapper: EasyPredictModelWrapper = {
+    val model = ModelSerializationSupport
+      .getMojoModel(mojoData)
+      .asInstanceOf[TargetEncoderMojoModel]
+    val config = new EasyPredictModelWrapper.Config()
+    config.setModel(model)
+    config.setConvertUnknownCategoricalLevelsToNa(true)
```
jakubhava: It this hardcoded in H2O as well? Should this be configurable via our H2OMojoSettings?

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.scala
```diff
@@ -0,0 +1,105 @@
...
+  protected final val noiseSeed = new LongParam(this, "noiseSeed", "A seed of the generator producing the random noise")
+
+  //
+  // Default values
+  //
+  setDefault(
+    foldCol -> null,
+    labelCol -> "label",
+    inputCols -> Array[String](),
+    holdoutStrategy -> H2OTargetEncoderHoldoutStrategy.None,
```
jakubhava: We can switch to string representation as discussed above

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderTestSuite.scala
```diff
@@ -0,0 +1,309 @@
+/*
```
jakubhava: Thanks for the tests!

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderModel.scala
jakubhava: Just saw tests bellow, I think we're good 👍

ml/src/main/java/ai/h2o/sparkling/ml/features/H2OTargetEncoderHoldoutStrategy.java
mn-mikke: I think this enum is not needed anymore. Will remove it.

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBase.scala
mn-mikke: TE returns a prior or posterior mean or a mixture of both.

https://github.com/h2oai/h2o-3/pull/3282/files#diff-2d747267b5dfe86c069723201c2cdd97R46

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMojoModel.scala
mn-mikke: If this option was false, `EasyPredictModelWrapper` would throw an exception and [output would be null](https://github.com/h2oai/sparkling-water/pull/1192/files#diff-e51c0ac92a4dbe5601ffae48ae3dfb4fR79). In a regular case, TE uses a global average as output value for unknown values.

I can expose this option, but is there use case for that?

jakubhava: No need to expose, thanks for explanation!

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBase.scala
jakubhava: Thanks!

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMojoModel.scala
mn-mikke: Let's double-check it with @deil87 and @michalkurka.

Can you think of an use case where would make more sense if an algorithm in a pipeline after TE would obtain NA (null -> NA) instead of the global average for unknown categoricals?

Btw: `TargetEncoderModel` (Java API) throws an exception in the same scenario:
```
water.exceptions.H2OConcurrentModificationException: Rollups not possible, because Vec was deleted: $04ff5e000000ffffffffb00376b65fbfebb9ad4aabb7d387d3$D

	at water.fvec.RollupStats.get(RollupStats.java:323)
	at water.fvec.RollupStats.get(RollupStats.java:353)
	at water.fvec.Vec.rollupStats(Vec.java:901)
	at water.fvec.Vec.naCnt(Vec.java:853)
	at ai.h2o.automl.targetencoding.TargetEncoder.imputeWithMean(TargetEncoder.java:294)
	at ai.h2o.automl.targetencoding.TargetEncoder.applyTargetEncoding(TargetEncoder.java:645)
	at ai.h2o.automl.targetencoding.TargetEncoderModel.transform(TargetEncoderModel.java:96)
	at ai.h2o.sparkling.ml.models.H2OTargetEncoderModel.transformTrainingDataset(H2OTargetEncoderModel.scala:61)
	at ai.h2o.sparkling.ml.features.H2OTargetEncoderTestSuite$$anonfun$6.apply$mcV$sp(H2OTargetEncoderTestSuite.scala:137)
	at ai.h2o.sparkling.ml.features.H2OTargetEncoderTestSuite$$anonfun$6.apply(H2OTargetEncoderTestSuite.scala:125)
	at ai.h2o.sparkling.ml.features.H2OTargetEncoderTestSuite$$anonfun$6.apply(H2OTargetEncoderTestSuite.scala:125)
	at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
```
  
But this is not a Sparkling Water use case.

deil87: @mn-mikke This is hardcoded in TargetEncoder because along the way I have found a [bug](https://0xdata.atlassian.net/browse/PUBDEV-5924) in merge functionality.... and I had to always substitute nulls with new category. We probably can remove this parameter at all later.

mn-mikke: https://0xdata.atlassian.net/browse/PUBDEV-6704

PR #1197, opened 2019-05-14
[SW-1279] Fix inconsistencies between H2OAutoML, H2OGridSearch & H2OALgorithm

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -89,8 +89,6 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag]
     val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID(binaryModel._parms.algoName()))
 
     // pass some parameters set on algo to model
-    model.setFeaturesCols(binaryModel.modelDescriptor().features())
```
jakubhava: features are already set inside createFromMojo and that is the only path how to create this spark model based on mojo

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -89,8 +89,6 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag]
...
-    model.setLabelCol(getLabelCol())
```
jakubhava: As we discussed, we won't put label col into the final model. Also we don't do it for mojo pipeline & automl & gridsearch

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -107,9 +107,12 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
         " your 'excludeAlgo', 'maxModels' or 'maxRuntimeSecs' properties.") with NoStackTrace
     }
 
-    val mojoData = ModelSerializationSupport.getMojoData(aml.leader())
+    val binaryModel = aml.leader()
+    val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
     val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID(aml.leader()._parms.algoName()))
-    model.setConvertUnknownCategoricalLevelsToNa(true)
```
jakubhava: I set default to true on automl and grieadseach so we have one place where this value is defined

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -126,9 +126,12 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
...
     }
 
-    val mojoData = ModelSerializationSupport.getMojoData(selectModelFromGrid(grid))
+    val binaryModel = selectModelFromGrid(grid)
+    val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
     val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID("gridSearch_mojoModel"))
-    model.setConvertUnknownCategoricalLevelsToNa(true)
+
+    // pass some parameters set on algo to model
+    model.setConvertUnknownCategoricalLevelsToNa(getConvertUnknownCategoricalLevelsToNa())
```
jakubhava: convertUnknownToCategoricals was missing on grid search

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOModelParams.scala
```diff
@@ -32,56 +32,33 @@ trait H2OMOJOModelParams extends DeprecatableParams {
   //
   // Param definitions
   //
-  private final val labelCol: Param[String] = new Param[String](this, "labelCol", "Label column name")
-  private final val featuresCols: StringArrayParam = new StringArrayParam(this, "featuresCols", "Name of feature columns")
+  private val labelCol: Param[String] = new Param[String](this, "labelCol", "Label column name")
```
jakubhava: keep for backward compatibility of already existing mojos

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOModelParams.scala
```diff
@@ -32,56 +32,33 @@ trait H2OMOJOModelParams extends DeprecatableParams {
...
   setDefault(featuresCols -> Array.empty[String])
   setDefault(outputCol -> "prediction_output")
   setDefault(convertUnknownCategoricalLevelsToNa -> false)
 
   //
   // Getters
   //
-  /** @group getParam */
-  @DeprecatedMethod("getLabelCol")
-  def getPredictionsCol(): String = getLabelCol()
```
jakubhava: remove getter for label col

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOModelParams.scala
```diff
@@ -32,56 +32,33 @@ trait H2OMOJOModelParams extends DeprecatableParams {
...
 
 
   //
   // Setters
   //
-  /** @group setParam */
-  def setFeaturesCols(cols: Array[String]): this.type = set(featuresCols, cols)
-
-  /** @group setParam */
-  @DeprecatedMethod("setLabelCol")
```
jakubhava: remove setters for features col & labelCol

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
mn-mikke: On second thoughts, what if the testing (PROD) dataframe has columns named differently?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
mn-mikke: Ditto

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -193,7 +196,7 @@ trait H2OAutoMLParams extends DeprecatableParams {
     stoppingTolerance -> 0.001,
     stoppingMetric -> ScoreKeeper.StoppingMetric.AUTO,
     nfolds -> 5,
-    convertUnknownCategoricalLevelsToNa -> false,
+    convertUnknownCategoricalLevelsToNa -> true,
```
mn-mikke: Is this the default behavior in H2O-3?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -414,6 +417,8 @@ trait H2OGridSearchParams extends DeprecatableParams {
   private val selectBestModelDecreasing = new BooleanParam(this, "selectBestModelDecreasing",
     "True if sort in decreasing order accordingto selected metrics")
   private val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
+  private val convertUnknownCategoricalLevelsToNa = new BooleanParam(this, "convertUnknownCategoricalLevelsToNa", "Convert unknown" +
```
mn-mikke: nit: formatting.. WDYT about:
```
private val convertUnknownCategoricalLevelsToNa = new BooleanParam(
   this,
   "convertUnknownCategoricalLevelsToNa",
   "Convert unknown categorical levels to NA during predictions")

File: ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala [hunk 10]
```diff
@@ -414,6 +417,8 @@ trait H2OGridSearchParams extends DeprecatableParams {
...
+    " categorical levels to NA during predictions")
```
mn-mikke: nit: What about a different comment?
```If set to 'true', the model converts unknown categorical levels to NA during making predictions.```

Is this also related to the training phase?

File: ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOModelParams.scala (hunk 6 above)
Comment (mn-mikke) [reply ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OMOJOModelParams.scala]: Shouldn't we do such changes only in major releases?

File: py/py_sparkling/ml/algos.py [hunk 11]
```diff
@@ -374,7 +374,7 @@ class H2OGridSearch(H2OGridSearchParams, JavaEstimator, JavaH2OMLReadable, JavaM
     def __init__(self, algo=None, ratio=1.0, hyperParameters={}, labelCol="label", allStringColumnsToCategorical=True,
                  columnsToCategorical=[], strategy="Cartesian", maxRuntimeSecs=0.0, maxModels=0, seed=-1,
                  stoppingRounds=0, stoppingTolerance=0.001, stoppingMetric="AUTO", nfolds=0, selectBestModelBy=None,
-                 selectBestModelDecreasing=True, foldCol=None, **deprecatedArgs):
+                 selectBestModelDecreasing=True, foldCol=None, convertUnknownCategoricalLevelsToNa=False, **deprecatedArgs):
```
mn-mikke: ```convertUnknownCategoricalLevelsToNa=True``` as it's proposed in the Scala part?

py/py_sparkling/ml/algos.py
```diff
@@ -384,15 +384,16 @@ def __init__(self, algo=None, ratio=1.0, hyperParameters={}, labelCol="label", a
...
+                         convertUnknownCategoricalLevelsToNa=False)
         kwargs = get_input_kwargs(self)
         self.setParams(**kwargs)
 
     @keyword_only
     def setParams(self, algo=None, ratio=1.0, hyperParameters={}, labelCol="label", allStringColumnsToCategorical=True,
                   columnsToCategorical=[], strategy="Cartesian", maxRuntimeSecs=0.0, maxModels=0, seed=-1,
                   stoppingRounds=0, stoppingTolerance=0.001, stoppingMetric="AUTO", nfolds=0, selectBestModelBy=None,
-                  selectBestModelDecreasing=True, foldCol=None, **deprecatedArgs):
+                  selectBestModelDecreasing=True, foldCol=None, convertUnknownCategoricalLevelsToNa=False, **deprecatedArgs):
```
mn-mikke: ditto

py/py_sparkling/ml/algos.py
```diff
@@ -384,15 +384,16 @@ def __init__(self, algo=None, ratio=1.0, hyperParameters={}, labelCol="label", a
                          maxRuntimeSecs=0.0, maxModels=0, seed=-1,
                          stoppingRounds=0, stoppingTolerance=0.001,
                          stoppingMetric=self._hc._jvm.hex.ScoreKeeper.StoppingMetric.valueOf("AUTO"), nfolds=0,
-                         selectBestModelBy=None, selectBestModelDecreasing=True, foldCol=None)
+                         selectBestModelBy=None, selectBestModelDecreasing=True, foldCol=None,
+                         convertUnknownCategoricalLevelsToNa=False)
```
mn-mikke: ditto

py/pysparkling/ml/params.py
```diff
@@ -1285,6 +1285,7 @@ class H2OGridSearchParams(Params):
     foldCol = Param(Params._dummy(), "foldCol", "Fold column name")
     selectBestModelBy = Param(Params._dummy(), "selectBestModelBy", "selectBestModelBy")
     selectBestModelDecreasing = Param(Params._dummy(), "selectBestModelDecreasing", "selectBestModelDecreasing")
+    convertUnknownCategoricalLevelsToNa = Param(Params._dummy(), "convertUnknownCategoricalLevelsToNa", "Convert unknown categorical levels to NA during predictions")
```
mn-mikke: nit: A better comment

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -107,9 +107,12 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
...
     }
 
-    val mojoData = ModelSerializationSupport.getMojoData(aml.leader())
+    val binaryModel = aml.leader()
+    val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
     val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID(aml.leader()._parms.algoName()))
-    model.setConvertUnknownCategoricalLevelsToNa(true)
+
+    // pass some parameters set on algo to model
+    model.setConvertUnknownCategoricalLevelsToNa(getConvertUnknownCategoricalLevelsToNa())
```
mn-mikke: Would it be possible to use ```copyValues``` for copying parameter values to the trained models or are there any limitations?

jakubhava: Good idea, let me try it

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: Default is false in H2O. For now, I will keep it consistent with current state of Sparkling Water, later we can iterate

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
jakubhava: I would not support that use-case. We would not now how to internally match the values. ( We use column name to fill the temporary data structure which is then passed to H2O).

We could use the order in which they came, but that is not really strong solution

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: It required that the default values are the same in the original and target param instance which right now is not true. I would iterate towards it, but right now we can't afford it

PR #1199, opened 2019-05-15
[SW-1281] Fix bad representation of predictionCol on H2OMOJOModel

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -77,7 +77,7 @@ class H2OMOJOModel(override val uid: String)
       case _ => throw new RuntimeException("Unknown model category")
     }
 
-    Seq(StructField(getOutputCol(), StructType(fields), nullable = false))
+    Seq(StructField(getPredictionCol, StructType(fields), nullable = false))
```
mn-mikke: nit: ```getPredictionCol()```

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -163,7 +163,7 @@ class H2OMOJOModel(override val uid: String)
     val flattenedDF = H2OSchemaUtils.flattenDataFrame(dataset.toDF())
     val relevantColumnNames = flattenedDF.columns.intersect(getFeaturesCols())
     val args = relevantColumnNames.map(flattenedDF(_))
-    flattenedDF.select(col("*"), getModelUdf()(struct(args: _*)).as(getOutputCol()))
+    flattenedDF.select(col("*"), getModelUdf()(struct(args: _*)).as(getPredictionCol))
```
mn-mikke: nit: ditto

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
```diff
@@ -18,13 +18,25 @@
...
   override def load(path: String): T  = {
-    val model = super.load(path)
+    val metadata = DefaultParamsReader.loadMetadata(path, sc)
+    val cls = Utils.classForName(metadata.className)
+    val instance =
+      cls.getConstructor(classOf[String]).newInstance(metadata.uid).asInstanceOf[Params]
+
+    val parsedParams = metadata.params.asInstanceOf[JsonAST.JObject].obj.map(_._1)
+    val allowedParams = instance.params.map(_.name)
+    val filteredParams = parsedParams.filter(!allowedParams.contains(_))
```
mn-mikke: WDYT about using the```intersect``` method?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
```diff
@@ -18,13 +18,25 @@
...
+    metadata.getAndSetParams(instance, Option(filteredParams))
```
mn-mikke: nit: Can ```filteredParams``` be ```null```? If not, I would prefer to use ```Some(filteredParams)```.

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
jakubhava: 👍

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
jakubhava: yup - I think it can't be null

jakubhava: setting to Some

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
jakubhava: Thanks -> as per our discussion in the office, intersect is not a good fit, but `diff` is

PR #1204, opened 2019-05-16
[SW-1248][FOLLOWUP] Introduce feature columns on GridSearch and AutoML

mn-mikke:
The PR contains:
* Introduction of feature columns on GridSearch and AutoML
* Separation of common parameters to a dedicated trait 
* Introduction of a weight column on GridSearch

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
@@ -47,7 +47,7 @@ import scala.collection.mutable
   * H2O Grid Search
   */
 class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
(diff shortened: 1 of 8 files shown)


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -87,6 +87,7 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
     algoParams._nfolds = getNfolds()
     algoParams._fold_column = getFoldCol()
     algoParams._response_column = getLabelCol()
+    algoParams._weights_column = getWeightCol()
```
mn-mikke: @jakubhava Please can you confirm that the weight column is also relevant for GridSearch?

jakubhava: Yup, this is fine

PR #1208, opened 2019-05-17
[SW-1286] Remove Param propagation of MOJOModels from Python to Java

mn-mikke:


mn-mikke:
@jakubhava IMHO, this the cleanest option how to avoid the hack introduced for the rel-2.3. I was also playing with the solution via standard params, but it would introduce much more lines of code and wouldn't give any extra value. Moreover, our overall target is to make MOJO models read-only without any setters.

jakubhava:
@mn-mikke good idea! There is however one exception -> we still need to expose a few setters on the mojo itself such as whether to convert unknown categoricals to NA etc.

But this solution could be used for that, couldn't it? We could override the method and set just specific values. At this moment it is:

`convert unknown categoricals` for mojo model and
`set mojo named output cols` for mojo pipeline model.

If people load mojo from h2o-3 or DAI, they might want to change this behaviour afterwards.

mn-mikke:
IMHO, this solution should also work with setters:
```
>>> gbm = H2OGBM(ntrees=2, seed=42, distribution="bernoulli", labelCol="capsule")
>>> model = gbm.fit(prostate_frame)
>>> model.getConvertUnknownCategoricalLevelsToNa()
False
>>> model.setConvertUnknownCategoricalLevelsToNa(True)
GBM_af4364933a79
>>> model.getConvertUnknownCategoricalLevelsToNa()
True
>>> pred_model = model.predict(prostate_frame).repartition(1).collect()
The method 'predict' is deprecated. Use 'transform' instead!
>>> model.getConvertUnknownCategoricalLevelsToNa()
True
```

The only downside of not having parameters, I can see, is doc generation. Parameters can keep useful information. E.g.: ```Param(Params._dummy(), "convertUnknownCategoricalLevelsToNa", "If set to 'true', the model converts unknown categorical levels to NA during making predictions.")```.
```
I'm not sure whether it's also relevant to Pysparkling or not.

PR #1209, opened 2019-05-21
[SW-1287] Use RPC endpoints to orchestrate cloud in internal mode

jakubhava:


mn-mikke:
@jakubhava Any idea why the integration tests are failing?

jakubhava:
CI/hadoop issues

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -0,0 +1,61 @@
...
+        val h2oNode = H2ONode.intern(InetAddress.getByName(ip), internalH2OPort)
+        H2O.addNodeToFlatfile(h2oNode)
+      }
+    case StopEndpointMsg =>
+      this.stop()
+  }
+
+  override def receiveAndReply(context: RpcCallContext): PartialFunction[Any, Unit] = {
+    case StartH2OWorkersMsg(conf) =>
+      val nodeDesc = InternalH2OBackend.startH2OWorker(conf)
```
mn-mikke: what if ```InternalH2OBackend.startH2OWorker(conf)``` throws an exception? Shouldn't we reply with a message that it wasn't possible to start a H2O worker?

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -0,0 +1,61 @@
...
+  */
+class H2ORpcEndpoint(override val rpcEnv: RpcEnv)
+  extends ThreadSafeRpcEndpoint with Logging {
+
+  override def receive: PartialFunction[Any, Unit] = {
+    case FlatFileMsg(nodes, portOffset) =>
+      nodes.map { pair =>
+        val ip = pair.hostname
+        // FlatFile contains API ports, but to directly add a node, we need to use internal port which is formed as
+        // API_PORT  + PORT_OFFSET
```
mn-mikke: nit: an extra space in ```API_PORT + PORT_OFFSET```

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendUtils.scala
```diff
@@ -73,162 +53,72 @@ private[internal] trait InternalBackendUtils extends SharedBackendUtils {
...
+      }
+    }
+    (
+      getH2OCommonArgs(conf)
+        ++ (if (!conf.h2oNodeWebEnabled) Seq("-disable_web") else Nil)
+        ++ Seq("-log_level", conf.h2oNodeLogLevel)
+        ++ Seq("-baseport", conf.nodeBasePort.toString)
+        ++ Seq("-log_dir", InternalBackendUtils.getH2ONodeLogDir(conf, SparkEnv.get))
+        ++ conf.nodeNetworkMask.map(mask => Seq("-network", mask)).getOrElse(Seq("-ip", ip))
+      ).toArray
```
mn-mikke: nit: indentation of parentheses

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -62,78 +63,105 @@ class InternalH2OBackend(@transient val hc: H2OContext) extends SparklingBackend
...
+}
+
+object InternalH2OBackend extends Logging {
+
+  val UNSUPPORTED_SPARK_OPTIONS: Seq[(String, String)] = Seq(
+    ("spark.dynamicAllocation.enabled", "true"),
+    ("spark.speculation", "true"))
+
+  private def startH2OCluster(hc: H2OContext): Array[NodeDesc] = {
+    if (hc.sparkContext.isLocal) {
```
mn-mikke: Nice!!!

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -62,78 +63,105 @@ class InternalH2OBackend(@transient val hc: H2OContext) extends SparklingBackend
...
+      Array(InternalH2OBackend.startH2OWorkerAsClient(hc._conf))
```
mn-mikke: Can we avoid usage of ```InternalH2OBackend.```?

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -62,78 +63,105 @@ class InternalH2OBackend(@transient val hc: H2OContext) extends SparklingBackend
...
-    RestAPIManager(hc).registerAll()
-    H2O.startServingRestApi()
-    executors
   }
 
-  override def epilog = ""
-}
+  private def startH2OWorkers(endpoints: Array[RpcEndpointRef], conf: H2OConf): Array[NodeDesc] = {
+    endpoints.map {
+      _.askSync[NodeDesc](StartH2OWorkersMsg(conf))
```
mn-mikke: Would it be nice to handle also situations when the start is not successful.

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
```diff
@@ -18,106 +18,111 @@
...
+  def getRef(conf: SparkConf): RpcEndpointRef = synchronized {
+    if (ref == null) {
+      ref = startEndpointOnH2OWorker(conf)
+    }
+    ref
+  }
+
+  private def startEndpointOnH2OWorker(conf: SparkConf): RpcEndpointRef = {
+    val securityMgr = SparkEnv.get.securityManager
+    val rpcEnv = RpcEnv.create("service", SharedBackendUtils.getHostname(SparkEnv.get), 0, conf, securityMgr)
```
mn-mikke: Can't the name of newly created ```RpcEnv``` get into a conflict with existing ones?

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
```diff
@@ -18,106 +18,111 @@
...
     }
   }
 
   private def isBackendReady() = sc.schedulerBackend.isReady()
 }
+
+object RpcReferenceCache {
+  private var ref: RpcEndpointRef = _
+
+  def getRef(conf: SparkConf): RpcEndpointRef = synchronized {
```
mn-mikke: What object is used for locking in this case? Isn't  it ```this```?

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
jakubhava: I have just tried starting the same endpoint twice and it just starts on a new port, no collision.

However I chose better name then `service`. Thanks, good point!

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
jakubhava: yup, this -> we should synchronize on `RpcReferenceCache`, thanks!

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
jakubhava: Definitely good idea - https://0xdata.atlassian.net/browse/SW-1302

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
jakubhava: Definitely good idea https://0xdata.atlassian.net/browse/SW-1302

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
mn-mikke: I'm afraid that `RpcReferenceCache` is pointing to the same instance as `this`.  The reason of my question was that we should limit the scope of synchronization only to `RpcReferenceCache`. Another thread could acquire lock  outside of `RpcReferenceCache` and this might lead to problems. Apologize, I should have been more explicit about this problem.

This will cause a deadlock for instance: 
```
object SomeObject {
  assert(this == SomeObject)
  def someMethod() = synchronized {
    println(this)
    println("Hello World!")
  }
}

object Application extends App {
  SomeObject.synchronized{
    val thread = new Thread(
      new Runnable {
        def run(): Unit = {
          SomeObject.someMethod()
        }
      })
    thread.start()
    thread.join()
  }
}
```

I good practice is to create a private instance dedicated just for locking. WDYT?

jakubhava: Great catch, thank you!

PR #1211, opened 2019-05-22
[SW-1259] Unify ratio param across pipeline api

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OCommonParams.scala
```diff
@@ -28,6 +28,8 @@ trait H2OCommonParams extends Params with Logging {
   private val labelCol = new Param[String](this, "labelCol", "Label column name")
   private val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
   private val weightCol = new NullableStringParam(this, "weightCol", "Weight column name")
+  private val splitRatio = new DoubleParam(this, "splitRatio",
+    "Accepts value from 0 to 1.0 and determine how large part of dataset will be used for training and for validation")
```
mn-mikke: Can you add notion that the number indicates the size of the training dataset? E. g. 0.8 -> 80% training 20% validation?

PR #1212, opened 2019-05-22
[SW-1289] Move shared params to H2OCommonParams

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -433,14 +428,11 @@ trait H2OGridSearchParams extends H2OCommonParams with DeprecatableParams {
     strategy -> HyperSpaceSearchCriteria.Strategy.Cartesian,
     maxRuntimeSecs -> 0,
     maxModels -> 0,
```
mn-mikke: Shouldn't ```allStringColumnsToCategorical``` and ```columnsToCategorical``` be removed as well?

py/pysparkling/ml/params.py
```diff
@@ -48,21 +77,36 @@ def setWeightCol(self, value):
...
+    def setNfolds(self, value):
+        assert_is_type(value, int)
+        return self._set(nfolds=value)
+
+    def setAllStringColumnsToCategorical(self, value):
+        assert_is_type(value, bool)
+        return self._set(allStringColumnsToCategorical=value)
+
+    def setColumnsToCategorical(self, value):
+        assert_is_type(value, [str])
```
mn-mikke: Shouldn't we also expose an overloaded method with a variable number of parameters?

jakubhava: done

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
jakubhava: Done

PR #1215, opened 2019-05-23
[SW-1298] Don't use deprecated methods

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "[SW=1298] Don't use deprecated methods" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #1218, opened 2019-05-23
[SW-1301] Fix class-loading for Sparkling Water assembly JAR in PySparkling

jakubhava:
Going to manually test on Databricks Cloud(DBC)

### DIFF
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -117,5 +117,4 @@ class Initializer(object):
                 return cl
             else:
                 cl = cl.getParent()
-            cl = cl.getParent()
         return None
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -120,4 +120,6 @@ class Initializer(object):
                 cl.addURL(url)
                 Initializer.__get_logger(jvm).info("Adding " + url.toString() + " to classloader"
                                                                                 " '" + cl.toString() + "'")
+            else:
+                Initializer.__get_logger(jvm).info("Skipping classloader '" + cl.toString() + "'")
             cl = cl.getParent()
--- a/py/pysparkling/initializer.py
+++ b/py/pysparkling/initializer.py
@@ -46,8 +46,7 @@ class Initializer(object):
 
         url = jvm.java.net.URL("file://{0}".format(sw_jar_file))
 
-        loader = Initializer.__find_spark_cl(jvm)
-        loader.addURL(url)
+        Initializer.__add(jvm, url)
 
         # Add Sparkling Water Assembly JAR to Spark's file server so executors can fetch it
         # when they need to use the dependency.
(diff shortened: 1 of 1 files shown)


py/pysparkling/initializer.py
```diff
@@ -46,21 +45,13 @@ def __add_sparkling_jar_to_spark(sc):
...
-        cl = jvm.Thread.currentThread().getContextClassLoader()
 
-        # Explicitly check if we run on databricks cloud since there we must add the jar to
-        # the parent of context class loader
-        if cl.getClass().getName() == 'com.databricks.backend.daemon.driver.DriverLocal$DriverLocalClassLoader':
-            cl.getParent().getParent().addURL(url)
-        else:
-            spark_cl = Initializer.__find_spark_cl(cl, 'org.apache.spark.util.MutableURLClassLoader')
-            spark_cl.addURL(url)
+        loader = Initializer.__find_spark_cl(jvm)
```
mn-mikke: @jakubhava WDYT about adding the url into each class loader from the hierarchy that has the `addUrl` method. I'm just thinking about cases where Spark is breaking standard lookup from parents to children. E.g. ```ExecutorClassLoader```.

jakubhava: Good idea! Will do

mn-mikke: Just one more note to that: I would also add some logging... (the list of loaders where url was added or log an error if there isn't such a classloader)

jakubhava:
@mn-mikke please have a second look, added the logging

py/pysparkling/initializer.py
```diff
@@ -118,11 +108,18 @@ def __get_sw_jar():
...
+    def __add_url_to_classloader(jvm, url):
+        cl = jvm.Thread.currentThread().getContextClassLoader()
         while cl:
-            name = cl.getClass().getName()
-            if (name == cl_name):
-                return cl
+            methods = [m.getName() for m in cl.getClass().getDeclaredMethods()]
+            if "addURL" in methods:
+                cl.addURL(url)
+                Initializer.__get_logger(jvm).info("Adding " + url.toString() + " to classloader"
```
mn-mikke: nit: Would it be possible to use some string interpolation (E.g. `.format(...)`) and have the message at one line?

jakubhava: Good idea, done!

jakubhava:
Should be good, please have a final look @mn-mikke

py/pysparkling/initializer.py
```diff
@@ -118,11 +108,18 @@ def __get_sw_jar():
...
+        cl = jvm.Thread.currentThread().getContextClassLoader()
+        logger = Initializer.__get_logger(jvm)
         while cl:
-            name = cl.getClass().getName()
-            if (name == cl_name):
-                return cl
+            methods = [m.getName() for m in cl.getClass().getDeclaredMethods()]
+            if "addURL" in methods:
+                cl.addURL(url)
+                logger.info("Adding {} to classloader '{}'".format(url.toString(), cl.toString()))
```
mn-mikke: This  looks great. I'm just wondering whether we shouldn't use the debug logging level instead. WDYT?

jakubhava: yup

PR #1219, opened 2019-05-24
[SW-1292] Project Skeleton for Benchmarks

mn-mikke:


assembly/build.gradle
```diff
@@ -97,6 +100,19 @@ shadowJar {
 
 
         include(dependency("org.eclipse.jetty:.*"))
+        include(dependency("org.eclipse.jetty.orbit:.*"))
```
jakubhava: Thank you for the fixes!

jakubhava: What do you think about switching to format which we discussed (and used in benchmarks)? ->  transitively include everything and exclude what we don't need as it is hard to track all the dependencies

mn-mikke: Yes, I will do it in a separate PR.

PR #1221, opened 2019-05-29
[SW-1304] Definition of assembly jar via transitive exclusions

mn-mikke:


gradle/shadowRuntime.gradle
```diff
@@ -0,0 +1,27 @@
...
+        exclude group: 'log4j', module: 'log4j' // a dependency of org.apache.hadoop:hadoop-auth
+        exclude group: 'com.google.protobuf' // a dependency of org.apache.hadoop:hadoop-common
+        exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core' // a dependency of org.apache.spark:spark-sql_2.11
+        exclude group: 'org.apache.httpcomponents' // a dependency of org.apache.hadoop:hadoop-auth
+        exclude group: 'com.github.rwl', module: 'jtransforms' // a dependency of org.apache.spark:spark-mllib
+        exclude group: 'com.google.code.findbugs', module: 'jsr305' // a dependency of org.apache.hadoop:hadoop-common
+        exclude group: 'javax.xml.bind', module: 'jaxb-api' // a dependency of org.apache.hadoop:hadoop-yarn-common
+        exclude group: 'net.sourceforge.f2j', module: 'arpack_combined_all' // a dependency of org.apache.spark:spark-graphx_2.11
+    }
+}
```
jakubhava: Nit : new line

gradle/shadowRuntime.gradle
```diff
@@ -0,0 +1,27 @@
+//
+// Contains runtime configuration for SW fat jars. This configuration transitively removes dependencies provided by
```
jakubhava: Can we replace `SW` with `Sparkling Water` ?

benchmarks/build.gradle
```diff
@@ -52,6 +30,8 @@ shadowJar {
     relocate 'org.eclipse.jetty', 'ai.h2o.org.eclipse.jetty'
     relocate 'org.eclipse.jetty.orbit', 'ai.h2o.org.eclipse.jetty.orbit'
     relocate 'com.google.common', 'ai.h2o.com.google.common'
+
+    exclude 'www/flow/packs/test-*/**'
```
jakubhava: I would create JIRA on H2O project, just to let folks know this can be removed

assembly/build.gradle
```diff
@@ -1,121 +1,43 @@
...
+    compile "org.apache.spark:spark-mllib_${scalaBaseVersion}:${sparkVersion}"
+    compile "org.apache.spark:spark-repl_${scalaBaseVersion}:${sparkVersion}"
 }
 
 jar {
     enabled = false // we do not need to generate jar file in this case
     archiveBaseName = "${jar.archiveBaseName.get()}_${scalaBaseVersion}"
 }
 
+apply plugin: 'com.github.johnrengelman.shadow'
```
jakubhava: I personally prefer to have plugin specifications at the top of the file

assembly/build.gradle
```diff
@@ -1,121 +1,43 @@
...
 
 description = "Sparkling Water Assembly"
 
 dependencies {
-    // Sparkling Water Core
+    // Sparkling Water libraries
+    compile project(":sparkling-water-ml")
+    compile project(":sparkling-water-repl")
     compile(project(":sparkling-water-core")) {
         exclude group: "javax.servlet", module: "servlet-api"
```
jakubhava: This was here before, but just an idea -> can we move this exclude to the `shadowRuntime` configuration?

jakubhava:
integration tests are failing because they don't have the classes available.

jakubhava:
I think that for now, we can put back the examples as they are really small and we can seperate the jars in the next iterations after release. What do you think @mn-mikke ?

mn-mikke:
Ok, I will put them back and create a separate jira for the removal.

assembly/build.gradle
mn-mikke: No problem :-)

benchmarks/build.gradle
mn-mikke: Sounds good to me. @michalkurka WDYT about having the testing flows completely separated from `h2o-docs` examples?

assembly/build.gradle
mn-mikke: Great idea!

benchmarks/build.gradle
michalkurka: I think that can be done :)

PR #1225, opened 2019-05-30
[SW-1307] Expose binary model in spark pipeline stage

jakubhava:


doc/src/site/sphinx/tutorials/sw_xgboost.rst
```diff
@@ -36,6 +36,11 @@ Train the model. You can configure all the available XGBoost arguments using pro
     val estimator = new H2OXGBoost().setLabelCol("AGE")
     val model = estimator.fit(sparkFrame)
 
+You can also get H2O's binary model after calling *fit* method on the estimator as:
```
mn-mikke: nit: the \*fit* method? on->of?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -102,9 +102,9 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
         " your 'excludeAlgo', 'maxModels' or 'maxRuntimeSecs' properties.") with NoStackTrace
     }
 
-    val binaryModel = aml.leader()
-    val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
-    val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID(aml.leader()._parms.algoName()))
+    binaryModel = Some(aml.leader())
+    val mojoData = ModelSerializationSupport.getMojoData(binaryModel.get)
```
mn-mikke: Can we avoid usage of `get` method and introduce a local variable that is later encapsulated into the option assigned to `binaryModel`?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -122,9 +122,9 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
       H2OMOJOModel.createFromMojo(data, Identifiable.randomUID(s"${$(gridAlgoParams).algoName()}_mojoModel"))
     }
 
-    val binaryModel = selectModelFromGrid(grid)
-    val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
-    val model = H2OMOJOModel.createFromMojo(mojoData, Identifiable.randomUID("gridSearch_mojoModel"))
+    binaryModel = Some(selectModelFromGrid(grid))
+    val mojoData = ModelSerializationSupport.getMojoData(binaryModel.get)
```
mn-mikke: ditto

py/py_sparkling/ml/algos.py
```diff
@@ -28,7 +30,42 @@ def propagate_value_from_deprecated_property(kwargs, from_deprecated, to_replaci
...
+
+    def getBinaryModel(self):
+        model_id = py4j.java_gateway.get_field(self._java_obj.getBinaryModel(), "_key").toString()
+
+        model_json = h2o.api("GET /%d/Models/%s" % (3, model_id))["models"][0]
+        metrics_class, model_class = H2OEstimator._metrics_class(model_json)
+        m = model_class()
+        m._id = model_id
+        m._model_json = model_json
+        m._have_pojo = model_json.get('have_pojo', True)
```
mn-mikke: Do we need a pojo?

py/py_sparkling/ml/algos.py
```diff
@@ -28,7 +30,42 @@ def propagate_value_from_deprecated_property(kwargs, from_deprecated, to_replaci
...
+        m._have_mojo = model_json.get('have_mojo', True)
+        m._metrics_class = metrics_class
+        m._start_time = model_json.get('output', {}).get('start_time', None)
+        m._end_time = model_json.get('output', {}).get('end_time', None)
+        m._run_time = model_json.get('output', {}).get('run_time', None)
+    
+        if model_id is not None and model_json is not None and metrics_class is not None:
```
mn-mikke: Can we break this if statement into smaller private methods?

doc/src/site/sphinx/tutorials/sw_xgboost.rst
jakubhava: Thx! What do you think about this rewording? `You can also get H2O's binary model by calling *fit* method available on the estimator as:` ?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: I was also thinking of instead of having binaryModel as Option, use BaseModel directly and use null instead of None. That would solve it, however I like options a bit more then using null

py/py_sparkling/ml/algos.py
jakubhava: This is based on the H2O's code -> wanted to keep parity of the information

doc/src/site/sphinx/tutorials/sw_xgboost.rst
mn-mikke: I would also add 'the' before *fit*, otherwise LGTM :-)

py/py_sparkling/ml/algos.py
mn-mikke: ok, no problem :-)

py/py_sparkling/ml/algos.py
jakubhava: Yup, after having second look on the code it does code does not seem completely right even though based on H2O. Will verify. Thank you!

jakubhava:
I have found better way how to expose details from time the model was trained then mutating Estimator & exposing H2O's binary model

PR #1226, opened 2019-05-30
[SW-1305] Move ability to change behavior of MOJO models to MOJOLoader

mn-mikke:


jakubhava:
@mn-mikke we need to also update docs http://docs.h2o.ai/sparkling-water/2.4/latest-stable/doc/deployment/load_mojo.html?highlight=createfrommojo

http://docs.h2o.ai/sparkling-water/2.4/latest-stable/doc/deployment/scoring_mojo_pipeline.html?highlight=createfrommojo

ml/src/test/scala/org/apache/spark/ml/spark/models/H2OMOJOPipelineModelTest.scala
```diff
@@ -106,10 +108,11 @@ class H2OMOJOPipelineModelTest extends FunSuite with SparkTestContext {
     // Test data
     val df = spark.read.option("header", "true").csv("examples/smalldata/prostate/prostate.csv")
     // Test mojo
+    val mojoSettings = H2OMOJOSettings(namedMojoOutputColumns = false)
```
jakubhava: should be `namedMojoOutputColumns = false`

mn-mikke: Good catch, thanks!

jakubhava: I missed that during review, sorry

mn-mikke: There is no reason to apologize. It was my mistake ;-)

PR #1227, opened 2019-05-30
[SW-1312] Warn that default value of convertUnknownCategoricalLevelsToNa will be changed to false on GridSearch & AutoML

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -43,6 +43,8 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
 
   // Override default values
   setDefault(nfolds, 5)
+  logWarning("Default value of 'convertUnknownCategoricalLevelsToNa' parameter on H2OAutoML will be changed to 'false' from 'true' in the next" +
```
mn-mikke: I would remove 'on H2OAutoML'  and also avoid string concatenation.

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -50,6 +50,8 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
   with H2OAlgorithmCommons with DefaultParamsWritable with H2OGridSearchParams {
 
   // Override default values
+  logWarning("Default value of 'convertUnknownCategoricalLevelsToNa' parameter on H2OGridSearch will be changed to 'false' from 'true' in the next" +
```
mn-mikke: ditto

mn-mikke:
@jakubhava Shouldn't we also raise warning on the Python site?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: Thanks!

jakubhava:
The warning on Python side is printed automatically from Scala

mn-mikke:
Ok, cool!

jakubhava:
@mn-mikke please have a look

PR #1230, opened 2019-05-31
[SW-1307] Expose model details

jakubhava:
Second iteration -> return JSON with raw model details as that will meet the requirement and is, at least a bit cleaner than returning directly H2O's java object. We can make it nicer in the future and parse internally the json and provide more fine-grained methods

### DIFF
--- a/doc/src/site/sphinx/tutorials/sw_xgboost.rst
+++ b/doc/src/site/sphinx/tutorials/sw_xgboost.rst
@@ -86,7 +86,7 @@ You can also get raw model details by calling the *getModelDetails()* method ava
 
 .. code:: python
 
-    print(model.getModelDetails())
+    model.getModelDetails()
 
 Run Predictions
 
--- a/doc/src/site/sphinx/tutorials/sw_xgboost.rst
+++ b/doc/src/site/sphinx/tutorials/sw_xgboost.rst
@@ -36,6 +36,11 @@ Train the model. You can configure all the available XGBoost arguments using pro
     val estimator = new H2OXGBoost().setLabelCol("AGE")
     val model = estimator.fit(sparkFrame)
 
+You can also get raw model details by calling the *getModelDetails()* method available on the estimator as:
+
+.. code:: scala
+
+    model.getModelDetails()
 
 Run Predictions
 
@@ -76,6 +81,12 @@ Train the model. You can configure all the available XGBoost arguments using pro
     estimator = H2OXGBoost(labelCol="AGE")
     model = estimator.fit(spark_frame)
 
+
+You can also get raw model details by calling the *getModelDetails()* method available on the estimator as:
+
+.. code:: python
+
+    print(model.getModelDetails())
 
 Run Predictions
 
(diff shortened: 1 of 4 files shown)


doc/src/site/sphinx/tutorials/sw_xgboost.rst
```diff
@@ -77,6 +82,12 @@ Train the model. You can configure all the available XGBoost arguments using pro
     model = estimator.fit(spark_frame)
 
 
+You can also get raw model details by calling the *getModelDetails()* method available on the estimator as:
+
+.. code:: python
+
+    print(model.getModelDetails())
```
mn-mikke: Scala exemple doesn't print details to console. It would be nice if it was consistent.

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -211,11 +217,38 @@ class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOMod
 
 object H2OMOJOModel extends H2OMOJOReadable[PyH2OMOJOModel] with H2OMOJOLoader[PyH2OMOJOModel] {
 
+  private def removeMetaField(json: JsonElement): JsonElement = {
+    if (json.isJsonObject) {
+      json.getAsJsonObject.remove("__meta")
+      json.getAsJsonObject.entrySet().asScala.map(_.getValue).foreach(removeMetaField)
```
mn-mikke: Can we merge `map` and `foreach` logic together? If we did that, we would avoid materialization of the collection produced by `map`.

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -211,11 +217,38 @@ class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOMod
...
+    if (json.isJsonArray) {
+      json.getAsJsonArray.asScala.foreach(removeMetaField)
+    }
+    json
+  }
+
+
+  private def getModelDetails(mojoData: Array[Byte]): String = {
+    val is = new ByteArrayInputStream(mojoData)
+    val reader = MojoReaderBackendFactory.createReaderBackend(is, MojoReaderBackendFactory.CachingStrategy.MEMORY)
```
mn-mikke: Just an idea... :-) WDYT about doing the following?
* move these two lines to `createFromMojo` 
* call `ModelMojoReader.readFrom(reader)` in `createFromMojo` instead of `ModelSerializationSupport.getMojoModel(mojoData)`
* pass the `reader` to `getModelDetails` instead of `mojoData`

jakubhava: Perfect idea! I had it like this originally as I wanted to reuse `reader`. However the method `ModelMojoReader.readFrom(reader)` has ugly side effect of closing the reader so we had to create a new reader

doc/src/site/sphinx/tutorials/sw_xgboost.rst
jakubhava: Agree! The issue here is that Python repl prints multi-line strings on one line with `\n` in it. So when we do:

`details = model.getModelDetails()` it is printed that ugly, but in Scala just fine. One way is to change method to `printModelDetails` however it does not sound right to me.

Any ideas please?

PR #1234, opened 2019-06-03
[SW-1317 Make repl unified accross different rel branches

jakubhava:


repl/src/main/spark_2.2/org/apache/spark/repl/h2o/SparkSpecificUtils.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def getJars(conf: SparkConf): String = {
+
+    import scala.reflect.runtime.{universe => ru}
+    val instanceMirror = ru.runtimeMirror(this.getClass.getClassLoader).reflect(Utils)
+    val methodSymbol = ru.typeOf[Utils.type].decl(ru.TermName("getLocalUserJarsForShell"))
+    val jars = if (methodSymbol.isMethod) {
+      val method = instanceMirror.reflectMethod(methodSymbol.asMethod)
+      method(conf).asInstanceOf[Seq[String]]
+    } else {
+      // Fallback to Spark 2.2.0
```
mn-mikke: Is this `if-else` branch specific only to 2.2.0 and the first one to other Spark 2.2.x versions?

jakubhava: Yup, it is. I will make it more specific in the comments, thanks!

PR #1237, opened 2019-06-03
[SW-1319] Remove jetty relocation

jakubhava:


assembly/build.gradle
```diff
@@ -30,9 +30,6 @@ shadowJar {
     mergeServiceFiles()
 
     relocate 'javassist', 'ai.h2o.javassist'
-    // Relocate Eclipse Jetty to avoid collisions with Spark Jetty (newer version)
-    relocate 'org.eclipse.jetty', 'ai.h2o.org.eclipse.jetty'
```
mn-mikke: Please can you do the same in `benchmarks/build.gradle`?

jakubhava: Good catch!

mn-mikke:
@jakubhava Do you think that would be possible to exclude jetty in `gradle/shadowRuntime.gradle` and let Spark to enforce its own version?

jakubhava:
That should work now when we have support for jetty 9 and Spark 2.1 and higher is using that. I will give it a try! Good suggestion

jakubhava:
Gave it a try, after exclusion

```
java.lang.NoClassDefFoundError: org/eclipse/jetty/server/Handler
  at water.webserver.jetty9.Jetty9Facade.createWebServer(Jetty9Facade.java:12)
  at water.init.NetworkInit.initializeNetworkSockets(NetworkInit.java:77)
  at water.H2O.startLocalNode(H2O.java:1621)

```

seems like jetty is shadowed in Spark

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -136,15 +136,14 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
...
-      val listener = new AppStatusListener(_conf.sparkConf, kvStore, live = true)
-      //sparkContext.listenerBus.addToStatusQueue(listener)
-      sparkContext.addSparkListener(listener)
-      val statusStore = new AppStatusStore(kvStore, Some(listener))
-      new SparklingWaterUITab(statusStore, sparkContext.ui.get)
-    }
+
+    val kvStore = sparkContext.statusStore.store.asInstanceOf[ElementTrackingStore]
+    val listener = new AppStatusListener(_conf.sparkConf, kvStore, live = true)
+    //sparkContext.listenerBus.addToStatusQueue(listener)
```
mn-mikke: Any reason for having the commented line?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -136,15 +136,14 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
...
+    sparkContext.addSparkListener(listener)
+    val statusStore = new AppStatusStore(kvStore, Some(listener))
+    new SparklingWaterUITab(statusStore, sparkContext.ui.get)
+
```
mn-mikke: nit: extra line

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: no reason, thanks!

PR #1239, opened 2019-06-04
[SW-1318] Add back the asDataFrame(.., SQLContext) method to JavaH2OContext

mn-mikke:


mn-mikke:
What about master?

jakubhava:
**rel-2.4** and **master** introduce the removal of sqlContext about a year ago, so we broke API only on rel-2.1, rel-2.2 and rel-2.3

mn-mikke:
ok, thanks!

mn-mikke:
Merged the PR to rel-2.3, rel-2.2 and rel-2.1 manually

PR #1242, opened 2019-06-05
[SW-1323] Expose 'modelId' as a parameter of H2O algorithms

mn-mikke:
The 'modelId' parameter will be exposed on:
* H2OGBM
* H2OGLM
* H2OXGBoost
* H2ODeepLearning

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
@@ -36,8 +36,7 @@ import scala.reflect.{classTag, ClassTag}
   * Base class for H2O algorithm wrapper as a Spark transformer.
   */
 abstract class H2OAlgorithm[B <: H2OBaseModelBuilder : ClassTag, M <: H2OBaseModel, P <: Model.Parameters : ClassTag]
-  extends Estimator[H2OMOJOModel]
-  with H2OAlgorithmCommons with DefaultParamsWritable with H2OAlgoParams[P] {
+  extends Estimator[H2OMOJOModel] with H2OAlgorithmCommons with DefaultParamsWritable with H2OAlgoParams[P] {
 
   override def fit(dataset: Dataset[_]): H2OMOJOModel = {
     // Update H2O params based on provided configuration
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
@@ -16,7 +16,7 @@
 */
 package org.apache.spark.ml.h2o.algos
 
-import hex.Model
+import hex.{Model, ModelBuilder}
 import hex.genmodel.utils.DistributionFamily
 import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.h2o._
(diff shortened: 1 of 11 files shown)


ml/src/test/scala/org/apache/spark/ml/spark/models/H2OAlgoTest.scala
```diff
@@ -58,6 +60,45 @@ class H2OAlgoTest extends FunSuite with SharedH2OTestContext {
     loadedModel.transform(dataset).count()
   }
 
+  test("H2OGLM with set modelId is trained mutliple times")
+  {
```
jakubhava: NIT: any reason why `{` is on the new line ?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGBM.scala
```diff
@@ -29,7 +29,11 @@ class H2OGBM(override val uid: String) extends H2OAlgorithm[GBMParameters] with
 
   def this() = this(Identifiable.randomUID("gbm"))
 
-  override def trainModel(params: GBMParameters): GBMModel = new GBM(params).trainModel().get()
+  override def trainModel(params: GBMParameters): GBMModel = {
+    val modelId = getModelId()
+    val builder = if (modelId == null || modelId.isEmpty) new GBM(params) else new GBM(params, createKey(modelId))
```
jakubhava: This code is duplicated across all H2OAlgorithms. I can see that we create different algo param instance in each, but what do you think about encapsulating this functionality into a method  which takes type parameter( such as `GBM`) and can create new instance using reflection?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -85,4 +85,20 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag] extends Estimator[
   }
 
   override def copy(extra: ParamMap): this.type = defaultCopy(extra)
+
+  protected def createKey[T <: Keyed[T]](modelId: String): Key[T] = {
+    val key = Key.make[T](modelId)
+    if(H2O.containsKey(key)) {
```
jakubhava: NIT: formatting - missing space after`if`

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -85,4 +85,20 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag] extends Estimator[
...
+      var i = 0
```
jakubhava: what do you think of renaming `i` to something like `suffix`?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
```diff
@@ -85,4 +85,20 @@ abstract class H2OAlgorithm[P <: Model.Parameters : ClassTag] extends Estimator[
...
+      var replacement = key
+      do {
```
jakubhava: Just nit: Maybe we could expose the replacement functionality to a new, small private method and keep this method as:

if(keyExists){
 createReplacement()
} else {
key}

ml/src/test/scala/org/apache/spark/ml/spark/models/H2OAlgoTest.scala
mn-mikke: There is no reason :-) I will fix it, thanks!

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGBM.scala
mn-mikke: Yeah, that' a great idea! Will do, thanks!

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAlgorithm.scala
mn-mikke: Yup, no problem :-)

mn-mikke:
@jakubhava Thank you for your valid comments! Please can you review the PR one more time?

PR #1252, opened 2019-06-10
[SW-1333] Move the conversion logic from Spark Row to H2O RowData toa separate entity

mn-mikke:


jakubhava:
@mn-mikke feel free to merge, it failed because jenkins killed the job during running script tests, otherwise all good

mn-mikke:
Will do, thanks a lot!

PR #1264, opened 2019-06-13
[SW-1346] Remove chache and unpersist call in SpreadRDDBuilder

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
```diff
@@ -56,13 +56,10 @@ class SpreadRDDBuilder(@transient private val hc: H2OContext,
     val expectedWorkers = numExecutors.orElse(numExecutorHint).getOrElse(if (nSparkExecBefore > 0) nSparkExecBefore else conf.defaultCloudSize)
 
     // Create some distributed data
-    val spreadRDD = sc.parallelize(0 until mfactor * expectedWorkers, mfactor * expectedWorkers + 1).persist()
-
+    val spreadRDD = sc.parallelize(0 until mfactor * expectedWorkers, mfactor * expectedWorkers + 1)
 
     // Start RPC Endpoint on all worker nodes
     val endpoints = spreadRDD.mapPartitions { _ => Iterator.single(RpcReferenceCache.getRef(sparkConf)) }.distinct().collect()
-    // Delete RDD
```
jakubhava: I don't think persisting RDD in this case has actual benefit as we use it just once. Also `unpersist` call may lead to RPC timeout issues (seen that on Spark 2.1.1), it can be fixed by increasing spark.rpc.timeout but ideally, if we have less places like that in code, the better.

Wanted to confirm with you @mn-mikke as well, thanks!

mn-mikke: If it used only once then there shouldn't any reason for adding it. Moreover, it will utilize some portion of memory that could be used for something else.

PR #1270, opened 2019-06-14
[SW-1326] Include SW version in more log statements

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -150,7 +150,7 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
     Log.trace("H2OContext initialized") // force creation of log files
     LogBridge.setH2OLogLevel(level) // set the level back on the level user want's to use
 
-    logInfo("Sparkling Water started, status of context: " + this)
+    logInfo(s"Sparkling Water ${BuildInfo.SWVersion} started, status of context: ${this}")
```
mn-mikke: Nit: Wouldn't be possible to use just `$this`?

PR #1271, opened 2019-06-14
[SW-1351] Unify jenkins test scripts accross all branches

jakubhava:


jakubhava:
Good question. This solution would be fine for now, but not ideal once we have a single branch. Still need to think about the right way how do to this without duplication.

The goal of this change was to minimize number of files we need to change when we bump spark version.

Thank you!

jakubhava:
Will go with different approach

PR #1276, opened 2019-06-18
[SW-1360] Explicitly send hearbeat after we have complete flatfile

jakubhava:


core/src/main/scala/water/HeartBeatUtils.scala
```diff
@@ -0,0 +1,26 @@
...
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package water
+
+object HeartBeatUtils {
+  def broadcastHeartbeat(): Unit = {
+    val ab = new AutoBuffer()
+    H2O.SELF._heartbeat.write(ab)._bb
```
jakubhava: note to myself: remove `._bb`

core/src/main/scala/water/HeartBeatUtils.scala
```diff
@@ -0,0 +1,26 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package water
+
+object HeartBeatUtils {
+  def broadcastHeartbeat(): Unit = {
```
jakubhava: rename `broadcastHearbeat` to `sendHeartbeatToAll()`

PR #1279, opened 2019-06-18
[SW-1353] Introduce logic flatting data frames with arbitrarily nested structures

mn-mikke:


core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+      }
+    }
+  }
+
+  private def fillArray(
+      flatSchemaIndexes: Map[String, Int],
+      buffer: ArrayBuffer[Any],
+      row: Row,
+      elementType: DataType,
+      index: Int,
```
jakubhava: I would pass `row.getSeq[Any](index)` as argument to this method in `fillBuffer` and remove index param from `fillArray`. I think `fillArray` method does not need to know about data it won't actually use.

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+      qualifiedName: String){
+    val seq = row.getSeq[Any](index)
+    val subRow = Row.fromSeq(seq)
+    val fillMethod = fillBuffer(flatSchemaIndexes, buffer, subRow)(_, _, _)
+    (0 until seq.size).foreach{ i =>
```
jakubhava: NIT: missing space after `foreach`

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+    val qualifiedName = getQualifiedName(prefix, name)
+    if(!row.isNullAt(index)) {
+      dataType match {
+        case BinaryType =>
+          fillArray(flatSchemaIndexes, buffer, row, ByteType, index, qualifiedName)
+        case MapType(_, valueType, containsNull) =>
+          val map = row.getMap[Any, Any](index)
+          val subRow = Row.fromSeq(map.values.toSeq)
+          val fillMethod = fillBuffer(flatSchemaIndexes, buffer, subRow)(_, _, _)
+          map.keys.zipWithIndex.foreach{ case (k, i) =>
```
jakubhava: NIT: Missing space after foreach.

Also what about renaming (k, i) to (key, idx) ?

I prefer the longer names for easier readability :D

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: Also i -> idx

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+            val mapField = StructField(k.toString, valueType, containsNull)
+            fillMethod(mapField, i, Some(qualifiedName))
+          }
+        case ArrayType(elementType, _) =>
+          fillArray(flatSchemaIndexes, buffer, row, elementType, index, qualifiedName)
+        case StructType(fields) =>
+          val subRow = row.getStruct(index)
+          val fillMethod = fillBuffer(flatSchemaIndexes, buffer, subRow)(_, _, _)
+          fields.zipWithIndex.foreach { case (f, i) => fillMethod(f, i, Some(qualifiedName)) }
```
jakubhava: DTTO short variable names

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+      (flatSchemaIndexes: Map[String, Int], buffer: ArrayBuffer[Any], row: Row)
+      (field: StructField, index: Int, prefix: Option[String] = None): Unit = {
+    val StructField(name, dataType, _, _) = field
+    val qualifiedName = getQualifiedName(prefix, name)
+    if(!row.isNullAt(index)) {
+      dataType match {
+        case BinaryType =>
+          fillArray(flatSchemaIndexes, buffer, row, ByteType, index, qualifiedName)
+        case MapType(_, valueType, containsNull) =>
+          val map = row.getMap[Any, Any](index)
```
jakubhava: What do you think about exporting functionality for filling map and struct to separate methods? Just like fillArray

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+        case dt => FieldWithOrder(StructField(qualifiedName, dt, nullableField), path.reverse) :: Nil
+      }
+    } else {
+      Nil
+    }
+  }
+
+  private case class FieldWithOrder(field: StructField, order: Seq[Any])
+
+  private def flattenArray(
```
jakubhava: I would somehow rename this method to explicitly tell that this logic just flattens schema

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+      row: Row,
+      elementType: DataType,
+      containsNull: Boolean,
+      index: Int,
+      path: List[Any],
+      qualifiedName: String,
+      nullableField: Boolean) = {
+    val values = row.getSeq[Any](index)
+    val subRow = Row.fromSeq(values)
+      (0 until values.size).flatMap{ i =>
```
jakubhava: DTTO - space after {

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+    }
+  }
+
+  private case class FieldWithOrder(field: StructField, order: Seq[Any])
+
+  private def flattenArray(
+      row: Row,
+      elementType: DataType,
+      containsNull: Boolean,
+      index: Int,
```
jakubhava: DTTO: Would also remove the index and pass just the data this method works on instead of the full row and idx

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+      isParentNullable: Boolean = false): Seq[FieldWithOrder] = {
+    val StructField(name, dataType, nullable, _) = originalField
+    val qualifiedName = getQualifiedName(prefix, name)
+    val nullableField = isParentNullable || nullable
+    if(!row.isNullAt(index)) {
+      dataType match {
+        case BinaryType =>
+          flattenArray(row, ByteType, false, index, path, qualifiedName, nullableField)
+        case MapType(_, valueType, containsNull) =>
+          val map = row.getMap[Any, Any](index)
```
jakubhava: What do you think also about exporting extra methods for  map and struct? It would make this method shorter.
Ideally just:

datatype match {
 case A => callA()
case B => handleB()
}

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+
+  private def getQualifiedName(prefix: Option[String], name: String) = prefix match {
+    case None => name
+    case Some(p) => s"${p}_$name"
+  }
+
+  private def flattenField(
+      originalField: StructField,
+      row: Row,
+      index: Int,
```
jakubhava: I would also remove `row` and `index` and pass just row.get(index) for the reasons mentioned above as well

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+          val subRow = Row.fromSeq(map.values.toSeq)
+          map.keys.zipWithIndex.flatMap { case (k, i) =>
+            val mapField = StructField(k.toString, valueType, containsNull)
+            flattenField(mapField, subRow, i, k :: path, Some(qualifiedName), nullableField)
+          }.toSeq
+        case ArrayType(elementType, containsNull) =>
+          flattenArray(row, elementType, containsNull, index, path, qualifiedName, nullableField)
+        case StructType(fields) =>
+          val subRow = row.getStruct(index)
+          fields.zipWithIndex.flatMap { case (f, i) => flattenField(f, subRow, i, i :: path, Some(qualifiedName), nullableField) }
```
jakubhava: f, i to better names

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+    val fields = df.rdd.map[Seq[FieldWithOrder]] { row: Row =>
+        originalSchema.fields.zipWithIndex.foldLeft(Seq.empty[FieldWithOrder]) {
+          case (acc, (field, index)) => acc ++ flattenField(field, row, index, index :: Nil)
+        }
+      }
+      .reduce { (first, second) =>
+        def convertToMap(collection: Seq[FieldWithOrder]) = collection.map(f => f.order -> f.field).toMap
+        val firstMap = convertToMap(first)
+        val secondMap = convertToMap(second)
+        val pathItemOrdering = new Ordering[Any] {
```
jakubhava: Can we convert the names of field to string on the nested conversion methods so we don't have to specify this ordering? Also Seq[Any] could become Seq[String] as they represent just names

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+  }
+
+  def flattenSchema(df: DataFrame): StructType = {
+    val originalSchema = df.schema
+    val fields = df.rdd.map[Seq[FieldWithOrder]] { row: Row =>
+        originalSchema.fields.zipWithIndex.foldLeft(Seq.empty[FieldWithOrder]) {
+          case (acc, (field, index)) => acc ++ flattenField(field, row, index, index :: Nil)
+        }
+      }
+      .reduce { (first, second) =>
```
jakubhava: This part was a bit hard to read -> can we split it to more smaller methods?

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+    val fillMethod = fillBuffer(flatSchemaIndexes, buffer, subRow)(_, _, _)
+    (0 until seq.size).foreach{ i =>
+      val arrayField = StructField(i.toString, elementType)
+      fillMethod(arrayField, i, Some(qualifiedName))
+    }
+  }
+
+  def flattenSchema(df: DataFrame): StructType = {
+    val originalSchema = df.schema
+    val fields = df.rdd.map[Seq[FieldWithOrder]] { row: Row =>
```
jakubhava: I was thinking -> why do we need to process schema on per-row basis? Can't we just process schema from `df.schema` directly?

core/src/test/scala/org/apache/spark/h2o/utils/TestFrameUtils.scala
```diff
@@ -75,4 +77,21 @@ object TestFrameUtils {
...
+
+  def assertFieldNamesAreEqual(expected: DataFrame, produced: DataFrame): Unit = {
+    def fieldNames(df: DataFrame) = df.schema.fields.map(_.name)
+    val expectedNames = fieldNames(expected)
+    val producedNames = fieldNames(produced)
+    producedNames shouldEqual expectedNames
+  }
+
+  def assertDataFramesAreIdentical(expected: DataFrame, produced: DataFrame): Unit = {
+    val numberOfExtraColumnsInExpected = expected.exceptAll(produced).count()
```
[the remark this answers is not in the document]
mn-mikke: Good catch, thanks!

jakubhava:
Based on our private discussion -> Should this tool be able to also flatten vectors? I think that would be helpful if its generic method to flatten data frames.

What do you think of allowing vectors to be flatten based on some boolean flag? If set to `true`, they would be flatten`, if set to false, the code would just ensure that the vector is not nested in another element but it is at top-level structure

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -52,7 +52,7 @@ private[h2o] object SparkDataFrameConverter extends Logging {
   def toH2OFrame(hc: H2OContext, dataFrame: DataFrame, frameKeyName: Option[String]): H2OFrame = {
     import H2OSchemaUtils._
     // Flatten the Spark data frame so we don't have any nested rows
-    val flatDataFrame = flattenDataFrame(dataFrame)
+    val flatDataFrame = flattenStructsInDataFrame(dataFrame)
```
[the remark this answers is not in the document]
jakubhava: Never mind about this, it is actually a name describing exactly what it does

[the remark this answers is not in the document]
mn-mikke: This function is pointing to the original logic. It seems that flattening of arrays, vectors are currently performed by the `expandedSchema` method.
```
  def flattenStructsInSchema(schema: StructType, prefix: String = null, nullable: Boolean = false): StructType = {

    val flattened = schema.fields.flatMap { f =>
      val escaped = if (f.name.contains(".")) "`" + f.name + "`" else f.name
      val colName = if (prefix == null) escaped else prefix + "." + escaped

      f.dataType match {
        case st: StructType => flattenStructsInSchema(st, colName, nullable || f.nullable)
        case _ => Array[StructField](StructField(colName, f.dataType, nullable || f.nullable))
      }
    }
    StructType(flattened)
  }

  def flattenStructsInDataFrame(df: DataFrame): DataFrame = {
    import org.apache.spark.sql.functions.col
    val flatten = flattenStructsInSchema(df.schema)
    val cols = flatten.map(f => col(f.name).as(f.name.replaceAll("`", "")))
    df.select(cols: _*)
  }
```

core/src/test/scala/org/apache/spark/h2o/H2OSchemaUtilsTestSuite.scala
```diff
@@ -71,12 +76,246 @@ class H2OSchemaUtilsTestSuite extends FunSuite {
...
+    import spark.implicits._
+
+    val input = Seq(
+      Seq((1, 2), (3, 4)),
+      Seq((1, 2), (3, 4), (5, 6))
+    ).toDF("arr")
+    val expected = Seq[(Integer, Integer, Integer, Integer, Integer, Integer)](
+      (1, 2, 3, 4, null, null),
+      (1, 2, 3, 4, 5, 6)
+    ).toDF("arr_0__1", "arr_0__2", "arr_1__1", "arr_1__2", "arr_2__1", "arr_2__2")
```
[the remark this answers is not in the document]
mn-mikke: ok, thanks. Will do :-)

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -52,7 +52,7 @@ private[h2o] object SparkDataFrameConverter extends Logging {
...
     val dfRdd = flatDataFrame.rdd
     val keyName = frameKeyName.getOrElse("frame_rdd_" + dfRdd.id + Key.rand())
```
jakubhava: If I'm correct, the flatting method you introduced also handles arrays of different size. If if would be also handle vectors as described in the last comment, we could actually simplify the code starting from line 59 a lot ( maybe actually remove). In order to do so, the output of the flatten method should be:

Maps, Structs, Arrays, Dense vectors should be flatten. Sparse vectors should not be flatten, but should be also at the top-level schema structure and should have the same max size.

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -158,7 +158,7 @@ class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOMod
   override def copy(extra: ParamMap): H2OMOJOModel = defaultCopy(extra)
 
   override def transform(dataset: Dataset[_]): DataFrame = {
-    val flattenedDF = H2OSchemaUtils.flattenDataFrame(dataset.toDF())
+    val flattenedDF = H2OSchemaUtils.flattenStructsInDataFrame(dataset.toDF())
     val relevantColumnNames = flattenedDF.columns.intersect(getFeaturesCols())
     val args = relevantColumnNames.map(flattenedDF(_))
     flattenedDF.withColumn(getPredictionCol(), getModelUdf()(struct(args: _*)))
```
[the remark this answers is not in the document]
mn-mikke: Good catch!

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
-  def flattenSchema(schema: StructType, prefix: String = null, nullable: Boolean = false): StructType = {
+  def flattenDataFrame(df: DataFrame): DataFrame = {
+    val schema = flattenSchema(df)
+    flattenDataFrame(df, schema)
+  }
+
+  def flattenDataFrame(df: DataFrame, flatSchema: StructType): DataFrame = {
+    implicit val rowEncoder = RowEncoder(flatSchema)
+    val numberOfColumns = flatSchema.fields.length
+    val flatSchemaIndexes = flatSchema.fields.map(_.name).zipWithIndex.toMap
```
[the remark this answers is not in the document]
mn-mikke: yep, no problem. WDYT about `nameToIndexMap`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: I will gladly change it..., but just one note that. In that case, the `fillBuffer` will call `row.getSeq[Any](index)` twice:
```
case BinaryType =>
  val data = row.getSeq[Any](index)
  fillArray(flatSchemaIndexes, buffer, data, ByteType, qualifiedName)
case ArrayType(elementType, _) =>
  val data = row.getSeq[Any](index)
  fillArray(flatSchemaIndexes, buffer, data, elementType, qualifiedName)
```

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: Will do, thanks!

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: I'm quite not happy of `Seq[Any]` as well. This collection contains indexes of fields for each level of nesting to ensure that order of columns won't change after flattening.
E.g.: `{index for level 0} :: {index for level 1} :: {index for level 2} :: Nil`. I wanted to to use `Seq[Int]` but for `MapType` I have to use keys instead of indexes. 

It seems that conversion of number would break ordering. E.g.: `"10"` would be before `"1"`.

Any thoughts how to get rid of `Seq[Any]` and preserve ordering of columns?

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: it's because of variable length of arrays and also arrays might be arbitrarily nested. In this solution the `map` calculates a specific schema for each record and `reduce` does the merge logic.

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
mn-mikke: Yes, you're 100% correct. Let's do such changes once the general flattening is merged. :-)

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: thank you for the explanation!

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
jakubhava: Cool!

mn-mikke:
@jakubhava Many thanks for your valuable comments. I added some responses that try to clarify some question. Please take a look.

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: That is still fine by me -> I like that the method has now clearer API, at least as I see it. What do you think?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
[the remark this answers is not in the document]
mn-mikke: If you don't mind, I will do this change in a different PR.

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: No problem at all :-)

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -70,23 +72,174 @@ object H2OSchemaUtils {
...
+    val originalSchema = df.schema
+    df.map[Row]{ r: Row =>
+      val result = ArrayBuffer.fill[Any](numberOfColumns)(null)
+      val fillMethod = fillBuffer(flatSchemaIndexes, result, r)(_, _, _)
```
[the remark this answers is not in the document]
mn-mikke: we can replace `(_, _, _)` with just `_`. WDYT if I renamed `fillMethod` to `fillBufferPartiallyApplied`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
mn-mikke: Actually this a great idea, it will simplify several things. Thanks a lot!

core/src/test/scala/org/apache/spark/h2o/H2OSchemaUtilsTestSuite.scala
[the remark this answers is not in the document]
mn-mikke: If you don't mind will do it in another PR. This would have a bigger impact on the current tests.

mn-mikke:
@jakubhava I've tried to address your comments. Can you have a second look?

jakubhava:
Time to have a :beer: ! Thanks @mn-mikke.

PR #1284, opened 2019-06-19
[SW-1364] Rename SVM to SparkSVM

jakubhava:


ml/src/main/java/org/apache/spark/ml/spark/models/svm/SparkSVM.java
```diff
@@ -45,16 +45,16 @@
 
 import static scala.collection.JavaConversions.*;
 
-public class SVM extends ModelBuilder<SVMModel, SVMParameters, SVMOutput> {
+public class SparkSVM extends ModelBuilder<SVMModel, SparkSVMParameters, SVMOutput> {
```
mn-mikke: Shouldn't the prefix be added to  `SVMModel` and `SVMOutput` as well?

PR #1292, opened 2019-06-21
[SW-1375] Single execution path for all spark->h2o frame conversions

jakubhava:
**The benefits:** 
 - Single code handling all conversions - no duplication & fixes are on one places
 - Consistency across product - this change already helped me discover 2 inconsistencies and one bug in the original SparkDataFrame converter
- Less code to maintain
- Major use-case - Spark DataFrames is not affected

**Disadvantages:** 
- Extra conversions from RDD to DF before actual conversion. However the major use-case ( SparkFrame -> H2OFrame) remains unaffected. Converting from RDD is really a minor use-cases, I think it's beneficial sacrifice.


core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -231,9 +231,14 @@ object H2OSchemaUtils {
     val maxCollectionSizes = if (sizeFromMetadata.forall(_.isDefined)) {
       sizeFromMetadata.map(_.get).toArray
     } else {
-      flatDataFrame.rdd.map { row =>
+      val sizes = flatDataFrame.rdd.map { row =>
         collectionIndices.map { idx => getCollectionSize(row, idx) }
-      }.reduce((a, b) => a.indices.map(i => if (a(i) > b(i)) a(i) else b(i))).toArray
+      }
+      if (sizes.isEmpty) {
```
jakubhava: This is the bug for normal SparkDataFrame conversions. Will fail on empty collections

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -438,9 +438,14 @@ object H2OSchemaUtils {
...
+        Array(0)
+      } else {
+        sizes.reduce((a, b) => a.indices.map(i => if (a(i) > b(i)) a(i) else b(i))).toArray
```
mn-mikke: WDYT about rewriting the logic to Dataframe API? 
Something like: ```df.groupBy().max().collect()```

jakubhava: I think we will get rid of this after our flatting method supports flattening dense vectors ( and sparse vectors aligns to the same size)

PR #1302, opened 2019-06-26
[SW-1383] Fix cloud up in external backend manual mode

jakubhava:
I also did a bit of deprecations as the original name was way too long

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
@@ -123,7 +123,7 @@ object H2OConf extends Logging {
           val currentValue = sparkConf.getOption(current)
           if (currentValue.isDefined) {
             logWarning(s"Both options '$deprecated' and '$current' are specified. " +
-              s"Using value ${currentValue.get} of $current as the later one is deprecated.")
+              s"Using value '${currentValue.get}' of '$current' as the later one is deprecated.")
           } else {
             logWarning(s"Please use '$current' as '$deprecated' is deprecated. Passing the value '${deprecatedValue.get}' to '$current'.")
           }
--- a/core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
@@ -24,12 +24,15 @@ import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.spark.internal.Logging
 import org.apache.spark.repl.h2o.H2OInterpreter
 import org.apache.spark.sql.SparkSession
+import water.util.Log
 
 /**
   * Configuration holder which is representing
   * properties passed from user to Sparkling Water.
   */
 class H2OConf(val sparkConf: SparkConf) extends Logging with InternalBackendConf with ExternalBackendConf with Serializable {
+
+  H2OConf.checkDeprecatedOptions(sparkConf)
 
   /** Support for creating H2OConf in Java environments */
   def this(jsc: JavaSparkContext) = this(jsc.sc.getConf)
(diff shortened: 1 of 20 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -296,6 +299,10 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
       """)
     }
 
+    if (conf.clusterSize.isEmpty) {
```
jakubhava: Require num of external h2o nodes in both auto and manual mode

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -35,7 +36,10 @@ trait ExternalBackendConf extends SharedBackendConf {
   def h2oClusterHost = sparkConf.getOption(PROP_EXTERNAL_CLUSTER_REPRESENTATIVE._1).map(_.split(":")(0))
   def h2oClusterPort = sparkConf.getOption(PROP_EXTERNAL_CLUSTER_REPRESENTATIVE._1).map(_.split(":")(1).toInt)
 
-  def numOfExternalH2ONodes = sparkConf.getOption(PROP_EXTERNAL_H2O_NODES._1)
+  @DeprecatedMethod("clusterSize")
+  def numOfExternalH2ONodes = clusterSize
+
+  def clusterSize = sparkConf.getOption(PROP_EXTERNAL_CLUSTER_SIZE._1)
```
mn-mikke: Shouldn't the method fall back to  `"spark.ext.h2o.external.cluster.num.h2o.nodes"` if `"spark.ext.h2o.external.cluster.size"` is not set?

jakubhava: Yup, did that in train, but pushed now :)

jakubhava:
@mn-mikke I'm getting `enable macro paradise to expand macro annotations` failure in tests, wanted to ask if you have seen this when you were working on deprecation utilities. Thanks!

mn-mikke:
@jakubhava As far I remember, i had to add this [`dependency`](https://github.com/h2oai/sparkling-water/pull/1131/files#diff-1d4bec2ed0eb20eda669e2f83c2341c8R16) to get rid of the exception. After that, I haven't see it anymore.

jakubhava:
Thanks!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mn-mikke: If this function is used only explicitly by an user or the original option is very unlikely to be specified then I think we don't have to stress about that. But if other part of SW utilize this function and a customer has the original option stored somewhere in some config (e.g. `spark-defaults.conf`), an upgrade to the next minor version of SW may cause unexpected problems.

jakubhava: Good point, but I think we are still fine as we set the new value based on the deprecated value. I changed the code to centralize the logic. Now if you have the old option specified, and try to use H2OConf, you get warning
```
WARN H2OConf: Please use 'spark.ext.h2o.external.cluster.size' as 'spark.ext.h2o.external.cluster.num.h2o.nodes' is deprecated. Passing the value '1' to 'spark.ext.h2o.external.cluster.size'.
```

jakubhava:
@mn-mikke updated the code, mainly the deprecation so it's more explicit. Can you please check?

H2OConf is either created by the user explicitly or internally in the H2OContext if not specified. Either way, after first contact with it, we check deprecated values and pass them to new corresponding ones.

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -107,7 +110,29 @@ class H2OConf(val sparkConf: SparkConf) extends Logging with InternalBackendConf
...
+        val deprecatedValue = sparkConf.getOption(deprecated)
+        if (deprecatedValue.isDefined) {
+          val currentValue = sparkConf.getOption(current)
+          if (currentValue.isDefined) {
+            logWarning(s"Both options '$deprecated' and '$current' are specified. " +
+              s"Using value '${currentValue.get}' of '$current' as the later one is deprecated.")
+          } else {
+            logWarning(s"Please use '$current' as '$deprecated' is deprecated. Passing the value '${deprecatedValue.get}' to '$current'.")
+          }
+          sparkConf.set(current, deprecatedValue.get)
```
mn-mikke: If both are specified, doesn't you override the current with a value from deprecated?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mn-mikke: Sounds good to me :-)

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -107,7 +110,29 @@ class H2OConf(val sparkConf: SparkConf) extends Logging with InternalBackendConf
   }
 }
 
-object H2OConf {
+object H2OConf extends Logging {
+
+  private val deprecatedOptions = Map("spark.ext.h2o.external.cluster.num.h2o.nodes" -> ExternalBackendConf.PROP_EXTERNAL_CLUSTER_SIZE._1)
+
+  private def checkDeprecatedOptions(sparkConf: SparkConf): Unit = {
+
```
mn-mikke: nit: an extra line?

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
jakubhava: Very good catch!

PR #1306, opened 2019-06-27
[SW-1386] Micro benchmark for conversion from a DataFrame to H2OFrame

mn-mikke:


core/src/integTest/scala/water/sparkling/itest/local/DataFrameConverterBenchSuite.scala
```diff
@@ -0,0 +1,120 @@
...
+  test("Measure performance of conversion to H2OFrame on a data frame with nested structs") {
+    testPerSchema(StructsOnlySchema)
+  }
+
+  test("Measure performance of conversion to H2OFrame on a data frame with flat arrays") {
+    testPerSchema(FlatArraysOnlySchema)
+  }
+
+  def testPerSchema(schemaHolder: TestFrameUtils.SchemaHolder): Unit = {
+    val df= TestFrameUtils.generateDataFrame(spark, schemaHolder, settings)
```
jakubhava: NIT: missing space before `=`

core/src/integTest/scala/water/sparkling/itest/local/DataFrameConverterBenchSuite.scala
```diff
@@ -0,0 +1,120 @@
...
+    val result = bench(iterations = 5, warmup = 1) {
+      val hf = hc.asH2OFrame(df)
+      hf.remove()
+    }
+    println(result.show())
```
jakubhava: I think we should move this from integTests to a new collection directory called `core/src/bench/scala/water/sparkling/bench/DataFrameConverterBenchSuite.scala` and create task on each scala gradle project `./gradlew bench` which would execute all benchmarks in each suproject. 

Also we should ideally report the outcome of bechmarks in a better way -> create some report which would be also archived on jenkins.

There are also libraries for Gralde which can help with setting up microbenchmarking infra - for example https://github.com/melix/jmh-gradle-plugin

mn-mikke: WDYT about using the new namespace `ai.h2o.sparkling` for the microbenchmarks?

jakubhava: perfect!

jakubhava:
@mn-mikke Actually I was thinking. What about merging the current state of this branch and implement the next steps in a few next iterations & PRs? 

Just for the sake of having smaller PRs

mn-mikke:
@jakubhava I separated tests into a dedicated gradle task and it seems that PR didn't explode much. We can see the summary in `build/reports/tests/bench/index.html` or via `test-results/bench/*.xml`. It's just a textual output. We can improve on this in future.

core/build.gradle
```diff
@@ -132,8 +132,13 @@ dependencies {
   integTestCompile "junit:junit:4.11"
   integTestCompile(project(":sparkling-water-extension-stack-trace"))
 
+  // Integration tests requirements
```
jakubhava: NIT: Wrong comment

core/build.gradle
```diff
@@ -182,4 +187,15 @@ integTest {
...
 
+bench {
+    systemProperty "spark.ext.h2o.backend.cluster.mode", detectBackendClusterMode()
+
+    systemProperty "spark.testing",   "true"
+    systemProperty "spark.test.home", "${sparkHome}"
+
+    // Pass list of jars required for testing
+    systemProperty "sparkling.assembly.jar", "${project(":sparkling-water-assembly").configurations.shadow.artifacts.file.join(',')}"
+    systemProperty "sparkling.bench.jar", "${benchJar.archivePath}"
```
jakubhava: The property `sparkling.bench.jar` is not used anywhere in the code, I think it can be removed.

Also if I'm correct, current bench test don't require `sparkling.assembly.jar` to be set

gradle/bench.gradle
```diff
@@ -0,0 +1,63 @@
...
+    appendix = 'bench'
+    from sourceSets.bench.output
+}
+
+// Run integration tests as part of build
+check.dependsOn bench
+
+// Dependencies
+bench.dependsOn benchJar
+bench.dependsOn ":sparkling-water-assembly:shadowJar"
```
jakubhava: bench tests are running locally so far, we don't need dependency on the assembly jar. It might be needed in future when we have some benchmarks running on YARN and we need to have the assembly jar build already

gradle/bench.gradle
```diff
@@ -0,0 +1,63 @@
...
+// Create jar containing bench classes
+task benchJar(type: Jar, dependsOn: benchClasses) {
+    group = "Build"
+    description = "Assembles a jar archive with benchmark classes."
+    baseName = "${project.name}_${scalaBaseVersion}"
+    appendix = 'bench'
+    from sourceSets.bench.output
+}
+
+// Run integration tests as part of build
```
jakubhava: comment referencing integration tests

PR #1307, opened 2019-06-28
[SW-1387] Handle vectors in SparkDataFrameConverter more explicitly

jakubhava:


core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -341,20 +344,22 @@ object H2OSchemaUtils {
...
       } else {
         val head = flatDataFrame.head()
-        val sparseOrNot = vectorIndices.map { idx =>
-          head.getAs[org.apache.spark.mllib.linalg.Vector](idx).isInstanceOf[SparseVector]
-        }
-        vectorIndices.zip(sparseOrNot).toMap
+        vectorIndices.map { idx =>
+          head.get(idx) match {
+            case _: mllib.linalg.SparseVector => (idx, true)
+            case _: ml.linalg.SparseVector => (idx, true)
```
jakubhava: We had bug here -> we were collecting sparse info only from mllib.linalg.Vector

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -109,26 +108,26 @@ object H2OSchemaUtils {
   }
 
   private def fillArray(
-      elementType: DataType,
-      flatSchemaIndexes: Map[String, Int],
-      buffer: ArrayBuffer[Any],
-      data: Any,
-      qualifiedName: String): Unit = {
+                         elementType: DataType,
```
mn-mikke: Do you prefer to use rather this formatting? The previous is based on coding conventions from Spark.  I am happy to use any type of formatting, but I would like to agree on which we will use in future.

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -109,26 +108,26 @@ object H2OSchemaUtils {
...
+                         flatSchemaIndexes: Map[String, Int],
+                         buffer: ArrayBuffer[Any],
+                         data: Any,
+                         qualifiedName: String): Unit = {
     val seq = data.asInstanceOf[Seq[Any]]
     val subRow = Row.fromSeq(seq)
     val fillBufferPartiallyApplied = fillBuffer(flatSchemaIndexes, buffer) _
-    (0 until seq.size).foreach { idx =>
+    seq.indices.foreach { idx =>
```
mn-mikke: Nice!

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -488,49 +483,40 @@ object H2OSchemaUtils {
...
-          case _: mllib.linalg.Vector =>
-            row.getAs[mllib.linalg.Vector](idx).size
-          case _: ml.linalg.Vector =>
-            row.getAs[ml.linalg.Vector](idx).size
-          case _ =>
-            throw new UnsupportedOperationException(s"User defined type is not supported: ${value.getClass}")
-        }
+      0
+    } else {
+      val dataType = row.schema.fields(idx).dataType
```
mn-mikke: Just a note: `schema` could be `null`

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: I always use the formatting by Idea after I click on `Reformat code` I don't have strong opinion on the formatting itself, but would like to be able to keep formatting using the Idea.

If there is any template idea can pick up and understand, I'm fine with that :)

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: Thanks IntelliJ :)

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: As we discussed on Slack, we are going to switch to formatting Spark is using, but in the following JIRA, thanks for bringing this up!

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
jakubhava: Good catch! Is it possible that row.schema is null even though whole dataFrame has schema defined?

PR #1310, opened 2019-06-28
[SW-1344] Unify ml package accross rel branches

jakubhava:
Another small step towards single branch


ml/src/main/java/org/apache/spark/ml/spark/models/svm/SparkSVM.java
```diff
@@ -216,8 +216,7 @@ public void computeImpl() {
                                 svm.run(training, vec2vec(_parms.initialWeights().vecs()));
                 training.unpersist(false);
 
-
-                sc.removeSparkListener(progressBar);
+                sc.listenerBus().removeListener(progressBar);
```
jakubhava: removeSparkListener is available only in 2.4 ( maybe 2.3 as well) but not older versions.

`sc.listenerBus().removeListener` is available in 2.4 - 2.1

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
```diff
@@ -149,8 +149,11 @@ class H2OMOJOPipelineModel(override val uid: String) extends H2OMOJOModelBase[H2
 
       // Transform the resulted Array of predictions into own but temporary columns
       // Temporary columns are created as we can't create the columns directly as nested ones
-      val predictionCols = $(outputCols).indices.map(idx => selectFromArray(idx)(frameWithPredictions.col(getPredictionCol() + ".preds")))
-      val frameWithExtractedPredictions = frameWithPredictions.withColumns(tempColNames, predictionCols)
+      var frameWithExtractedPredictions: DataFrame = frameWithPredictions
```
jakubhava: Replaced with code which is available across Spark 2.4 to 2.1

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
```diff
@@ -149,8 +149,11 @@ class H2OMOJOPipelineModel(override val uid: String) extends H2OMOJOModelBase[H2
...
+      $(outputCols).indices.foreach { idx =>
+        frameWithExtractedPredictions = frameWithExtractedPredictions.withColumn(tempColNames(idx),
```
mn-mikke: nit: can you try to format it into more lines to improve code readability?

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.json4s.jackson.JsonMethods._
+import org.json4s.{DefaultFormats, JObject}
+
+object SparkSpecificUtils extends CrossSparkUtils {
+
+  private def getAndSetParams(
+                               instance: Params,
+                               metadata: Metadata,
+                               skipParams: Option[List[String]] = None): Unit = {
+    implicit val format = DefaultFormats
```
mn-mikke: Can't we use this piece of code for 2.3 and 2.4 as well?

ml/src/main/spark_2.2/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.spark.ml.param.Params
+import org.apache.spark.ml.util.DefaultParamsReader
+import org.apache.spark.ml.util.DefaultParamsReader.Metadata
+import org.json4s.jackson.JsonMethods._
+import org.json4s.{DefaultFormats, JObject}
+
+object SparkSpecificUtils extends CrossSparkUtils {
+
+  private def getAndSetParams(
+                               instance: Params,
```
mn-mikke: Is it formatted with scalafmt?

jakubhava: No, formatting with Idea default configuration -> I think that we agreed to format using the spark way after we create the configuration file and put it in a repo or did I miss anything?

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
jakubhava: I was thinking about that, but that would complicate source directory definitions. We would need to have multiple CrossSpark traits and source directories named like spark-2.1-2.3. Doable, but not sure it it brings any real benefit compared to complexity it brings

ml/src/main/spark_2.2/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
mn-mikke: Spark use scalafmt, they use a custom configuration of it to be mere restrictive (number chars on lines etc.) I think we can use scalafmt defaults for now.

jakubhava: @mn-mikke oki, please create PR and let's start applying after that PR is in master

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
jakubhava: If you have any idea how to make this better, please let me know :)

ml/src/main/spark_2.2/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
mn-mikke: :-) yep, no problem

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
mn-mikke: Maybe we should leave it as it's for now. I don't know the whole context.

ml/src/main/spark_2.2/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
jakubhava: Thanks! I'm all for consistent code, but it will be easier to enforce once we have the template in the code https://0xdata.atlassian.net/browse/SW-1391

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
jakubhava: I will still think about it, I don't like code duplication, but was having hard time creating nice solution. Maybe we can brainstorm a bit tomorrow? :)

ml/build.gradle
```diff
@@ -3,14 +3,6 @@ description = "Sparkling Water ML Pipelines"
 apply from: "$rootDir/gradle/sparkTest.gradle"
 apply from: "$rootDir/gradle/utils.gradle"
 
-sourceSets {
-  main {
```
jakubhava: Actually, we don't need distinction on spark-version basis in ml package :) Thanks @mn-mikke !!

ml/src/main/spark_2.1/org/apache/spark/ml/h2o/SparkSpecificUtils.scala
jakubhava: Thank for today's discussion -> was able to simplify even further

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
mn-mikke: nit: Would it be possible to avoid using `var` variables? Maybe via `foldLeft`?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
```diff
@@ -18,14 +18,37 @@
...
+import org.json4s.JsonAST.JObject
+import org.json4s.jackson.JsonMethods.{compact, render}
+import org.json4s.{DefaultFormats, JsonAST}
 
 private[models] class H2OMOJOReader[T <: HasMojoData] extends DefaultParamsReader[T] {
 
+  private def getAndSetParams(
+                               instance: Params,
+                               metadata: Metadata,
+                               skipParams: Option[List[String]] = None): Unit = {
```
mn-mikke: nit: Do we need an option here? Wouldn't `Nil` play the same role as `None`?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
```diff
@@ -18,14 +18,37 @@
...
+    implicit val format = DefaultFormats
+    metadata.params match {
+      case JObject(pairs) =>
+        pairs.foreach { case (paramName, jsonValue) =>
```
mn-mikke: nit: Wouldn't be a solution with using the `filter` high-order function more compact?

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
jakubhava: Thanks, will do these improvements in the next PR

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
jakubhava: DTTO

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOReader.scala
jakubhava: DTTO

ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala
jakubhava: Good one!

PR #1313, opened 2019-07-01
[SW-1390] Fix NPE when reading modelDetails in Mojo

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -185,12 +185,16 @@ object H2OMOJOModel extends H2OMOJOReadable[PyH2OMOJOModel] with H2OMOJOLoader[P
     val reader = MojoReaderBackendFactory.createReaderBackend(is, MojoReaderBackendFactory.CachingStrategy.MEMORY)
 
     val modelOutputJson = JsonModelDescriptorReader.parseModelJson(reader).getAsJsonObject("output")
```
mn-mikke: Just double-checking `null` can be only returned from `getAsJsonObject`? and not by `parseModelJson`?

jakubhava: Good question! Double checked and if the file can't be read properly it throws out an exception. 

We already discovered with @Pscheidl that the file is created but some sections may be missing.

PR #1314, opened 2019-07-01
[SW-1386][FOLLOWUP] Replace df.isEmpty with df.take(1).headOption

mn-mikke:
This PR should fix the build of rel-2.1, rel-2.2 and rel-2.3.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
@@ -341,20 +341,17 @@ object H2OSchemaUtils {
     */
   def collectSparseInfo(flatDataFrame: DataFrame, elemMaxSizes: Array[Int]): Array[Boolean] = {
     val vectorIndices = collectVectorLikeTypes(flatDataFrame.schema)
-
-    val sparseInfoForVec = {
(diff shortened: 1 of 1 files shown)


mn-mikke:
@jakubhava Merging to fix the build of failing branches.

core/src/main/scala/org/apache/spark/h2o/utils/H2OSchemaUtils.scala
```diff
@@ -341,21 +341,18 @@ object H2OSchemaUtils {
     */
   def collectSparseInfo(flatDataFrame: DataFrame, elemMaxSizes: Array[Int]): Array[Boolean] = {
     val vectorIndices = collectVectorLikeTypes(flatDataFrame.schema)
-
-    val sparseInfoForVec = {
-      if (flatDataFrame.isEmpty) {
```
mn-mikke: @jakubhava The main reason for creating this PR is that [`isEmpty`](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/Dataset.scala#L559) was introduced in Spark 2.4. Older versions doesn't support the method.

jakubhava: oki, thanks. One would think this method was there from beginning of dataframes :)

PR #1348, opened 2019-07-16
[SW-1421] Integrate H2O 3.26.0.1

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOModel.scala
```diff
@@ -175,7 +175,8 @@ object H2OMOJOModel extends H2OMOJOReadable[PyH2OMOJOModel] with H2OMOJOLoader[P
     val is = new ByteArrayInputStream(mojoData)
     val reader = MojoReaderBackendFactory.createReaderBackend(is, MojoReaderBackendFactory.CachingStrategy.MEMORY)
 
-    val modelOutputJson = JsonModelDescriptorReader.parseModelJson(reader).getAsJsonObject("output")
+
+    val modelOutputJson = ModelJsonReader.parseModelJson(reader).getAsJsonObject("output")
```
jakubhava: was moved in H2O to ModelJsonReader

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -129,62 +122,51 @@ private[spark] trait H2OContextUtils extends Logging {
...
-        val ze = new ZipEntry(filename)
-        zos.putNextEntry(ze)
-        zos.write(result)
-        zos.closeEntry()
-      }
+  /**
+    * @param destination directory where the logs will be downloaded
+    */
+  def downloadH2OLogs(destination: URI, logContainer: LogArchiveContainer): URI = {
+    val workersLogs = getLogsFromWorkers(logContainer)
```
jakubhava: I created PR on H2O https://github.com/h2oai/h2o-3/pull/3698 which would allow us to remove the logic for downloading logs from SW and use directly H2O implementation. Right now this is duplication code in H2O

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -129,62 +122,51 @@ private[spark] trait H2OContextUtils extends Logging {
...
-        zos.write(clientResult)
-        zos.closeEntry()
-      }
-      // Close the top-level directory.
-      zos.closeEntry()
-    } finally {
-      // Close the full zip file.
-      zos.close()
+    val finalArchiveByteArray = try {
+      val method = classOf[RequestServer].getDeclaredMethod("archiveLogs", classOf[LogArchiveContainer],
```
mn-mikke: It would be nice to change the `archiveLogs` method to public and call the method directly in a different PR.

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -129,62 +122,51 @@ private[spark] trait H2OContextUtils extends Logging {
...
     import java.io.FileOutputStream
-    val outputStream = new FileOutputStream(destination.toString)
+    val destinationFile = new File(destination.toString,  outputFileStem + "." + logContainer.getFileExtension)
+    val outputStream = new FileOutputStream(destinationFile)
     try {
-      baos.writeTo(outputStream)
+      outputStream.write(finalArchiveByteArray)
     }
     finally {
       if (outputStream != null) outputStream.close()
```
mn-mikke: I know that this is unrelated to this PR, but can `outputStream`ever be `null`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Yup -> https://github.com/h2oai/h2o-3/pull/3698

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Good catch, we can remove this check!

PR #1354, opened 2019-07-17
[SW-1425] Implement H2OTargetEncoder for Python API

mn-mikke:


py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
```diff
@@ -0,0 +1,108 @@
...
+        self.setParams(**kwargs)
+
+
+    @keyword_only
+    def setParams(self, foldCol=None, labelCol="label", inputCols=[], holdoutStrategy = "None",
+                  blendedAvgEnabled=False, blendedAvgInflectionPoint=10.0, blendedAvgSmoothing=20.0, noise=0.01, noiseSeed=-1):
+        kwargs = get_input_kwargs(self)
+
+        if "holdoutStrategy" in kwargs:
+            kwargs["holdoutStrategy"] = getValidatedEnumValue(self.__getHoldoutStrategyEnumName(), kwargs["holdoutStrategy"])
```
jakubhava: Just created helper method for this in master.

Just need to do `validatedEnumValue(self._H2OTargetEncoderParams__getHoldoutStrategyEnumName(), kwargs, "holdoutStrategy"`

jakubhava:
@mn-mikke Thanks for creating the new structure for PySparkling!

We also need to ensure `ai` package is bundled into the distribution. 

In `build.gradle` we need to "add" it -> https://github.com/h2oai/sparkling-water/blob/f2bdeb32a229c6c394e398885c98f4c153aaa096/py/build.gradle#L151

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
jakubhava: For reference: https://stackoverflow.com/questions/70528/why-are-pythons-private-methods-not-actually-private

Calling python private method from "outside" -> that is why the `_H2OTargetEncoderParams__` prefix

mn-mikke: Should I move the `__getHoldoutStrategyEnumName` method if to `H2OTargetEncoderParams` if setters are only in `TargetEncoder`?

jakubhava: I probably don't understand :(.

mn-mikke: Setters are defined in `TargetEncoders` whereas other algos have them in `...Params`. I made update, please check whether it's ok?

jakubhava: I see this is fine, thanks! :) Just question -> why are setters and getters separated? No need to change, maybe I'm just missing some benefit of that approach

mn-mikke: `H2OTargetEncoder`, `H2OTargetEncoderModel` and `H2OTargetEncoderMOJOModel` inherit from `H2OTargetEncoderParams`. Since parameters on models are read-only, setters are defined on `H2OTargetEncoder` and not on `H2OTargetEncoderParams`. If I'm not mistaken, Spark estimators use a similar design.

jakubhava: great, thanks!

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
```diff
@@ -0,0 +1,107 @@
...
+from pysparkling.context import H2OContext
+from pysparkling.spark_specifics import get_input_kwargs
+from ai.h2o.sparkling.ml.params import H2OTargetEncoderParams
+from py_sparkling.ml.util import set_double_values, JavaH2OMLReadable
+from ai.h2o.sparkling.ml.models import H2OTargetEncoderModel
+
+from h2o.utils.typechecks import assert_is_type, Enum
+
+
+class H2OTargetEncoder(H2OTargetEncoderParams, JavaEstimator, JavaH2OMLReadable, JavaMLWritable):
```
jakubhava: We should use `JavaMLReadable` instead `JavaMLReadable`. JavaH2OMLReadable was introduced because of the py_sparkling hack.

Also we need a test which serializes & deserializes both unfitted pipeline and model

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @keyword_only
+    def __init__(self, foldCol=None, labelCol="label", inputCols=[], holdoutStrategy = "None",
+                 blendedAvgEnabled=False, blendedAvgInflectionPoint=10.0, blendedAvgSmoothing=20.0, noise=0.01, noiseSeed=-1):
+        super(H2OTargetEncoder, self).__init__()
+        self._hc = H2OContext.getOrCreate(SparkSession.builder.getOrCreate(), verbose=False)
+        self._java_obj = self._new_java_obj("ai.h2o.sparkling.ml.features.H2OTargetEncoder", self.uid)
```
jakubhava: Just FYI: When working on kmeans, I had a problem with putting just kmeans into a single file. It occured during loading the serialized pipeline because it thought the name of the file itself is part of the java package.

mn-mikke: I will cover that with tests. Thanks a lot!

jakubhava: For now I'm using the shared file

mn-mikke: Do you use the new package?

jakubhava: Yup. But let's say my file was at `at/h2o/sparking/algos/kmeans.py`

Then I guess the readable tried to read it from scala as `ai.h2o.sparkling.algos.kmeans.H2OKmeans`

I have idea how to fix that using custom JavaMLReadable, but would like to focus on this after release as this change would not affect users

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
mn-mikke: Changed `JavaH2OMLReadable` to `JavaMLReadable` and also added a test covering serde. You can check [here](https://github.com/h2oai/sparkling-water/pull/1354/files#diff-084cb88c118d3617b8cbe7299394efadR70)

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
mn-mikke: Hope my [test](https://github.com/h2oai/sparkling-water/pull/1354/files#diff-084cb88c118d3617b8cbe7299394efadR70) covers the scenario properly. It seems it works for both cases with `JavaH2OMLReadable` and `JavaMLReadable`.

Have you tried to add:
```
from ai.h2o.sparkling.algos.kmeans  import H2OKMeans
```
to `ai/h2o/sparking/algos/__init__.py`?

jakubhava: Let's have a all on Monday

jakubhava: The import did the trick as the package name is the one Spark expects now. Yup, the test is good!

Do we also have a test where we serialize and de serialize the fitted TE Model?

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
jakubhava: Thanks! After we switch all algos to ai.h2o, JavaH2OMLReadable is planned to be removed

py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
jakubhava: The import did the trick as the package name is the one Spark expects now. Yup, the test is good!

Do we also have a test where we serialize and de serialize the fitted TE Model?

mn-mikke: The below tests should cover various aspects of serde. Are they ok?
- [a test testing prediction abilities](https://github.com/h2oai/sparkling-water/pull/1354/files#diff-084cb88c118d3617b8cbe7299394efadR119)
- [propagation of parameters to mojo model] - [here](https://github.com/h2oai/sparkling-water/pull/1354/files#diff-084cb88c118d3617b8cbe7299394efadR57) and [here](https://github.com/h2oai/sparkling-water/pull/1354/files#diff-084cb88c118d3617b8cbe7299394efadR88).

jakubhava: Perfect, thanks!!

PR #1359, opened 2019-07-22
[SW-1439] [Prototype] Switch to using String value on the setters & getters in the ml API ( just on distribution param so far)

jakubhava:
So far just on single parameter, just as a prototype. Once we have approved solution, we can apply the deprecation to rest of the code.

The benefit is simpler usage of API( no need to reference different namespaces) + simple code on both, scala + python side -> Which can lead to simpler code generation once we start generation these APIs for Python & R

### DIFF
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
@@ -145,7 +145,7 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
     }
   }
 
-  // Expects that checkAllowedEnumValues was called before this method
+  // Expects that checkAllowedEnumValues method was called before this method
   protected def getCorrectEnumCase[T <: Enum[T]](name: String)
                                                 (implicit ctag: reflect.ClassTag[T]): String = {
     val names = ctag.runtimeClass.getDeclaredMethod("values").invoke(null).asInstanceOf[Array[T]].map(_.name())
--- a/ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
+++ b/ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
@@ -130,31 +130,28 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
     new NullableStringArrayParam(this, name, getDoc(doc, name))
   }
 
-  protected def checkAllowedEnumValues[T <: Enum[T]](name: String, nullAllowed: Boolean = false)
-                                                    (implicit ctag: reflect.ClassTag[T]): Unit = {
+  def getValidatedEnumValue[T <: Enum[T]](name: String, nullAllowed: Boolean = false)
+                                         (implicit ctag: reflect.ClassTag[T]): String = {
+    H2OAlgoParamsHelper.getValidatedEnumValue(name, nullAllowed)
+  }
+}
+
+object H2OAlgoParamsHelper {
+  def getValidatedEnumValue[T <: Enum[T]](name: String, nullAllowed: Boolean = false)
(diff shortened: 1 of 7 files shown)


ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
```diff
@@ -133,4 +129,13 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
   protected def nullableStringArrayParam(name: String, doc: Option[String] = None): NullableStringArrayParam = {
     new NullableStringArrayParam(this, name, getDoc(doc, name))
   }
+
+  protected def checkAllowedEnumValues[T <: Enum[T]](name: String, nullAllowed: Boolean = false)
+                                                    (implicit ctag: reflect.ClassTag[T]): Unit = {
+    val names = ctag.runtimeClass.getDeclaredMethod("values").invoke(null).asInstanceOf[Array[T]].map(_.name())
+    if (name == null && !nullAllowed || !names.contains(name)) {
```
mn-mikke: Wouldn't be more convenient to be case insensitive?

jakubhava: Good point, we had that on Python side as well, will apply that on Scala side as well

jakubhava: Hm, more I think about it, it seems better to me to be strict about this. I would actually prefer to be case senstive.

jakubhava:
@mn-mikke made the code case insensitive, please have a look

py/pysparkling/ml/params.py
```diff
@@ -166,10 +165,7 @@ def setParallelizeCrossValidation(self, value):
         return self._set(parallelizeCrossValidation=value)
 
     def setDistribution(self, value):
-        assert_is_type(value, None, Enum("AUTO", "bernoulli", "quasibinomial", "modified_huber", "multinomial", "ordinal", "gaussian", "poisson", "gamma", "tweedie", "huber", "laplace", "quantile"))
```
jakubhava: I wanted to reuse method defined on the Scala as defining the fields from enum manually is error prone. Plus created tests for the method validating enum values on Scala side.

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
```diff
@@ -133,4 +129,37 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
...
+      throw new IllegalArgumentException(s"Null is not a valid value. Allowed values are: ${names.mkString(", ")}")
+    } else if (nullAllowed && name == null) {
+      return null
+    }
+
+    if (!names.map(_.toLowerCase()).contains(name.toLowerCase())) {
+      val nullStr = if (nullAllowed) "null or " else ""
+      throw new IllegalArgumentException(s"'$name' is not a valid value. Allowed values are: $nullStr${names.mkString(", ")}")
+    }
+    names.find(_.toLowerCase() == name.toLowerCase).get
```
mn-mikke: For the `Distribution` enum, it seems to be fine. But what enums where values are represented several words. E.g. `someEnumValue`. Wouldn't a getter of a given parameter return `someenumvalue`?

Shouldn't this method serve only for checking the input value and a setter would store the original value?

jakubhava: I think the method behaves as you mentioned:

The `find` just finds the elements based on the conditions, but still contains the original elements

mn-mikke: Ok, thanks for your explanation!

jakubhava:
Thank you @mn-mikke!

PR #1363, opened 2019-07-23
[SW-1444] Use string representations instead of enums on Pipeline API

jakubhava:


ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
```diff
@@ -73,14 +74,14 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
     spec.input_spec.weights_column = getWeightCol()
     spec.input_spec.ignored_columns = getIgnoredCols()
     spec.input_spec.sort_metric = getSortMetric()
-    spec.build_models.exclude_algos = if (getExcludeAlgos() == null) null else Array(getExcludeAlgos(): _*)
-    spec.build_models.include_algos = if (getIncludeAlgos() == null) null else Array(getIncludeAlgos(): _*)
+    spec.build_models.exclude_algos = if (getExcludeAlgos() == null) null else Array(getExcludeAlgos().map(Algo.valueOf): _*)
```
mn-mikke: Do we need to support `null` values for `excludeAlgos` and `includeAlgos`. Wouldn't empty arrays play the same role?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
```diff
@@ -390,14 +392,14 @@ trait H2OGridSearchParams extends H2OCommonParams with Params {
...
+  private val strategy = new Param[String](this, "strategy", "Search criteria strategy")
   private val maxRuntimeSecs = new DoubleParam(this, "maxRuntimeSecs", "maxRuntimeSecs")
   private val maxModels = new IntParam(this, "maxModels", "maxModels")
   private val stoppingRounds = new IntParam(this, "stoppingRounds", "Early stopping based on convergence of stoppingMetric")
   private val stoppingTolerance = new DoubleParam(this, "stoppingTolerance", "Relative tolerance for metric-based" +
     " stopping criterion: stop if relative improvement is not at least this much.")
-  private val stoppingMetric = new StoppingMetricParam(this, "stoppingMetric", "Stopping Metric")
-  private val selectBestModelBy = new MetricParam(this, "selectBestModelBy", "Select best model by specific metric." +
+  private val stoppingMetric = new Param[String](this, "stoppingMetric", "Stopping Metric")
+  private val selectBestModelBy = new NullableStringParam(this, "selectBestModelBy", "Select best model by specific metric." +
```
mn-mikke: Wouldn't be `Param[String]` sufficient if the default value is `"AUTO"`?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
```diff
@@ -130,36 +132,64 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
...
-                                                   (implicit ctag: reflect.ClassTag[T]): String = {
-    H2OAlgoParamsHelper.getValidatedEnumValue(name, nullAllowed)
-  }
 }
 
 object H2OAlgoParamsHelper {
-  def getValidatedEnumValue[T <: Enum[T]](name: String, nullAllowed: Boolean = false)
-                                         (implicit ctag: reflect.ClassTag[T]): String = {
-    getValidatedEnumValue(ctag.runtimeClass, name, nullAllowed)
+
```
mn-mikke: nit: extra line

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
jakubhava: Unfortunately yes, empty array is valid value for AutoML. If include_algos is set to empty array then no algos are selected.

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OGridSearch.scala
jakubhava: 👍 great catch!

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
mn-mikke: Ok, thanks!

jakubhava: I was playing with a thought of setting includeAlgos by default to all & excludeAlgos to empty array, but AutoML complains as these 2 options are mutually exclusive and we can't programmatically decide which to choose.


One solution to avoid null on the API which would be possible is to do the following:

By default, have:
```
includeAlgos = [ALL]
excludeAlgos = null
```

After calling setExcludeAlgos(), set includeAlgos to null, print out warning and set excludeAlgos.

The same for the opposite way. What do you think?

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
```diff
@@ -130,36 +132,63 @@ trait H2OAlgoParamsHelper[P <: Parameters] extends Params {
...
-      val nullStr = if (nullAllowed) "null or " else ""
-      throw new IllegalArgumentException(s"'$name' is not a valid value. Allowed values are: $nullStr${names.mkString(", ")}")
+      throw new IllegalArgumentException(s"'$name' is not a valid value. Allowed values are: ${names.mkString(", ")}")
     }
     names.find(_.toLowerCase() == name.toLowerCase).get
   }
 
+  private def getValidatedEnumValues(clazz: Class[_], inputNames: Array[String], nullEnabled: Boolean): Array[String] = {
+    val names = getEnumValues(clazz)
+    if (nullEnabled && inputNames == null) {
```
mn-mikke: nit: WDYT about nested `if` statements here?

ml/src/main/scala/org/apache/spark/ml/h2o/algos/H2OAutoML.scala
mn-mikke: I think we can leave it as it is :-)

jakubhava: Cool, I will at least create a JIRA for it, so we can think of it later in scope of a smaller change

ml/src/main/scala/org/apache/spark/ml/h2o/param/H2OAlgoParamsHelper.scala
jakubhava: yes!

PR #1370, opened 2019-07-24
[SW-1454] It is no longer possible to specify predictionCol :(

jakubhava:


py/py_sparkling/ml/models.py
```diff
@@ -11,18 +11,22 @@
 class H2OMOJOSettings(JavaWrapper):
 
     def __init__(self,
+                 predictionCol = "prediction",
```
jakubhava: Todo: get default values from scala side so they are not duplicated

PR #1375, opened 2019-07-25
[SW-1460] Infrastructure for prediction column with a simple prediction value

jakubhava:
The goal is to have single prediction value in prediction column.

As we are introducing new algo -> H2OKmeans, I wanted to have this algorithm implemented with the new approach

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
@@ -26,8 +26,10 @@ import org.apache.spark.ml.param._
 trait H2OCommonParams extends Params with Logging {
 
   protected final val predictionCol = new Param[String](this, "predictionCol", "Prediction column name")
-  protected final val detailedPredictionCol = new Param[String](this, "detailedPredictionCol", "Column containing additional prediction details, its content depends on the model type.")
-  protected final val withDetailedPredictionCol = new BooleanParam(this, "withDetailedPredictionCol", "Enables or disables generating additional prediction column, but with more details")
+  protected final val detailedPredictionCol = new Param[String](this, "detailedPredictionCol",
+    "Column containing additional prediction details, its content depends on the model type.")
+  protected final val withDetailedPredictionCol = new BooleanParam(this, "withDetailedPredictionCol",
+    "Enables or disables generating additional prediction column, but with more details")
   protected final val featuresCols = new StringArrayParam(this, "featuresCols", "Name of feature columns")
   protected final val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
   protected final val weightCol = new NullableStringParam(this, "weightCol", "Weight column name")
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
@@ -26,6 +26,8 @@ import org.apache.spark.ml.param._
 trait H2OCommonParams extends Params with Logging {
 
   protected final val predictionCol = new Param[String](this, "predictionCol", "Prediction column name")
+  protected final val detailedPredictionCol = new Param[String](this, "detailedPredictionCol", "Column containing additional prediction details, its content depends on the model type.")
+  protected final val withDetailedPredictionCol = new BooleanParam(this, "withDetailedPredictionCol", "Enables or disables generating additional prediction column, but with more details")
   protected final val featuresCols = new StringArrayParam(this, "featuresCols", "Name of feature columns")
   protected final val foldCol = new NullableStringParam(this, "foldCol", "Fold column name")
   protected final val weightCol = new NullableStringParam(this, "weightCol", "Weight column name")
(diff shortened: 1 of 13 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
```diff
@@ -58,6 +60,8 @@ trait H2OCommonParams extends Params with Logging {
   //
   setDefault(
     predictionCol -> "prediction",
+    detailedPredictionCol -> "detailed_prediction",
+    withDetailedPredictionCol -> false,
```
mn-mikke: Could you tell me a benefit of disabling `detailedPredictionCol`? Is it heavy to compute?

py/py_sparkling/ml/algos.py
```diff
@@ -132,7 +150,10 @@ def __init__(self, featuresCols=[], labelCol="label", allStringColumnsToCategori
                  stoppingTolerance=0.001, stoppingMetric="AUTO", nfolds=5, convertUnknownCategoricalLevelsToNa=True, seed=-1,
                  sortMetric="AUTO", balanceClasses=False, classSamplingFactors=None, maxAfterBalanceSize=5.0,
                  keepCrossValidationPredictions=True, keepCrossValidationModels=True, maxModels=0,
-                 predictionCol="prediction", **deprecatedArgs):
+                 predictionCol="prediction",
```
mn-mikke: nit: Shouldn't the new parameter use the same formatting as previous parameters?

py/pysparkling/ml/params.py
```diff
@@ -35,6 +39,9 @@ class H2OCommonParams(Params):
     def getPredictionCol(self):
         return self.getOrDefault(self.predictionCol)
 
+    def getDetailedPredictionCol(self):
+        return self.getOrDefault(self.detailedPredictionCol)
+
```
mn-mikke: What about `getWithDetailedPredictionCol`?

py/pysparkling/ml/params.py
```diff
@@ -69,6 +76,10 @@ def setPredictionCol(self, value):
         assert_is_type(value, str)
         return self._set(predictionCol=value)
 
+    def setDetailedPredictionCol(self, value):
```
mn-mikke: What about `setWithDetailedPredictionCol`?

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
jakubhava: It works in a way that we always get full prediction (with all details) and select just the most interesting information and put that into `prediction` column.

If this is set to false, we remove the column with detailed predictions. I would like to avoid confusing users with 2 new colums for predictions. This way always the `prediction` is generated, and if the option is set to true, the `prediction_details` is kept as well.

See the transform logic in H2OMOjOModel

py/py_sparkling/ml/algos.py
jakubhava: I found myself really hard to read the options as they are now, wanted to switch to different formatting, one option per line, but would like to do it when we have more time

py/pysparkling/ml/params.py
jakubhava: perfect catch!

py/py_sparkling/ml/algos.py
mn-mikke: ok

py/pysparkling/ml/params.py
jakubhava: great catch!

py/pysparkling/ml/params.py
jakubhava: fixed, thanks!

py/py_sparkling/ml/algos.py
jakubhava: But if you prefer the consistency, I can switch it back, no worries :)

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
mn-mikke: And couldn't we generate `prediction_details `  all the time? It would be upto user whether he/she wants read the value or not.

jakubhava: My opinion is that I would like to optimize the case for the basic user -> not adding unnecessary column for them if they are not interested in that. Advanced users can ask for it.

mn-mikke: ok, no problem

py/py_sparkling/ml/algos.py
mn-mikke: I would prefer to be consistent and change formatting of all parameters in a different PR. But this a really minor thing. If you leave it as it's, it works for me as well.

jakubhava: No worries, valid point, will keep the current formatting :)

jakubhava: Fixed, thx

jakubhava:
Let me know if looks good know, comments should be addressed :)

PR #1379, opened 2019-07-25
[SW-1456] Documentation of Target Encoder

mn-mikke:


mn-mikke:
@jakubhava I've updated the description of edge cases. Can you have a second look?

PR #1382, opened 2019-07-27
[SW-1207][merge-2] Implement H2OTargetEncoder for Scala API

mn-mikke:
This PR merges h2oai/sparkling-water#1192 again and waits for the upgrade to 3.26.0.2.

Reverts h2oai/sparkling-water#1380


jakubhava:
Thanks @mn-mikke , just integrated with 3.26.0.2, merging this change as well

PR #1390, opened 2019-07-30
[SW-1471] [PySparkling] Minor bugs on TE

jakubhava:


py/ai/h2o/sparkling/ml/features/H2OTargetEncoder.py
```diff
@@ -92,15 +92,15 @@ def setBlendedAvgEnabled(self, value):
 
     def setBlendedAvgInflectionPoint(self, value):
         assert_is_type(value, int, float)
-        return self._set(blendedAvgInflectionPoint=value)
+        return self._set(blendedAvgInflectionPoint=float(value))
```
jakubhava: We need to explicitly convert the value to float even in setters, for the same reason as we do in constructor

py/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.py
```diff
@@ -56,7 +56,7 @@ def getInputCols(self):
         return self.getOrDefault(self.inputCols)
 
     def getOutputCols(self):
-        return map(lambda c: c + "_te", self.getInputCols())
+        return list(map(lambda c: c + "_te", self.getInputCols()))
```
jakubhava: on Python3 this returns object, we need to pass it into list method, which returns list on python2 and python3

PR #1395, opened 2019-07-31
[SW-1476] Avoid duplication between mojo params and algo params

jakubhava:
As new parameters are being exposed on MOJO Model, the goal of this change is to minimize the code which needs to be changed in order to do so.


ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OMOJOAlgoSharedParams.scala
```diff
@@ -47,21 +47,18 @@ trait H2OMOJOModelParams extends Params with Logging {
...
-    convertUnknownCategoricalLevelsToNa -> H2OMOJOSettings.default.convertUnknownCategoricalLevelsToNa,
-    convertInvalidNumbersToNa -> H2OMOJOSettings.default.convertInvalidNumbersToNa,
-    namedMojoOutputColumns -> H2OMOJOSettings.default.namedMojoOutputColumns,
     detailedPredictionCol -> H2OMOJOSettings.default.detailedPredictionCol,
     withDetailedPredictionCol -> H2OMOJOSettings.default.withDetailedPredictionCol,
-    modelDetails -> null
+    featuresCols -> Array.empty[String],
+    convertUnknownCategoricalLevelsToNa -> H2OMOJOSettings.default.convertUnknownCategoricalLevelsToNa,
+    convertInvalidNumbersToNa -> H2OMOJOSettings.default.convertInvalidNumbersToNa,
+    namedMojoOutputColumns -> H2OMOJOSettings.default.namedMojoOutputColumns
```
mn-mikke: Is there any relationship between `namedMojoOutputColumns` and `predictionCol`and `predictionDetailsCol`?

jakubhava: no

mn-mikke: May i know what's their purpose?

jakubhava: predictionCol and predictionDetailsCols are obvious I guess.

for `namedMojoOutputColumns` -> used only on MojoPipeline, when enabled, the outputs are named by the label cols, instead of returning regular array. We will get rid of this option after we have simple output in predictionCol.
https://github.com/h2oai/sparkling-water/blob/610332aa421f9b20f38bf9a582d915cc5aa689b8/ml/src/main/scala/org/apache/spark/ml/h2o/models/H2OMOJOPipelineModel.scala#L128

Though changing semantic of these is not goal of this change

mn-mikke: Thanks for the explanation!

jakubhava: no worries

jakubhava:
Thank you @mn-mikke !

PR #1398, opened 2019-07-31
[SW-1478] Don't need to start H2O to initialize algo on PySparkling side

jakubhava:


py/ai/h2o/sparkling/ml/algos/H2OKmeans.py
```diff
@@ -59,7 +58,7 @@ def setStandardize(self, value):
         return self._set(standardize=value)
 
     def setInit(self, value):
-        validated = getValidateEnumValue(self.__getInitEnum(), value)
+        validated = getValidatedEnumValue(self.__getInitEnum(), value)
```
jakubhava: Bug fix here

PR #1399, opened 2019-07-31
[SW-1368][Step-2] Preparation for MOJO scoring package

jakubhava:


mn-mikke:
Please can share details why these classes are taken from the `sparkling-water-ml` package? What's supposed to go to the `ml` package and what to the `scoring` package from now?

jakubhava:
Good question! All classes which needs to be available for performing predictions on MOJO need to be in the mojo scoring package. The MOJO scoring package does not require dependency in H2O run-time.

PR #1406, opened 2019-08-01
[SW-1483]  Deprecate algos and features in org.apache.spark package

jakubhava:
Last in this row of bigger, deprecation changes.

It's quite a big change, I could create PR for moving each algo separately if this would be hard to review


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -0,0 +1,278 @@
+/*
```
jakubhava: All these files are really just refactored -> moved from one dir to another

PR #1410, opened 2019-08-02
[SW-1483][Step-2] Deprecate estimators in org.apache.spark package.

jakubhava:
This is first step to deprecate algos in old package. Split the original PR into multiple ones to have better history as we discussed.

Hopefully should be easier to review


jakubhava:
Yup, we should, thanks for spotting that!

jakubhava:
Fixed. Please have a look

PR #1412, opened 2019-08-02
[SW-1293] Benchmarks: Test Cases

mn-mikke:


mn-mikke:
> It would be great if we will run the benchmarks as part of, for example, nightly sparkling water jobs and archive the benchmarks results as jenkins artifact

Sounds good to me.

Tagar:
Would this profiling be exposed to users? 
So we could see for example where most of the time is spent during model building 
to better understand how to tune some of model parameters if we're concerned about 
model build time.. 
Some tree-based models can take an hour to train on our datasets, in a good cluster,
making such algorithms a no-go just based on train time. 
Would be nice to see some details where time is spent.
Thank you.

PR #1416, opened 2019-08-05
[SW-1488] Benchmarks: Use persist(StorageLevel.DISK_ONLY) for Materialization of Parsed DataFrames

mn-mikke:


mn-mikke:
@jakubhava I had to introduce a custom `localCheckpoint()` since the original one was introduced in Spark 2.3. The solution is based on https://github.com/apache/spark/commit/13268a58f8f67bf994f0ad5076419774c45daeeb.

mn-mikke:
Can you have a second look?

mn-mikke:
@jakubhava I've changed the solution to use `persist(StorageLevel.DISK_ONLY)`, please take a look.

PR #1418, opened 2019-08-06
[SW-1489] Use TypeConverters on AutoML pysparkling wrapper

jakubhava:
Also fixes bug in H2OAutoML constructor. `ClassSamplingFactors` was not converted to double values

### DIFF
--- a/py/ai/h2o/sparkling/ml/algos/H2OAutoML.py
+++ b/py/ai/h2o/sparkling/ml/algos/H2OAutoML.py
@@ -25,7 +25,6 @@ from pyspark.sql.dataframe import DataFrame
 from ai.h2o.sparkling import Initializer
 from ai.h2o.sparkling.ml.models import H2OMOJOModel
 from ai.h2o.sparkling.ml.params import H2OAutoMLParams
-from ai.h2o.sparkling.ml.utils import set_double_values, validateEnumValue, validateEnumValues
 from ai.h2o.sparkling.sparkSpecifics import get_input_kwargs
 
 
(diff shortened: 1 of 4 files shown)


py/ai/h2o/sparkling/ml/params/H2OCommonParams.py
```diff
@@ -17,6 +17,7 @@
 
 from h2o.utils.typechecks import assert_is_type
 from pyspark.ml.param import *
+from pyspark.ml.param import TypeConverters
```
mn-mikke: Isn't this line extra if we have the line above?

py/ai/h2o/sparkling/ml/algos/H2OAutoML.py
```diff
@@ -64,18 +63,9 @@ def setParams(self, featuresCols=[], labelCol="label", allStringColumnsToCategor
 
         kwargs = get_input_kwargs(self)
 
-        validateEnumValues(self._H2OAutoMLParams__getAutomlAlgoEnum(), kwargs, "includeAlgos", nullEnabled=True)
-        validateEnumValues(self._H2OAutoMLParams__getAutomlAlgoEnum(), kwargs, "excludeAlgos", nullEnabled=True)
-        validateEnumValue(self._H2OAutoMLParams__getStoppingMetricEnum(), kwargs, "stoppingMetric")
-        validateEnumValue(self._H2OAutoMLParams__getSortMetricEnum(), kwargs, "sortMetric")
-
         if "projectName" in kwargs and kwargs["projectName"] is None:
```
mn-mikke: Shouldn't this be moved to parama spec as well?

jakubhava: This is a bit special -> we are generating random string on Scala part, but that is not propagated to Python. But please see https://github.com/h2oai/sparkling-water/pull/1419 where I should improve this part

py/ai/h2o/sparkling/ml/params/H2OCommonParams.py
jakubhava: Good catch, fixed

py/ai/h2o/sparkling/ml/algos/H2OAutoML.py
mn-mikke: After #1419 gets merged, should this be completely removed? Just wondering whether the constructor parameter and the setter will behave the same way.

jakubhava: Yup, my bad #1419 should delete that, will fix it on that PR side

PR #1419, opened 2019-08-06
[SW-1490] Immutable projectName on H2OAutoML

jakubhava:
This change has more actually multiple motivations:

- Immutability -> don't change user specified value on the estimator
- Keep all the logic in Scala wrapper, have PySparkling just as dummy client. 
- Keep the client simply and generic so we can later generate all algos with the same generator
- Move special methods which need to be on H2OAutoML to Extension class. The generator can check if the extension class exists and make sure the Algo, in this case H2OAutoML, will extend from it. The Extensions methods can't be generated so that is why I extracted them.

### DIFF
--- a/py/ai/h2o/sparkling/ml/algos/H2OAutoMLExtensions.py
+++ b/py/ai/h2o/sparkling/ml/algos/H2OAutoMLExtensions.py
@@ -1,0 +1,28 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


py/ai/h2o/sparkling/ml/algos/H2OAutoMLExtensions.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from pyspark.sql.dataframe import DataFrame
+
+
+class H2OAutoMLExtensions(object):
+
+    def leaderboard(self):
```
mn-mikke: What is the reason for getting it separated? Just asking.... I would like to know the context

jakubhava: This is done for the generation -> I wanted to extract the algo specific bits outside the algo definition to be able to generated all algos in the same way

PR #1421, opened 2019-08-07
[SW-1494] Upgrade Terraform Templates to AWS Provider 2.23

mn-mikke:


templates/src/terraform/aws/modules/emr/main.tf
```diff
@@ -127,16 +127,14 @@ resource "aws_emr_cluster" "sparkling-water-cluster" {
...
     name = "SparklingWater"
   }
 
-  bootstrap_action = [
-    {
-      path = "${format("s3://%s/install_sw.sh", aws_s3_bucket.sw_bucket.bucket)}"
-      name = "Custom action"
-    }
-  ]
+  bootstrap_action {
```
jakubhava: Have you tried the template with this changed syntax?

mn-mikke: Yup, I made these change to be able to deploy SW to EMR.

jakubhava: Perfect!

PR #1422, opened 2019-08-08
[SW-1499] Extract zip file into temp directory owned and configured by Spark

jakubhava:
This directory is also automatically deleted when the Spark JVM shuts down


jakubhava:
Good question! -> It is unique for each Spark run. ( also verified manually)

PR #1424, opened 2019-08-09
[SW-1503] Avoid null on AutoML include & exlude Algos params

jakubhava:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -107,6 +107,15 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
       modelSettings)
   }
 
+  private def determineIncludedAlgos(): Array[Algo] = {
+    val bothIncludedExcluded = getIncludeAlgos().filter(algo => getExcludeAlgos().contains(algo))
+    bothIncludedExcluded.foreach { algo =>
+      logWarning(s"Algorithm '$algo' was specified in both include and exclude parameters. " +
+        s"Excluding the algorithm.")
+    }
+    getExcludeAlgos().filter(algo => !bothIncludedExcluded.contains(algo)).map(Algo.valueOf)
```
mn-mikke: WDYT about `getExcludeAlgos().diff(bothIncludedExcluded).map(Algo.valueOf)`?

jakubhava:
Yup, by default all algorithms are used

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -107,6 +107,15 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
       modelSettings)
   }
 
+  private def determineIncludedAlgos(): Array[Algo] = {
+    val bothIncludedExcluded = getIncludeAlgos().filter(algo => getExcludeAlgos().contains(algo))
```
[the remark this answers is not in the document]
jakubhava: nice!

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
jakubhava: Good tip! I think it should be  `getIncludeAlgos().diff(bothIncludedExcluded).map(Algo.valueOf)` to get the right results

PR #1429, opened 2019-08-12
[SW-1502] Upgrade to mojo2 library v2.1.3

jakubhava:


jakubhava:
@mmalohlava locally verified that 2.1.3 works as expected, it is just not released on Maven yet

doc/src/site/sphinx/deployment/scoring_mojo_pipeline.rst
```diff
@@ -3,8 +3,6 @@ Using the MOJO Scoring Pipeline with Spark/Sparkling Water
 
 MOJO scoring pipeline artifacts can be used in Spark to deploy predictions in parallel using the Sparkling Water API. This section shows how to load and run predictions on the MOJO scoring pipeline in Spark using Scala and the Python API.
 
-In the event that you upgrade H2O Driverless AI, we have a good news! Sparkling Water is backwards compatible with MOJO versions produced by older Driverless AI versions.
```
mn-mikke: Does it mean that we are not backwards compatible with older MOJO versions anymore?

jenkins/sparklingWaterPipeline.groovy
```diff
@@ -244,13 +244,11 @@ def buildAndLint() {
...
         stage('QA: Build and Lint - ' + config.backendMode) {
             try {
-                withCredentials([usernamePassword(credentialsId: "LOCAL_NEXUS", usernameVariable: 'LOCAL_NEXUS_USERNAME', passwordVariable: 'LOCAL_NEXUS_PASSWORD')]) {
-                    sh """
-                        ${config.gradleCmd} clean build -x check scalaStyle -PlocalNexusUsername=$LOCAL_NEXUS_USERNAME -PlocalNexusPassword=$LOCAL_NEXUS_PASSWORD
-                       """
-                    if (config.runIntegTests.toBoolean()) {
-                        stash "sw-build-${config.sparkMajorVersion}"
-                    }
+                sh """
```
mn-mikke: Can we do it in one line?

ml/build.gradle
```diff
@@ -21,12 +21,11 @@ dependencies {
 
 
     compile "ai.h2o:mojo2-runtime-api:${mojoPipelineVersion}"
+    runtime "ai.h2o:mojo2-runtime-impl:${mojoPipelineVersion}"
```
mn-mikke: I was asked to test it. There is a breaking conflict between versions of the `com.google.protobuf` library. `hadoop-commons` (hadoop 2.7) utilizes version 2.5 and the new MOJO utilizes the version 3.7. It worked for me to make the dependency as `compile` and include and relocate the `com.google.protobuf` within the assembly jar.

doc/src/site/sphinx/deployment/scoring_mojo_pipeline.rst
jakubhava: We are still backwards compatible, I just want to change place where we mention it

ml/build.gradle
jakubhava: Can you share reproducible code please? It works on:

```
from pysparkling.ml import *
path = "/Users/kuba/devel/repos/sparkling-water/ml/src/test/resources/mojo2data/pipeline.mojo"
settings = H2OMOJOSettings(namedMojoOutputColumns=False)
mojo = H2OMOJOPipelineModel.createFromMojo(path, settings)

data = spark.read.csv("/Users/kuba/devel/repos/sparkling-water/examples/smalldata/prostate/prostate.csv", header=True)
mojo.transform(data).collect()
```

jenkins/sparklingWaterPipeline.groovy
jakubhava: Yup, done

ml/build.gradle
mn-mikke: Is this `pipeline.mojo` file produced by the latest version of DAI? I will send you the script that I tested on ;-)

jakubhava: I see! This pipeline is using the toml config file -> I guess this error occurs for proto based pipelines :) Will add test for it and apply your suggestion, thanks!

jakubhava:
Thanks @mn-mikke, I merged your changes go this PR & also ported MM's test in our tests. Passing now.

It also helped me discover a different bug -> when mojo expects String & we have parsed data to more specific types, the conversion fails. So in case the mojo expects string on specific column we now convert it to String

build.gradle
```diff
@@ -178,17 +178,7 @@ configure(subprojects) { project ->
         maven {
             url "http://h2o-release.s3.amazonaws.com/h2o/master/$h2oBuild/maven/repo/"
         }
-
```
jakubhava: No need anymore, was used for fetching mojo from private nexus

ml/build.gradle
jakubhava: note: protobuf will be shadowed in mojo library, but it won't solve the issue with exceeding limit for pypi upload as we still bundle the libraries into our fat jar

PR #1433, opened 2019-08-13
[SW-1501] Check the case of JAR being attached prior the start of PySparkling

jakubhava:


py/ai/h2o/sparkling/Initializer.py
```diff
@@ -106,11 +111,12 @@ def check_different_h2o():
             command_sys_path = "import sys; sys.path = " + str(path_without_sw).replace("'", "\"") + ";"
             command_import_h2o = "import h2o; print(h2o.__version__)"
             full_command = "python -c '" + command_sys_path + command_import_h2o + "'"
-            previous_version = subprocess.check_output(full_command, shell=True, stderr=DEVNULL).decode('utf-8').replace("\n", "")
+            previous_version = subprocess.check_output(full_command, shell=True, stderr=DEVNULL).decode(
```
mn-mikke: nit: Can  we format the piece of code in a better way? Maybe:
```
previous_version = subprocess
    .check_output(full_command, shell=True, stderr=DEVNULL)
    .decode('utf-8')
    .replace("\n", "")
```

py/ai/h2o/sparkling/Initializer.py
```diff
@@ -141,3 +147,13 @@ def __add_url_to_classloader(jvm, url):
             else:
                 logger.debug("Skipping classloader '{}'".format(cl.toString()))
             cl = cl.getParent()
+
+    @staticmethod
+    def getVersion():
+        here = path.abspath(path.dirname(__file__))
+        if '.zip' in here:
+            with zipfile.ZipFile(here[:-len("ai/h2o/sparkling/")], 'r') as archive:
+                return archive.read('ai/h2o/sparkling/version.txt').decode('utf-8').strip()
```
mn-mikke: Can happen that the version file is still in the `pysparkling` folder?

jakubhava: Good question, but not anymore, we are generating it into `ai/h2o/sparkling`

py/ai/h2o/sparkling/Initializer.py
jakubhava: 👍

jakubhava:
@mn-mikke we are not changing the order of dependencies. Spark is still loading the dependencies the same way as before. What we just changed is that we added verification of the Sparkling Water JAR. We check if the JAR is already attached to the cluster and if not, attach it. We have already been attaching the JAR before, but we ignored the fact that the JAR might have been already attached.

What will change for the user that they will see exception when they have conflicting jars, but that is a bug fix as before, they could hit weird issues. This way, we warn them early

PR #1437, opened 2019-08-14
[SW-1520] Fix warning 'H2OMOJOSettings' object has no attribute '_java_obj'

jakubhava:


jakubhava:
We can't do it because the default method is specified as default argument -> if we would init java_obj in init method then the code would expect that Sparkling Water JAR is already attached and Spark running during import, which does not have to be true in all cases

PR #1440, opened 2019-08-14
[SW-1523] Fix IP based cloud-up on client side

jakubhava:
As we don't need this configuration, I would like to remove it so it is easier for users start using Sparkling Water


jakubhava:
Thank you @mn-mikke !

PR #1441, opened 2019-08-14
[SW-1522] createTempDir in SharedBackendUtils should create temp directories in the Spark temporary directory

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
```diff
@@ -33,7 +34,7 @@ import scala.annotation.tailrec
   */
 private[spark]
 class SpreadRDDBuilder(@transient private val hc: H2OContext,
-                       numExecutorHint: Option[Int] = None) extends SharedBackendUtils {
+                       numExecutorHint: Option[Int] = None) extends Logging {
```
mn-mikke: Can you please help me understand why this has changed?

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -204,7 +192,6 @@ object SharedBackendUtils extends SharedBackendUtils {
 
   def saveFlatFileAsFile(content: String): File = {
     val tmpDir = createTempDir()
-    tmpDir.deleteOnExit()
```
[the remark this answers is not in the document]
jakubhava: Spark takes care of it

core/src/main/scala/org/apache/spark/h2o/backends/internal/SpreadRDDBuilder.scala
jakubhava: Unnecessary import

PR #1447, opened 2019-08-16
[SW-1527] Verify version between H2O external back-end & H2O client on Spark driver

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -383,6 +390,36 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
+    val clientH2OVersion = BuildInfo.H2OVersion
+
+    val jarFile = new JarFile(driverPath)
+    val entry = jarFile.getJarEntry("h2o.version")
+    val is = jarFile.getInputStream(entry)
+    val externalVersion = scala.io.Source.fromInputStream(is).mkString
+    jarFile.close()
+
+    if (clientH2OVersion != externalVersion) {
+      throw new RuntimeException(s"The driver '$driverPath' is for H2O of version $externalVersion but Sparkling Water " +
```
mn-mikke: nit: Can we use a multi-line string representation?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -383,6 +390,36 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
+    }
+  }
+
+  private def verifyVersionFromRuntime(): Unit = {
+    val clientH2OVersion = BuildInfo.H2OVersion
+    new MRTask() {
+      override def setupLocal(): Unit = {
+        val externalVersion = AbstractBuildVersion.getBuildVersion.projectVersion()
+        if (clientH2OVersion != externalVersion) {
+          throw new RuntimeException(s"The external H2O cluster is of version $externalVersion but Sparkling Water " +
```
mn-mikke: ditto

mn-mikke: Is the exception propagated back to the driver or does it cause just a failure of a node?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -383,6 +390,36 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
+        s"is using version of H2O $clientH2OVersion. Please make sure to use the corresponding extended H2O JAR.")
+    }
+  }
+
+  private def verifyVersionFromRuntime(): Unit = {
+    val clientH2OVersion = BuildInfo.H2OVersion
+    new MRTask() {
+      override def setupLocal(): Unit = {
+        val externalVersion = AbstractBuildVersion.getBuildVersion.projectVersion()
+        if (clientH2OVersion != externalVersion) {
```
mn-mikke: Maybe we could move the check with an exception to a separate method. it seems to be very similar to the check above.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: It is propagated to the driver, just verified

PR #1448, opened 2019-08-16
[SW-1526] PySparkling cannot parse embedded version.txt file

mn-mikke:


mn-mikke:
@jakubhava Mering to master in order to create a new nightly build. If you find any drawbacks, I will fix them in a follow up PR.

py/ai/h2o/sparkling/__init__.py
```diff
@@ -18,3 +18,4 @@
 from ai.h2o.sparkling.H2OConf import H2OConf
 from ai.h2o.sparkling.H2OContext import H2OContext
 from ai.h2o.sparkling.Initializer import Initializer
+from ai.h2o.sparkling.VersionComponents import VersionComponents
```
jakubhava: nl

py/ai/h2o/sparkling/VersionComponents.py
```diff
@@ -0,0 +1,32 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import re
+
+class VersionComponents(object):
+
+    @staticmethod
+    def parseFromVersion(version):
```
jakubhava: Would be good to have tests for this method to verify the the regexp for normal & nightly version

mn-mikke: I will add them, thanks!

PR #1451, opened 2019-08-19
[SW-1530] Conversion to H2OFrame needs to work without running H2O client

jakubhava:
First step is to use the init and finalize methods on the remote backend side

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
@@ -17,31 +17,22 @@
 
 package org.apache.spark.h2o.backends.external
 
+import org.apache.spark.h2o.H2OFrame
 import org.apache.spark.h2o.converters.WriteConverterCtx
 import org.apache.spark.h2o.converters.WriteConverterCtxUtils.UploadPlan
 import org.apache.spark.h2o.utils.SupportedTypes._
 import org.apache.spark.h2o.utils.{NodeDesc, ReflectionUtils}
 import org.apache.spark.ml.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.spark.sql.types._
-import water.{ExternalFrameConfirmationException, ExternalFrameUtils, ExternalFrameWriterClient}
+import water.fvec.Frame
+import water._
 
 class ExternalWriteConverterCtx(nodeDesc: NodeDesc, totalNumOfRows: Int, writeTimeout: Int, driverTimeStamp: Short) extends WriteConverterCtx {
 
-  val socketChannel = ExternalFrameUtils.getConnection(nodeDesc.hostname, nodeDesc.port, driverTimeStamp)
-  val externalFrameWriter = new ExternalFrameWriterClient(socketChannel)
+  private val externalFrameWriter = ExternalFrameWriterClient.create(nodeDesc.hostname, nodeDesc.port, driverTimeStamp)
 
-  /**
-    * This method closes the communication after the chunks have been closed
-    * @throws ExternalFrameConfirmationException in case of confirmation failure.
-    *         This has also effect of Spark stopping the current job and
-    *         rescheduling it
-    */
   override def closeChunks(numRows: Int): Unit = {
-    try{
-      externalFrameWriter.waitUntilAllWritten(writeTimeout)
-    } finally {
-      socketChannel.close()
-    }
+    externalFrameWriter.close()
   }
 
   /**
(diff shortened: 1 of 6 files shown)


jakubhava:
@mn-mikke good to review, however will be red until we integrate h2o 3.26.0.3 as it has necessary changes

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
```diff
@@ -17,54 +17,56 @@
...
+    externalFrameWriter.close()
   }
 
   /**
     * Initialize the communication before the chunks are created
     */
   override def createChunks(keystr: String, expectedTypes: Array[Byte], chunkId: Int, maxVecSizes: Array[Int],
                             sparse: Array[Boolean], vecStartSize: Map[Int, Int]): Unit = {
-    externalFrameWriter.createChunks(keystr, expectedTypes, chunkId, totalNumOfRows, maxVecSizes)
+    externalFrameWriter.createChunk(keystr, expectedTypes, chunkId, totalNumOfRows, maxVecSizes)
```
mn-mikke: what is a difference between `createChunk` and `createChunks` method?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
```diff
@@ -92,12 +95,34 @@ object ExternalWriteConverterCtx extends ExternalBackendUtils {
...
-      case _ : DataType => ReflectionUtils.supportedTypeOf(dt).javaClass
+      case _: DateType => Long.javaClass
+      case _: UserDefinedType[_ /*ml.linalg.Vector */ ] => classOf[Vector]
+      case _: DataType => ReflectionUtils.supportedTypeOf(dt).javaClass
     }
   }
+
+  def initFrame(key: String, columns: Array[String], timeout: Int): Unit = {
+    val (ip, port, timestamp) = leaderInfo()
+    val client = ExternalFrameWriterClient.create(ip, port, timestamp, timeout)
```
mn-mikke: Isn't there any overhead with establishing a connection to a leader. Shouldn't we create a connection to a leader only once?

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
```diff
@@ -111,3 +116,98 @@ class InternalWriteConverterCtx extends WriteConverterCtx {
...
+    if (!fr.vecs().isEmpty) {
+      val layouts = fr.vecs().map(_.espc())
+      val first = layouts.head
+      layouts.tail.foreach { espc =>
+        if (first != espc) {
+          throw new IllegalArgumentException(
+            s"""
+               | Invalid shape of H2O Frame:
+               |
+          | ${layouts.zipWithIndex.map { case (arr, idx) => s"Vec $idx has layout: ${arr.mkString(", ")}\n" }}
```
mn-mikke: nit: indentation

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
```diff
@@ -111,3 +116,98 @@ class InternalWriteConverterCtx extends WriteConverterCtx {
...
+    }
+
+    // Validate num of chunks in each vector
+    if (!fr.vecs().isEmpty) {
+      val first = fr.vecs.head
+      fr.vecs.tail.foreach { vec =>
+        if (vec.nChunks() != first.nChunks()) {
+          throw new IllegalArgumentException(
+            s"""
+               | Vectors have different number of chunks: ${fr.vecs().map(_.nChunks()).mkString(", ")}
```
mn-mikke: nit: one line?

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
```diff
@@ -111,3 +116,98 @@ class InternalWriteConverterCtx extends WriteConverterCtx {
...
+        nodes.foreach {
+          case (node, n) =>
+            Log.debug(node + ": " + n + " chunks.")
+        }
+      }
+    }
+
+    // Validate num of chunks in each vector
+    if (!fr.vecs().isEmpty) {
+      val first = fr.vecs.head
```
mn-mikke: Would it be possible to extract validation logic to a separated method? [MRTask](https://github.com/h2oai/sparkling-water/pull/1451/files#diff-21fbbe6ed7d8cdb3212cdcd8dc0c2bceR193) seems to use something similar.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
jakubhava: They are the same, we changed the name of the method on the H2O backend and this just just updates the code according to that

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
jakubhava: 👍

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
jakubhava: Good point!

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
```diff
@@ -111,3 +128,63 @@ class InternalWriteConverterCtx extends WriteConverterCtx {
     (vector.size until maxVecSize).foreach(idx => put(startIdx + idx, 0.0))
   }
 }
+
+object InternalWriteConverterCtx {
+
+  private def validateFrame(fr: Frame): Unit = {
+
+
```
mn-mikke: nit: extra lines

jakubhava: thanks!

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalWriteConverterCtx.scala
jakubhava: I have tried that, it creates slightly shorter code, but hurts readability.

PR #1452, opened 2019-08-19
[SW-1540] Remove unnecessary read confirmation timeout

jakubhava:
Can be merged after we integrate SW with H2O 3.26.0.3


jakubhava:
Thanks, will merge after H2O 3.26.0.3 is integrated

PR #1454, opened 2019-08-20
[SW-1541] Fix get-extended-h2o script to reflect new location of sparkling water releases

jakubhava:


bin/get-extended-h2o.sh
```diff
@@ -40,7 +40,7 @@ else
 fi
 
 h2odriver_file="h2odriver-$H2O_VERSION.$H2O_BUILD-${hadoop_version}extended.jar"
-h2odriver_url="${S3_RELEASE_BUCKET}/rel-${MAJOR_VERSION}/${PATCH_VERSION}/extended/${h2odriver_file}"
+h2odriver_url="${S3_RELEASE_BUCKET}/spark-${SPARK_MAJOR_VERSION}/${VERSION}/extended/${h2odriver_file}"
```
mn-mikke: What about the path of nightly builds? Is it relevant?

jakubhava: Good point

mn-mikke: Yep, no problem ;-)

jakubhava: thank you!

PR #1457, opened 2019-08-21
[SW-1543] Expose driver if, port, port range and extra memory percent configuration for external H2O cluster

jakubhava:


py/ai/h2o/sparkling/H2OConf.py
```diff
@@ -437,6 +437,22 @@ def set_run_as_user(self, user):
         self._jconf.setRunAsUser(user)
         return self
 
+    def setExternalH2ODriverIf(self, iface):
```
jakubhava: On purpose I'm changing to camel case as later, I would like to start deprecating the snake case getters and setters on `H2OConf`. After the deprecation is done, we can start generating this configuration class

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -62,6 +62,10 @@ trait ExternalBackendConf extends SharedBackendConf {
   def kerberosPrincipal = sparkConf.getOption(PROP_EXTERNAL_KERBEROS_PRINCIPAL._1)
   def kerberosKeytab = sparkConf.getOption(PROP_EXTERNAL_KERBEROS_KEYTAB._1)
   def runAsUser = sparkConf.getOption(PROP_EXTERNAL_RUN_AS_USER._1)
+  def externalH2ODriverIf = sparkConf.getOption(PROP_EXTERNAL_DRIVER_IF._1)
```
mn-mikke: WDYT about `sparkConf.getOption(PROP_EXTERNAL_DRIVER_IF._1, PROP_EXTERNAL_DRIVER_IF._2)`? I know that in both cases `None` is returned, but if 
```val PROP_EXTERNAL_DRIVER_IF = ("spark.ext.h2o.external.driver.if", None)``` specifies a default value...

File: doc/src/site/sphinx/configuration/configuration_properties.rst [hunk 3]
```diff
@@ -319,6 +319,24 @@ External backend configuration properties
...
++-------------------------------------------------------+----------------+-------------------------------------+
+| ``spark.ext.h2o.external.driver.port``                |  ``None``      | Port of mapper->driver callback     |
+|                                                       |                | interface. Default value means      |
+|                                                       |                | automatic detection.                |
++-------------------------------------------------------+----------------+-------------------------------------+
+| ``spark.ext.h2o.external.driver.port.range``          | ``None``       | Range portX-portY of mapper->driver |
+|                                                       |                | callback interface; eg:             |
+|                                                       |                | 50000-55000.                        |
++-------------------------------------------------------+----------------+-------------------------------------+
+| ``spark.ext.h2o.external.extra.memory.percent``       | ``10``         | extramempercent is a percentage of  |
```
mn-mikke: Is `extramempercent` mentioned somewhere in the documentation?

py/ai/h2o/sparkling/H2OConf.py
```diff
@@ -437,6 +437,22 @@ def set_run_as_user(self, user):
...
+    def setExternalH2ODriverPort(self, port):
+        self._jconf.setExternalH2ODriverPort(port)
+        return self
+
+    def setExternalH2ODriverPortRange(self, portrange):
+        self._jconf.setExternalH2ODriverPortRange(portrange)
+        return self
+
+    def setExternalExtraMemoryPercent(self, memoryPercent):
+        self._jconf.etExternalExtraMemoryPercent(memoryPercent)
```
mn-mikke: `etExternalExtraMemoryPercent` -> `setExternalExtraMemoryPercent`

doc/src/site/sphinx/configuration/configuration_properties.rst
jakubhava: Good catch, copy paste error. We can write the documentation as : `Specifies a percentage ..`, what do you think?

py/ai/h2o/sparkling/H2OConf.py
jakubhava: thanks!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
jakubhava: If I understand the motivation, this is to ensure we are prepared for cases when we change the default value of the underlying option & forget to change the getter?

The idea makes sense, however the code `getOption` does not accept second argument and `get` methods does not accept Option as value. Given these ^, I would keep it as it is. WDYT?

doc/src/site/sphinx/configuration/configuration_properties.rst
mn-mikke: `This option is a percentage of ...` sounds good to me.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mn-mikke: Ok, let's leave it as it is.

doc/src/site/sphinx/configuration/configuration_properties.rst
jakubhava: cool!

PR #1458, opened 2019-08-21
[SW-1539] Avoid boiler-plate code when introducing new test suite in PySparkling

jakubhava:


mn-mikke:
Do we still need the following? :
```
if __name__ == '__main__':
    generic_test_utils.run_tests([H2OMojoPredictionsTest], file_name="py_unit_tests_mojo_predictions_report")
```

jakubhava:
@mn-mikke please have another look. It's a bigger change as I needed to switch to pytest ( more recommended library which does all kind of neat stuff).

With this change, there is no boiler plate at all. One can just create test file in the desired directory and write test method. The pre-defined fixtures can be used to get reference to h2ocontext or spark

jakubhava:
Tests passing on jenkins as well (the external are failing but it's unrelated)

jakubhava:
@mn-mikke I have moved the files as suggested ( not delete & create). There just have been so many changes git shows the dfif like this

PR #1461, opened 2019-08-22
[SW-1547] Update H2OTargetEncoder according to changes in PUBDEV-6776

mn-mikke:
This PR requires https://github.com/h2oai/h2o-3/pull/3794 to be part of a offical H2O-3 release for a successful build.


jakubhava:
I had to revert, the build still fails against SW with H2O 3.26.0.4, can you please have a look @mn-mikke? Sorry for this

mn-mikke:
I still need to upgrade the change according to another PR https://github.com/h2oai/h2o-3/pull/3850.

PR #1464, opened 2019-08-23
[SW-1551] Fix path to external jars generated by ./gradlew extendJar

jakubhava:
The version suffix was newly added even though we explicitly set `version = null`. To overcome this, we explicitly configure the full jar name instead of it's separate components.


jakubhava:
@mn-mikke I'm going to merge this one to make the infra green asap, if you please have any comments, very happy to discuss them and apply in the following PR

PR #1467, opened 2019-08-24
[SW-1554] :sparkling-water-assembly-h2o:printExtendedJarLocation prints a wrong jar file name

mn-mikke:


jakubhava:
I believe this was fixed by https://github.com/h2oai/sparkling-water/commit/095ee7ef242b7266bd37010f1cdb22eb1a6c0a93 which is already merged. Update of Gradle caused that setting the version specifier to null did not have any effect, so we forced the full name manually. I wanted to keep the file name consistent with what was there before Gradle upgrade

assembly-h2o/build.gradle
```diff
@@ -106,7 +106,7 @@ if (project.hasProperty("doExtend")) {
     task printExtendedJarLocation() {
         doLast {
             def fileName = getH2OJarName(getOrDownloadOriginalJar())
-            println(file("private/extended/${fileName}.jar"))
+            println(file("private/extended/${fileName}-${version}.jar"))
```
jakubhava: I thought this was fixed by https://github.com/h2oai/sparkling-water/commit/095ee7ef242b7266bd37010f1cdb22eb1a6c0a93 but still not, just tried. We need to ensure the generated name of the extended jar does not include the version as that could break some existing users environments. and keep the `printExtendedJarLocation` as it was before

jakubhava:
@mn-mikke please have a look here https://github.com/h2oai/sparkling-water/pull/1470/files This change ensures we generate the name as it was before, thus also the location the print task prints is same now. 

I would close this PR if you agree and proceed with the one on the link

PR #1476, opened 2019-08-28
[SW-1564] Create  gradle task to build scala image for kubernetes

jakubhava:


jakubhava:
@mn-mikke when you have a sec, please have a look on this one :) Thanks a ton!

PR #1482, opened 2019-08-29
[SW-1496] Expose DRF in Sparkling Water and PySparkling

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "[SW_1496][MAJOR] Expose DRF in Sparkling Water and PySparkling" is missing JIRA issue number! Please specify it in the following form [SW-]

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
```diff
@@ -0,0 +1,73 @@
...
+#
+
+from pyspark import keyword_only
+
+from ai.h2o.sparkling import Initializer
+from ai.h2o.sparkling.ml.Utils import Utils
+from ai.h2o.sparkling.ml.algos.H2OAlgoBase import H2OAlgoBase
+from ai.h2o.sparkling.ml.params import H2ODRFParams
+
+javaMaxDoubleValue = (2 - 2 ** (-52)) * (2 ** 1023)
```
mn-mikke: WDYT about getting this separated to some utils class?

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class H2ODRF(H2ODRFParams, H2OAlgoBase):
+
+    @keyword_only
+    def __init__(self,
+                 binomialDoubleTrees=False,
+                 mtries=-1,
+                 ntrees=50,
+                 maxDepth=5,
```
mn-mikke: Shouldn't it be `maxDepth=20`?

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
```diff
@@ -0,0 +1,73 @@
...
+                 minRows=10.0,
```
mn-mikke: Shouldn't it be `minRows=1.0`?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
```diff
@@ -0,0 +1,84 @@
...
+import hex.tree.drf.DRFModel.DRFParameters
+import hex.tree.drf.{DRF, DRFModel}
+import org.apache.spark.ml.util.Identifiable
+
+/**
+  * H2O DRF algorithm exposed via Spark ML pipelines.
+  */
+class H2ODRF(override val uid: String) extends H2OSupervisedAlgorithm[DRF, DRFModel, DRFParameters] with H2ODRFParams {
+
+  def this() = this(Identifiable.randomUID("drf"))
```
mn-mikke: WDYT about `"drf"` -> `"H2ODRF"`?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
```diff
@@ -0,0 +1,84 @@
...
+}
+
+object H2ODRF extends H2OParamsReadable[H2ODRF]
+
+/**
+  * Parameters for Spark's API exposing underlying H2O model.
+  */
+trait H2ODRFParams extends H2OAlgoSharedTreeParams[DRFParameters] {
```
mn-mikke: WDYT about having `H2ODRFParams` in a separate file as it's in PySparkling?

py/tests/unit/with_runtime_sparkling/test_drf.py
```diff
@@ -0,0 +1,115 @@
...
+import os
+from pyspark.ml import Pipeline, PipelineModel
+from pyspark.mllib.linalg import *
+from pyspark.mllib.linalg import *
+from pyspark.sql.types import *
+from pyspark.sql.types import *
+from pysparkling.ml import H2ODRF
+
+def testParams():
+    drf = H2ODRF(binomialDoubleTrees=False,
```
mn-mikke: What is the purpose of this test? Wouldn't it be better to set values to something non-default?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
jakubhava: fine with me, we can actually do `classOf[H2OGBM].getSimpleName`. WDYT?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
jakubhava: we can do that to be consistent 👍

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
jakubhava: Great catch!, copy paste error

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
jakubhava: Great catch!, copy paste error

py/tests/unit/with_runtime_sparkling/test_drf.py
jakubhava: Initially, the purpose of this test was that all options can be passed to the algorithms without any exceptions - back then we did not have the type converters validating the input.

Maybe we do not need to do such tests anymore and just create separate test suite for the type converts. WDYT?

py/src/ai/h2o/sparkling/ml/algos/H2ODRF.py
jakubhava: 👍 great idea

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
mn-mikke: SGTM! I hope it will work. I've tried something similar in past, but it wasn't possible because it was used in constructor, but maybe this case a bit different.

py/tests/unit/with_runtime_sparkling/test_drf.py
mn-mikke: I'm definitely for having such a tests in Python. Typo can be made very easily. Just would like to understand the context. Just wondering whether it's possible to write such a test in a generic way via reflection to able to test all algorithms with one logic.

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2ODRF.scala
jakubhava: just tried, works fine :)

py/tests/unit/with_runtime_sparkling/test_drf.py
jakubhava: I will do this in a following PR what do you think? Already started working on it

mn-mikke: ok, no problem

PR #1483, opened 2019-08-29
[SW-1445] Remove Deprecated setters on algorithms which has enum as argument

jakubhava:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -60,7 +59,9 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
   private var gridMojoModels: Array[H2OMOJOModel] = _
 
   override def fit(dataset: Dataset[_]): H2OMOJOModel = {
-    val algoParams = ${gridAlgoParams}
+    val algoParams = $ {
```
mn-mikke: Are we going to use this formating from now? What is negative about `val algoParams = ${gridAlgoParams}`?

jakubhava: I actually prefer the original one, Intelij auto-formated this like this. I will try to see what we can do about this. Thanks for catching this!

mn-mikke: no problem

jakubhava: Ah, I have found the cause why it was formatted like this. Actually we should use the parenthesis like this `$(gridAlgoParams)`

PR #1484, opened 2019-08-29
[SW-1573] Remove use of deprecated option spark.ext.h2o.external.cluster.num.h2o.nodes in the code

jakubhava:


mn-mikke:
Can you help me understand where is `spark.ext.h2o.external.cluster.num.h2o.nodes` deprecated? I can't see any removal of deprecation warnings

jakubhava:
Sure -> https://github.com/h2oai/sparkling-water/blob/9e27595090e84fc642f8f7215a2b54a9ec345001/core/src/main/scala/org/apache/spark/h2o/H2OConf.scala#L114  

https://github.com/h2oai/sparkling-water/blob/9e27595090e84fc642f8f7215a2b54a9ec345001/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala#L39

https://github.com/h2oai/sparkling-water/blob/9e27595090e84fc642f8f7215a2b54a9ec345001/py/src/ai/h2o/sparkling/H2OConf.py#L330

PR #1487, opened 2019-08-29
[SW-1518] Remove deprecated parameter colsampleBytree and related methods from H2OXGBoost

jakubhava:


jakubhava:
@mn-mikke what do you mean please? The fix removes the deprecated method on Scala and Python side, but maybe I am missing something

PR #1488, opened 2019-08-29
[SW-1575] Remove deprecated option spark.ext.h2o.external.read.confirmation.timeout and related getters and setters

jakubhava:


jakubhava:
Yup, when reading data from external backend we are just polling information and actually don't need any timeout, we get the info/data directly. 

When writing to external h2o backend though, we need to wait for previous operation to finish.
Such as writing to chunks can start only when frame was initialized on the backend. The code takes care of this protocol on H2O side and sparkling water does not now about this. However this option is still exposed on Sparkling Water side to give users control of this timeout in some slow network environments.

Does it make sense?

mn-mikke:
> Yup, when reading data from external backend we are just polling information and actually don't need any timeout, we get the info/data directly.
What if the external backend gets stuck for some reason? (bug in H2O-3)  Wouldn't SW get stuck as well? Can it happen?

jakubhava:
Yup, that is possible. However in this case I believe the right solution would be to provide a fix instead of introducing configuration, wdyt?

mn-mikke:
> the right solution would be to provide a fix instead of introducing configuration
Definitely agree with you.

On the other hand, the timeout gives users an opportunity to fail fast if there is a problem. In cases when you train heavy algorithms like `H2OAutoML` would be hard to distinguish whether the application runs for a long time on purpose or not.

I don't know, I'm just making up a scenario that couldn't happen in practice.

PR #1492, opened 2019-08-29
[SW-1577] Generic logic to test parametr passing to algo wrappers on PySParkling

jakubhava:
Requires jh/sw-1585 to be merged first

### DIFF
--- a/py/tests/unit/with_runtime_sparkling/algo_test_utils.py
+++ b/py/tests/unit/with_runtime_sparkling/algo_test_utils.py
@@ -1,0 +1,52 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from pysparkling.ml import *
+
+
+def getParamPairs(algoClass):
+    # Create dummy instance which we use just to obtain default values
+    params = algoClass()._defaultParamMap
+    kwargs = {}
+    for param in list(params):
+        kwargs[param.name] = params[param]
+    return kwargs
+
+
+def assertParamsViaConstructor(algoName, skip=[]):
+    AlgoClass = globals()[algoName]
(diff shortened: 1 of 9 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGLM.scala
```diff
@@ -62,8 +63,8 @@ trait H2OGLMParams extends H2OAlgoSupervisedParams[GLMParameters] {
   private val solver = stringParam("solver", "solver")
   private val tweedieVariancePower = doubleParam("tweedieVariancePower")
   private val tweedieLinkPower = doubleParam("tweedieLinkPower")
-  private val alpha = nullableDoubleArrayParam("alpha")
-  private val lambda_ = nullableDoubleArrayParam("lambda_", "lambda")
```
jakubhava: this `_` is breaking the api style. It was introduced because lambda is keyword on python side, but it is a hack. Rather rename the param. Renaming alpha as to be consistent

py/tests/unit/with_runtime_sparkling/algo_test_utils.py
```diff
@@ -0,0 +1,51 @@
...
+    instance = AlgoClass(**kwargs)
+
+    for name in kwargs:
+        if name not in skip:
+            # Assert that the getter is giving the value we passed via setter
+            setter = getattr(instance, "set" + name[:1].upper() + name[1:])
+            getter = getattr(instance, "get" + name[:1].upper() + name[1:])
+            setter(kwargs[name])
+            assert getter() == kwargs[name]
+
```
mn-mikke: nit: It seems that git wants one more line

jakubhava:
Will Add DRF, thanks!

Regarding setting up default values -> at this point the purpose of this test is to verify that all getters and setters are named without typos. We can improve testing non-default values, but would target that for another PR - https://0xdata.atlassian.net/browse/SW-1590( we would need to automatically determine the type of parameter, pass the value etc. It is doable, but I would say not critical with the basic coverage we have now )

py/tests/unit/with_runtime_sparkling/algo_test_utils.py
jakubhava: 👍

py/tests/unit/with_runtime_sparkling/algo_test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+from pysparkling.ml import *
+
+
+def getParamPairs(algoClass):
+    # Create dummy instance which we use just to obtain default values
+    params = algoClass()._defaultParamMap
+    kwargs = {}
+    for param in list(params):
+        kwargs[param.name] = params[param]
+        # Create instance with all parameters set up in constructor
```
mn-mikke: nit: Is this comment accurate?

jakubhava: Good catch, should be removed

jakubhava: fixed

PR #1493, opened 2019-08-30
[SW-1582] Make sure H2OGLM API is consistent with others and does not use the labmda_ hack

jakubhava:


py/src/ai/h2o/sparkling/ml/algos/H2OGLM.py
```diff
@@ -71,10 +71,14 @@ def __init__(self,
                  withDetailedPredictionCol=False,
                  featuresCols=[],
                  convertUnknownCategoricalLevelsToNa=False,
-                 convertInvalidNumbersToNa=False):
+                 convertInvalidNumbersToNa=False,
+                 namedMojoOutputColumns=True,
```
jakubhava: `namedMojoOutputColumns=True` parameter was missing

PR #1494, opened 2019-08-30
[SW-1583] Add missing namedMojoOutputParameter to PySparkling Algo constructors

jakubhava:
H2OGLM has it added as part of https://github.com/h2oai/sparkling-water/pull/1493/files


mn-mikke:
For `H2OGLM`, I've noticed the change in a [different PR](https://github.com/h2oai/sparkling-water/pull/1493/files#r319384622)

jakubhava:
Great catch, thanks!

PR #1496, opened 2019-08-30
[SW-1585] Several H2OGrid Search bug fixes & improvements

jakubhava:
This PR does:
 - Deprecate the old `selectBestModelByDecresing` in favor of `selectBestModelOrdering` as that allows the user more explicit control and we can avoid the constant warning
- Fix bug in strategy param ( `link` -> `strategy` )
- Fix critical bug! updateH2OParams method needs to be called before passing parameters to grid to ensure the H2O parameters object is filled with params from Spark wrapper
- Ensure we have getter and setter for algo param as `algo` param was the only one across the API of all algos without getter
- Improve internal algo handling ( just single algo parameter, no need for the dummy one)
-> Bug fix! Pass params from python to java before calling updateH2OParameters
- Don't serialize binary objects


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -206,24 +206,21 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
       (m, extractMetrics(m).find(_._1 == metric).get._2)
     }
 
-    val ordering = if (getSelectBestModelBy() == H2OGridSearchMetric.AUTO.name()) {
-      logWarning("'selectBestModelBy' parameter is specified to 'AUTO', but you also specified 'selectBestModelDecreasing'." +
```
jakubhava: Make it more transparent by making creating Enum for ordering which says auto as well.

By default both values are auto and we can avoid this warninig which was printed all the time

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -441,21 +436,24 @@ trait H2OGridSearchParams extends H2OCommonSupervisedParams {
...
 
-  def getSelectBestModelDecreasing(): Boolean = $(selectBestModelDecreasing)
+  def getSelectBestModelOrdering(): String = $(selectBestModelOrdering)
 
   //
   // Setters
   //
 
   def setAlgo(value: H2OSupervisedAlgorithm[_, _, _ <: Model.Parameters]): this.type = {
+    val m = value.getClass.getDeclaredMethod("updateH2OParams")
```
jakubhava: This was major bug in Grid search! We were passing algo to Grid on which it needs to perform the search, but the parameters for the model were still default before the method `updateH2OParams` which sets the paramters from spark wrapper to H2O parameter class wasn't called

py/src/ai/h2o/sparkling/ml/H2OStageBase.py
```diff
@@ -21,13 +21,12 @@
 class H2OStageBase(JavaMLReadable, JavaMLWritable):
 
     # Set default values directly from Scala so we don't have to duplicate it on PySpark side
-    def _setDefaultValuesFromJava(self, skip=[]):
+    def _setDefaultValuesFromJava(self):
```
jakubhava: no need to skip, it's consistent now

py/src/ai/h2o/sparkling/ml/params/H2OTypeConverters.py
```diff
@@ -221,3 +222,46 @@ def convert(value):
...
+            else:
+                raise TypeError("Invalid type.")
+
+        return convert
+
+    @staticmethod
+    def __updateH2OParams(algo):
+        gateway = SparkContext._gateway
+        classArray = gateway.new_array(_jvm().Class, 0)
+        m = algo.getClass().getDeclaredMethod("updateH2OParams", classArray)
```
jakubhava: Updating H2O params in the same way as ^

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -441,21 +436,24 @@ trait H2OGridSearchParams extends H2OCommonSupervisedParams {
 
   def getSelectBestModelBy(): String = $(selectBestModelBy)
 
-  def getSelectBestModelDecreasing(): Boolean = $(selectBestModelDecreasing)
```
mn-mikke: Shouldn't this be deprecated?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -487,6 +485,8 @@ trait H2OGridSearchParams extends H2OCommonSupervisedParams {
     set(selectBestModelBy, validated)
   }
 
-
-  def setSelectBestModelDecreasing(value: Boolean): this.type = set(selectBestModelDecreasing, value)
```
mn-mikke: Shouldn't this be deprecated?

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
```diff
@@ -79,28 +78,17 @@ class H2OGridSearchParams(H2OCommonSupervisedParams):
...
         H2OTypeConverters.toEnumString("ai.h2o.sparkling.ml.algos.H2OGridSearchMetric"))
 
-    selectBestModelDecreasing = Param(
+    selectBestModelOrdering = Param(
         Params._dummy(),
-        "selectBestModelDecreasing",
-        "selectBestModelDecreasing",
-        H2OTypeConverters.toBoolean())
+        "selectBestModelOrdering",
+        "selectBestModelOrdering",
```
mn-mikke: nit: Better description?

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
```diff
@@ -123,23 +111,20 @@ def getStoppingMetric(self):
     def getSelectBestModelBy(self):
         return self.getOrDefault(self.selectBestModelBy)
 
-    def getSelectBestModelDecreasing(self):
```
mn-mikke: ditto

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
```diff
@@ -159,5 +144,5 @@ def setStoppingMetric(self, value):
     def setSelectBestModelBy(self, value):
         return self._set(selectBestModelBy=value)
 
-    def setSelectBestModelDecreasing(self, value):
```
mn-mikke: ditto

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: I think at this particular change we are fine without deprecation as it's not really highly exposed API, but we can deprecated, sure

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
jakubhava: Fixed, good catch!

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
jakubhava: 👍

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
jakubhava: 👍

jakubhava:
@mn-mikke please have a second look

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -120,15 +120,15 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
     gridModels = sortGrid(grid)
     gridMojoModels = gridModels.map { m =>
       val data = ModelSerializationSupport.getMojoData(m)
-      H2OMOJOModel.createFromMojo(data, Identifiable.randomUID(s"${$(gridAlgoParams).algoName()}_mojoModel"))
+      H2OMOJOModel.createFromMojo(data, Identifiable.randomUID(s"${algoParams.algoName()}_mojoModel"))
```
jakubhava: The returned model is regular model like GBM or GLM ( not grid). This rename ensure this

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: done

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -364,12 +359,22 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
   }
 
   override def copy(extra: ParamMap): this.type = defaultCopy(extra)
+
+  private def extractH2OParameters(algo: H2OSupervisedAlgorithm[_, _, _ <: Model.Parameters]): Model.Parameters = {
+    val m = algo.getClass.getDeclaredMethod("updateH2OParams")
+    m.setAccessible(true)
+    m.invoke(algo)
+
+    val field = PojoUtils.getFieldEvenInherited(algo, "parameters")
```
mn-mikke: Isn't a concept of `POJO` deprecated?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -364,12 +359,22 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
...
+    field.setAccessible(true)
+    field.get(algo).asInstanceOf[Model.Parameters]
+  }
 }
 
 object H2OGridSearch extends H2OParamsReadable[H2OGridSearch] {
 
   object SupportedAlgos extends Enumeration {
-    val gbm, glm, deeplearning, xgboost = Value // still missing pipeline wrappers for KMeans & drf
+    val gbm, glm, deeplearning, xgboost = Value
```
mn-mikke: Couldn't we add DRF as well?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -442,22 +441,18 @@ trait H2OGridSearchParams extends H2OCommonSupervisedParams {
 
   def getSelectBestModelBy(): String = $(selectBestModelBy)
 
-  def getSelectBestModelDecreasing(): Boolean = $(selectBestModelDecreasing)
+  @DeprecatedMethod("getSelectBestModelOrdering")
+  def getSelectBestModelDecreasing(): Boolean = {
+    val ordering = getSelectBestModelOrdering()
+    ordering == "ASC" // The default value was DESC as well
```
mn-mikke: `ordering == "DESC"` ?

What about the case  with `"AUTO"`?

ml/src/main/scala/ai/h2o/sparkling/ml/params/AlgoParam.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import ai.h2o.sparkling.ml.algos._
+import hex.Model
+import org.apache.spark.ml.param.{Param, ParamPair, Params}
+import org.json4s.JsonAST.{JArray, JNull, JString}
+import org.json4s.JsonDSL._
+import org.json4s._
+import org.json4s.jackson.JsonMethods.{compact, parse, render}
+
+class AlgoParam(parent: Params, name: String, doc: String, isValid: H2OSupervisedAlgorithm[_, _, _ <: Model.Parameters] => Boolean)
```
mn-mikke: Is the serde logic covered somewhere in unit tests?

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
```diff
@@ -124,22 +138,24 @@ def getSelectBestModelBy(self):
         return self.getOrDefault(self.selectBestModelBy)
 
     def getSelectBestModelDecreasing(self):
-        return self.getOrDefault(self.selectBestModelDecreasing)
+        Utils.deprecationWarning("getSelectBestModelDecreasing", "getSelectBestModelOrdering")
+        ordering = self.getOrDefault(self.selectBestModelOrdering)
+        return ordering is "ASC"  # The default value was DESC as well
```
mn-mikke: `ordering is "DESC"`?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -442,22 +441,18 @@ trait H2OGridSearchParams extends H2OCommonSupervisedParams {
...
-    val field = PojoUtils.getFieldEvenInherited(value, "parameters")
-    field.setAccessible(true)
-    val algoParams = field.get(value).asInstanceOf[Model.Parameters]
-    if (!H2OGridSearch.SupportedAlgos.isSupportedAlgo(algoParams.algoName())) {
-      throw new IllegalArgumentException(s"Grid Search is not supported for the specified algorithm '$value'. Supported " +
-        s"algorithms are ${H2OGridSearch.SupportedAlgos.allAsString}")
-    }
-    set(gridAlgoParams, algoParams)
-  }
+  def setAlgo(value: H2OSupervisedAlgorithm[_, _, _ <: Model.Parameters]): this.type = set(algo, value)
```
mn-mikke: Couldn't the output algorithm be any from supported algorithms? Is this information determined during `fit` phase?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: This is something a bit different -> PojoUtils means `Plain Old Java Object Utils` and are some sort H2O's reflection utils. This particular call just finds the `parameters` field, even if it's defined in any of the super classes

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: Exactly!, but that will go only to master as DRF is in master only - see this jira https://0xdata.atlassian.net/browse/SW-1588

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -207,24 +207,19 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
...
-        " order.")
-      // in case the user did not specified the metric, override the ordering to ensure we return the best model first
-      grid.getModels()(0)._output._training_metrics match {
-        case _: ModelMetricsRegression => Ordering.Double
-        case _: ModelMetricsBinomial => Ordering.Double.reverse
-        case _: ModelMetricsMultinomial => Ordering.Double
-      }
-    } else {
-      if (getSelectBestModelDecreasing()) {
+    val ordering = H2OGridSearchOrdering.valueOf(getSelectBestModelOrdering()) match {
```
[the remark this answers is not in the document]
jakubhava: This is actually good though and I didn't think of it during this clean-up. The purpose of this option is historic and after giving it a thought, agree, we should not need it.

Thanks!

py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py
jakubhava: Good catch! This won't be needed at the end as the option will just get deprecated without replacement in a form of `selectBestModelOrdering`

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: This won't be needed at the end as the option will just get deprecated without replacement in a form of `selectBestModelOrdering`

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
mn-mikke: Ok, thanks for your explanation!

ml/src/main/scala/ai/h2o/sparkling/ml/params/AlgoParam.scala
jakubhava: Good catch, wrote several tests in the H2OGridSearchTestSuite

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: I moved the checked for the algorithm into the setter, so the user is warned right away if the incorrect algo is being set

PR #1505, opened 2019-09-04
[SW-1600] Fix IIOB when using calibrated probabilities on MOJO

EstianDS:
`calibrateClassProbabilities` expects an array of size 3, but its pasing in an array of size 2 for `supportsCalibratedProbabilities`, this causes an `java.lang.ArrayIndexOutOfBoundsException: 2` error when trying to score a mojo that was trained with the calibration option enabled.


h2o-ops:
Hi @EstianDS, your PR title "Update H2OMOJOPredictionBinomial.scala" is missing JIRA issue number! Please specify it in the following form [SW-]

jakubhava:
@EstianDS This is great catch, thank you for your contribution!

Could I ask you to write a test for this behaviour? It would help us a lot! It can be put into class `ai.h2o.sparkling.ml.algos.BinomialPredictionTestSuite`

PR #1506, opened 2019-09-04
[SW-1597] Add a note informing that terraform template is available only for Spark 2.4

mn-mikke:
- Spark 2.1, 2.2 -  JupyterHub available in EMR only with versions Spark 2.3+
- Livy 0.6 supporting Spark 2.2-2.4 - Only available in the EMR version with Spark 2.4

### DIFF
--- a/doc/src/site/sphinx/deployment/terraform_emr.rst
+++ b/doc/src/site/sphinx/deployment/terraform_emr.rst
@@ -1,7 +1,7 @@
 Start Sparkling Water on Amazon EMR using our Terraform Template
 ----------------------------------------------------------------
 
-**This terraform template supports only Sparkling Water builds for Apache Spark 2.4**
+**This terraform template supports only Sparkling Water builds for Apache Spark 2.4 and higher.**
 
 Sparkling Water comes with the pre-defined Terraform templates that can be used to
 deploy Sparkling Water to Amazon EMR.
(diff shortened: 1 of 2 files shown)


doc/src/site/sphinx/deployment/terraform_emr.rst
```diff
@@ -49,10 +51,10 @@ Sparkling Water provides 3 templates/modules:
    - ``aws_secret_key`` (mandatory) - secret key to access AWS
    - ``aws_ssh_public_key`` (optional) - public key (to be able to access EC2 instances via ssh later)
    - ``aws_region`` (optional) - AWS region. Defaults to ``us-east-1``.
-   - ``aws_emr_version`` (optional) - EMR version. Defaults to ``emr-2.20.0``.
+   - ``aws_emr_version`` (optional) - EMR version. Defaults to ``emr-2.26.0``.
```
jakubhava: We could do do `SUBST_EMR_VERSION` instead of `emr-2.26.0` so it is less error-prone

doc/src/site/sphinx/deployment/terraform_emr.rst
```diff
@@ -1,6 +1,8 @@
 Start Sparkling Water on Amazon EMR using our Terraform Template
 ----------------------------------------------------------------
 
+**This terraform template supports only Sparkling Water builds for Apache Spark 2.4**
```
jakubhava: Should we mention Apache Spark 2.4 and higher?

doc/src/site/sphinx/deployment/terraform_emr.rst
mn-mikke: My original intention was to write details for Spark 2.4 to all SW versions, but you are right these things are easy to forget to change when emr gets upgraded.

PR #1507, opened 2019-09-04
[SW-1598] Deprecate cases in H2OMojoPrediction where the prediction column does not directly contain the predicted value.

jakubhava:
The other algos have it deprecated already. Starting from next release, the predicted column will contain the most important prediction output, the rest will be in detailed_prediction


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPredictionAnomaly.scala
```diff
@@ -25,6 +25,8 @@ import org.apache.spark.sql.{Column, Row}
 trait H2OMOJOPredictionAnomaly extends H2OMOJOPredictionUtils {
   self: H2OMOJOModel =>
   def getAnomalyPredictionUDF(): UserDefinedFunction = {
+    logWarning("Starting from next major release, the content of 'prediction' column will be generated to " +
```
mn-mikke: nit: `next major release` -> `the next major release` in all warnings

jakubhava: thanks!

PR #1508, opened 2019-09-04
[SW-1519] Switch to single value in the predictionCol and put all the details on the detailedPredictionCol

jakubhava:


ml/src/test/scala/ai/h2o/sparkling/ml/algos/RegressionPredictionTestSuite.scala
```diff
@@ -100,8 +100,7 @@ class RegressionPredictionTestSuite extends FunSuite with Matchers with SharedH2
     val model = algo.fit(dataset)
 
     val datasetFields = dataset.schema.fields
-    val valueField = StructField("value", DoubleType, nullable = false)
-    val predictionColField = StructField("prediction", StructType(valueField :: Nil), nullable = false)
+    val predictionColField = StructField("prediction", DoubleType, nullable = false)
```
mn-mikke: How we handle cases when H2O-3 throws an exception on a record? Can it ever happen?

jakubhava: I think we do not handle such cases -> I can see that the prediction methods have PredictException in their signature on H2O's side. We can improve this as part of another PR. Good catch!

PR #1513, opened 2019-09-10
[SW-1553][FOLLOWUP] Fix the benchmark pipeline

mn-mikke:


jenkins/Jenkinsfile-benchmarks
```diff
@@ -14,7 +14,7 @@ String getDockerImageVersion() {
 }
 
 def withDocker(code) {
-    def image = 'opsh2oai/sparkling_water_tests:' + getDockerImageVersion()
+    def image = 'harbor.h2o.ai/opsh2oai/sparkling_water_tests:' + getDockerImageVersion()
```
jakubhava: Should be done using this approach https://github.com/h2oai/sparkling-water/blob/f31195e0960025139d25900260317bf00f7518ba/jenkins/Jenkinsfile-release#L36

mn-mikke: Thanks!

PR #1514, opened 2019-09-10
[SW-1603] Install Terraform to Docker Image Running Tests

mn-mikke:


mn-mikke:
Thanks for catching this!!!

PR #1515, opened 2019-09-11
[SW-1604] PySparkling fails to parse pyspark version with build number, such as 2.3.1.dev0

jakubhava:


py/src/ai/h2o/sparkling/VersionComponents.py
```diff
@@ -20,13 +20,23 @@
...
         result.sparklingMajorVersion = match.group(2)
         result.sparklingMinorVersion = match.group(3)
         result.nightlyVersion = match.group(5)
-        result.sparkVersion = match.group(6)
+        result.sparkMajorVersion = match.group(6)
+        return result
+
+    @staticmethod
+    def parseFromPySparkVersion(version):
+        match = re.search(r"^(\d+.\d+).(\d+)(.([0-9A-Za-z]+))?$", version)
```
mn-mikke: nit: `.` will match to any character, `\.` will match just a dot. It's just a note, it won't have any effect on components parsing. The same nit you can find in the regexp above!

py/src/ai/h2o/sparkling/VersionComponents.py
```diff
@@ -20,13 +20,23 @@
...
+    def parseFromSparklingWaterVersion(version):
         match = re.search(r"^((\d+\.\d+).(\d+))(-(\d+))?-(\d+.\d+)$", version)
         result = VersionComponents()
         result.fullVersion = match.group(0)
         result.sparklingVersion = match.group(1)
         result.sparklingMajorVersion = match.group(2)
         result.sparklingMinorVersion = match.group(3)
         result.nightlyVersion = match.group(5)
-        result.sparkVersion = match.group(6)
+        result.sparkMajorVersion = match.group(6)
```
mn-mikke: nit: Good point with `sparkVersion`, it's not the full version! But since Spark uses versioning schame `major.minor.patch` would other people associate it with `2` instead of `2.4`?

py/tests/unit/simple/test_version_components.py
```diff
@@ -15,29 +15,49 @@
...
+
+def testParseSparkVersion():
+    version = "2.4.1"
+
+    components = VersionComponents.parseFromPySparkVersion(version)
+
+    assert components.fullVersion == version
+    assert components.sparkMajorVersion == "2.4"
+    assert components.sparkPatchVersion == "1"
+    assert components.sparkBuildVersion == None
```
mn-mikke: nit: `==` => `is`?

py/src/ai/h2o/sparkling/VersionComponents.py
jakubhava: Thanks, will fix!

py/tests/unit/simple/test_version_components.py
jakubhava: 👍

py/src/ai/h2o/sparkling/VersionComponents.py
jakubhava: We can call it as `result.sparkMajorMinorVersion`?

However we don't expose( I would say on purpose) the VersionComponents from `ai.h2o package` into `pysparkling` which is main entry point to PySparkling users.

But agree, will change it so are definitely sure what it means 👍 Thanks!

mn-mikke: `sparkMajorMinorVersion` sounds good to me ;-)

PR #1520, opened 2019-09-11
[SW-1588] Add DRF to Grid Search

mn-mikke:


ml/src/test/scala/ai/h2o/sparkling/ml/algos/H2OGridSearchTestSuite.scala
```diff
@@ -69,6 +69,17 @@ class H2OGridSearchTestSuite extends FunSuite with Matchers with SharedH2OTestCo
     testGridSearch(xgboost, hyperParams)
   }
 
+  test("H2O Grid Search DRF Pipeline") {
+    val drf = new H2ODRF()
+    val hyperParams: mutable.HashMap[String, Array[AnyRef]] = mutable.HashMap()
+    hyperParams += "_ntrees" -> Array(1, 10, 30).map(_.asInstanceOf[AnyRef])
```
mn-mikke: I think it would be nice if we referenced hyper parameters with spark names in future

jakubhava: yes, absolutely agree that we should use arguments we use in sw, not the internal H2O ones. Can you please create JIRA?

jakubhava:
One last place needs to be changed: https://github.com/h2oai/sparkling-water/blob/ad3529c4211b62ac4554431d153845e567f07fde/py/src/ai/h2o/sparkling/ml/params/H2OGridSearchParams.py#L91

ml/src/test/scala/ai/h2o/sparkling/ml/algos/H2OGridSearchTestSuite.scala
mn-mikke: https://0xdata.atlassian.net/browse/SW-1608

jakubhava: 👍 thank you!

mn-mikke:
> One last place needs to be changed
Thanks!

PR #1525, opened 2019-09-13
[SW-1615] Remove examples from assembly jar

jakubhava:


examples/build.gradle
```diff
@@ -21,14 +21,14 @@ jar {
...
+
 dependencies {
-    // Sparkling Water Core
-    compile(project(":sparkling-water-core")) {
-        exclude group: "javax.servlet", module: "servlet-api"
-    }
 
-    // Sparkling Water ML
-    compile project(":sparkling-water-ml")
+    sparklingWaterAssemblyJar project(path: ':sparkling-water-assembly', configuration: 'shadow')
```
jakubhava: Whole example should depend on the full assembly  as it has correct shadowing rules applied and and simulates how user is using sw

examples/build.gradle
```diff
@@ -44,18 +44,15 @@ dependencies {
     testCompile "junit:junit:4.11"
     // to reuse some utils test classes defined in the core
     testCompile project(path: ':sparkling-water-core', configuration: 'testArchives')
-    testCompile project(":sparkling-water-extension-stack-trace")
```
jakubhava: in assembly already

PR #1526, opened 2019-09-13
[SW-1611] Script tests are not being executed correctly - some tests are not being executed

jakubhava:
TODO: Pipeline tests are still failing, rest is fixed


examples/build.gradle
```diff
@@ -30,6 +30,16 @@ dependencies {
     sparklingWaterAssemblyJar project(path: ':sparkling-water-assembly', configuration: 'shadow')
     compile project(path: ':sparkling-water-assembly', configuration: 'shadow')
 
+    // H2O HDFS Persistent layer
```
jakubhava: For Some reason, Spark 2.1 requires this dependency in order to find some apache classes, will investigate later why

PR #1532, opened 2019-09-17
[SW-1618] Correctly handle feature names during prediction time

jakubhava:
@mn-mikke I will wait for the tests to finish, but this change should be final.

What we need to do is to store the names of the original features, but after the "flattening" of structs

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
@@ -48,7 +48,7 @@ trait H2OAlgoCommonUtils extends H2OCommonParams {
     val input = h2oContext.asH2OFrame(dataset.select(cols: _*).toDF())
 
     // Our MOJO wrapper needs the full column name before the array/vector expansion in order to do predictions
-    val internalFeatureCols = SchemaUtils.flattenStructsInDataFrame(dataset.select(getFeaturesCols().map(col): _*).toDF).columns
+    val internalFeatureCols = SchemaUtils.flattenStructsInDataFrame(dataset.select(getFeaturesCols().map(col): _*)).columns
     if (getAllStringColumnsToCategorical()) {
       H2OFrameSupport.allStringVecToCategorical(input)
     }
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
@@ -17,6 +17,7 @@
 package ai.h2o.sparkling.ml.algos
 
 import ai.h2o.sparkling.ml.params.H2OCommonParams
+import ai.h2o.sparkling.ml.utils.SchemaUtils
 import org.apache.spark.h2o.H2OContext
 import org.apache.spark.sql.functions.col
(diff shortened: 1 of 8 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgorithm.scala
```diff
@@ -56,10 +56,11 @@ abstract class H2OAlgorithm[B <: H2OBaseModelBuilder : ClassTag, M <: H2OBaseMod
     val binaryModel: H2OBaseModel = trainModel(parameters)
     val mojoData = ModelSerializationSupport.getMojoData(binaryModel)
     val modelSettings = H2OMOJOSettings.createFromModelParams(this)
-    H2OMOJOModel.createFromMojo(
+    val model = H2OMOJOModel.createFromMojo(
       mojoData,
       Identifiable.randomUID(binaryModel._parms.algoName()),
       modelSettings)
+    model.set(model.featuresCols, getFeaturesCols())
```
jakubhava: When the mojo is trained in Sparkling Water, we override the feature names with the non-expanded ones as the code later takes care of the column expansion.

jakubhava:
@mn-mikke this change is ready to review. It ti still possible it might be red on external backend due to the intermittent bug there, but that is fixed as part of https://github.com/h2oai/sparkling-water/pull/1541

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
```diff
@@ -46,6 +47,8 @@ trait H2OAlgoCommonUtils extends H2OCommonParams {
     val h2oContext = H2OContext.getOrCreate(SparkSession.builder().getOrCreate())
     val input = h2oContext.asH2OFrame(dataset.select(cols: _*).toDF())
 
+    // Our MOJO wrapper needs the full column name before the array/vector expansion in order to do predictions
+    val internalFeatureCols = SchemaUtils.flattenStructsInDataFrame(dataset.select(getFeaturesCols().map(col): _*).toDF).columns
```
mn-mikke: Isn't `.toDF` extra in this context?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -134,5 +134,15 @@ object H2OMOJOModel extends H2OMOJOReadable[H2OMOJOModel] with H2OMOJOLoader[H2O
     model.setMojoData(mojoData)
     model
   }
-}
 
+  // Internal method used only within Sparkling Water pipelines.
+  // When H2OMOJOModel is created from existing mojo created in H2O-3, we do not need to set original features,
```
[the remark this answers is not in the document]
jakubhava: great catch!

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
jakubhava: good point!

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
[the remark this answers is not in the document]
jakubhava: I was about to fix it, but it is actually correct. I rephrased it to make more sense

PR #1536, opened 2019-09-18
[SW-1626] Intermittent failure during conversion to h2o frame on External backend in PySparking

jakubhava:


jakubhava:
Still does not fix the issue, need to look more

PR #1539, opened 2019-09-18
[SW-1630] Check Version of H2O Assembly Jar via Rest call

mn-mikke:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -407,23 +407,13 @@ object H2OContext extends Logging {
 
   private class H2OContextRestAPIBased(spark: SparkSession, conf: H2OConf) extends H2OContext(spark, conf) with H2OContextRestAPIUtils {
     // Once H2O exposes leader node via rest api, remove the nodes argument
-    override protected def getFlowIp(nodes: Array[NodeDesc]): String = nodes.head.hostname
+    override protected def getFlowIp(nodes: Array[NodeDesc]): String = getLeaderNode(nodes).hostname
```
jakubhava: note: the list of nodes is at this stage static, however leader node might change during H2O run-time

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -106,12 +106,8 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
       cmdToLaunch = cmdToLaunch ++ Seq[String]("-driverif", conf.h2oDriverIf.get)
     }
 
-    if (hc.getConf.h2oNodeWebEnabled) {
-      if (hc.getConf.contextPath.isDefined) {
-        cmdToLaunch = cmdToLaunch ++ Seq("-context_path", hc.getConf.contextPath.get)
-      }
-    } else {
-      cmdToLaunch = cmdToLaunch ++ Seq[String]("-J", "-disable_web")
```
mn-mikke: To be able to compare H2O-3 versions via REST call, I need to enable web on executors. @jakubhava if you agree with this approach, I will create a seperate PR deprecating related properties for the next fix release.

jakubhava: Yes, I think this is fine. We just need to check our security docs.

mn-mikke:
@jakubhava I had to do some changes after rebase, can you take a look one more time?

PR #1540, opened 2019-09-20
[SW-1636] Enable to specify outputCols on H2OTargetEncoder

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -74,6 +74,8 @@ class H2OTargetEncoder(override val uid: String)
 
   def setInputCols(values: Array[String]): this.type = set(inputCols, values)
 
+  def setOutputCols(values: Array[String]): this.type = set(outputCols, values)
```
jakubhava: Should we check that `values` can't be null?

scoring/src/main/scala/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.scala
```diff
@@ -75,7 +77,14 @@ trait H2OTargetEncoderParams extends Params {
 
   def getInputCols(): Array[String] = $(inputCols)
 
-  def getOutputCols(): Array[String] = getInputCols().map(_ + "_te")
+  def getOutputCols(): Array[String] = {
+    val columns = $(outputCols)
+    if (columns == null || columns.isEmpty) {
```
jakubhava: `outputCols` is `StringArrayParam` and it is not nullable, `null` should not be a valid option. I think we might just prevent user setting null, what do you thing?

jakubhava: Or we can keep the option to specify `null` but the parameter needs to be changed to `NullableStringArrayParam`

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
mn-mikke: That's a really good point! In Scala world, it's not a practice, but an user can pass anything. Do check this in other parts of API? I don't have any strong opinion about this, I will tailor it according to you preferences ;-)

scoring/src/main/scala/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.scala
mn-mikke: Yes, you are right, `null` is not a valid option. `columns == null` is extra

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
jakubhava: After today's discussion I would stick with the current API :)

PR #1541, opened 2019-09-21
[SW-1642] Prevent sending empty partitions when converting Spark Data frame to H2OFrame

jakubhava:
This one is finally fixing the missing chunk issue (EDIT, not yet :( )

Explanation here https://0xdata.atlassian.net/browse/PUBDEV-6904, I have put the fix into H2O so it can again receive and confirm empty chunks but also ensured that we do not send empty chunks for conversions as there is no value in that.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
@@ -99,14 +99,15 @@ private[h2o] object SparkDataFrameConverter extends Logging {
                       writeTimeout: Int, driverTimeStamp: Short, sparse: Array[Boolean], partitions: Seq[Int])
                      (context: TaskContext, it: Iterator[Row]): (Int, Long) = {
 
+    val chunkIdx = partitions.indexOf(context.partitionId())
     val (iterator, dataSize) = WriteConverterCtxUtils.bufferedIteratorWithSize(uploadPlan, it)
-    val con = WriteConverterCtxUtils.create(uploadPlan, context.partitionId(), writeTimeout, driverTimeStamp)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -52,7 +52,13 @@ private[h2o] object SparkDataFrameConverter extends Logging {
   def toH2OFrame(hc: H2OContext, dataFrame: DataFrame, frameKeyName: Option[String]): H2OFrame = {
     import ai.h2o.sparkling.ml.utils.SchemaUtils._
 
-    val flatDataFrame = flattenDataFrame(dataFrame)
+    val flatDataFrameOriginal = flattenDataFrame(dataFrame)
+    val numPartitions = flatDataFrameOriginal.rdd.partitions.length
+    val flatDataFrame = if (flatDataFrameOriginal.count() < numPartitions) {
+      flatDataFrameOriginal.coalesce(numPartitions)
```
mn-mikke: Please can you share details how `coalesce` can prevent sending empty partitions?

jakubhava: Yup. The issue is as explained here  https://0xdata.atlassian.net/browse/PUBDEV-6904. My thought was that if we ensure coalesce to number of elements in the spark data frame ( if num. of elements is smaller then number of partitions) then the coalesce would ensure that at least 1 element is in each partition. But as the tests are still failing, it might not behave like that ( did not debug since then). Any idea why? The H2O fix would fix it for sure, this is just to understand the issue a bit better

jakubhava:
Tried different approach: the method `sparkContext.runJob` accepts parameter which lists partitions on which we should run the job -> we can list all non-empty partitions and run the job on them

PR #1543, opened 2019-09-22
[SW-1631] Automatically increase client timeout on top of Azure

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -121,6 +121,7 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
       .add("-stacktrace_collector_interval", Some(conf.stacktraceCollectorInterval).filter(_ > 0))
       .add("-nthreads", Some(conf.nthreads).filter(_ > 0).orElse(conf.sparkConf.getOption("spark.executor.cores")))
       .add("-internal_security_conf", conf.sslConf)
+      .add("-client_disconnect_timeout", conf.clientCheckRetryTimeout)
```
mmalohlava: Nit: the naming does not match: disconnect v. retry.

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -64,6 +64,10 @@ class InternalH2OBackend(@transient val hc: H2OContext) extends SparklingBackend
       conf.setCloudName("sparkling-water-" + System.getProperty("user.name", "cluster") + "_" + conf.sparkConf.getAppId)
     }
 
+    if (AzureDatabricksUtils.isRunningOnAzureDatabricks(conf)) {
+      conf.setClientCheckRetryTimeout(600000)
```
mmalohlava: Is not it dangerous to force it? Better would be to use this path on Azure only if user does not specify the config property manually (use-case: imagine `600000` is not enough - how can user bump it  to `100000`?)

Also nit: i would put some message to log that we do not forget about it .

jakubhava: Thanks, good suggestions!

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
jakubhava: Yup, agree the naming is not ideal :( but it is already used it the same in external backend

PR #1544, opened 2019-09-22
[SW-1641] Remove requirement for the cluster size specification

jakubhava:
First draft


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -398,49 +402,50 @@ object H2OContext extends Logging {
...
+    override protected def initBackend(): Unit = {
+      val nodes = backend.init()
+      h2oNodes.append(nodes: _*)
+    }
   }
 
   private class H2OContextRestAPIBased(spark: SparkSession, conf: H2OConf) extends H2OContext(spark, conf) with H2OContextRestAPIUtils {
     // Once H2O exposes leader node via rest api, remove the nodes argument
-    override protected def getFlowIp(nodes: Array[NodeDesc]): String = nodes.head.hostname
+    override protected def getFlowIp(): String = conf.h2oCluster.get.split(":")(0)
```
jakubhava: h2oCluster is the IP:Port user specified to connect to external H2O

PR #1546, opened 2019-09-25
[SW-1644] PySparkling fails to parse spark version ending with +hadoop2.6

jakubhava:


py/src/ai/h2o/sparkling/VersionComponents.py
```diff
@@ -33,12 +33,12 @@ def parseFromSparklingWaterVersion(version):
 
     @staticmethod
     def parseFromPySparkVersion(version):
-        match = re.search(r"^((\d+)\.(\d+))\.(\d+)(\.([0-9A-Za-z]+))?$", version)
+        match = re.search(r"^((\d+)\.(\d+))\.(\d+)([\\.\\+]+([0-9A-Za-z\\+\\.]+))?$", version)
```
mn-mikke: WDYT about simplifying the part `([\\.\\+]+([0-9A-Za-z\\+\\.]+))?` of the regular expression to `(.+)?`? Is the suffix used somewhere in the code? Can the suffix start with delimiter `.`,`+`?

I'm just wondering how to generalize the solution for all possible suffixes of spark custom builds ( '-SNAPSHOT', '.cloudera3', etc.)

jakubhava: That's good point, we don't use the suffix at all at this moment

jakubhava: Maybe we can just collect anything after the patch version as suffix, what do you think?

mn-mikke: Sounds good to me!

jakubhava:
@mn-mikke please have a look, the regex should be generalized

PR #1547, opened 2019-09-26
[SW-1641] Remove requirement for the cluster size specification

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -375,15 +375,19 @@ abstract class H2OContext private(val sparkSession: SparkSession, conf: H2OConf)
 object H2OContext extends Logging {
 
   private class H2OContextClientBased(spark: SparkSession, conf: H2OConf) extends H2OContext(spark, conf) {
-    override protected def getFlowIp(nodes: Array[NodeDesc]): String = {
+
+    /** Runtime list of active H2O nodes */
+    protected val h2oNodes = mutable.ArrayBuffer.empty[NodeDesc]
```
mn-mikke: Any reason for using mutable collection?

jakubhava: not any more, good point! But will need to turn val into var

PR #1554, opened 2019-10-01
[SW-1654] Fix H2OConf test suite

jakubhava:
The issue was that H2OContext is now abstract class. However H2OConfTestSuite does not need started H2OContext, it is supposed to test H2OConf

### DIFF
--- a/core/src/test/scala/org/apache/spark/h2o/H2OConfTestSuite.scala
+++ b/core/src/test/scala/org/apache/spark/h2o/H2OConfTestSuite.scala
@@ -36,7 +36,7 @@ class H2OConfTestSuite extends FunSuite
       .set("spark.ext.h2o.client.ip", "10.0.0.100")
       .set("spark.ext.h2o.client.port.base", "1267")
       .set("spark.ext.h2o.node.port.base", "32333")
-      .set("spark.ext.h2o.cloud.timeout", (10*1000).toString)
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/apache/spark/h2o/H2OConfTestSuite.scala
```diff
@@ -50,15 +50,15 @@ class H2OConfTestSuite extends FunSuite
 
     val spark = SparkSession.builder().master("local").appName("test-local").config(sparkConf).getOrCreate()
 
-    val hc = H2OContext.getOrCreate(spark, new H2OConf(spark).setClusterSize(1))
```
mn-mikke: @jakubhava Please can you share details why the test was failing? the original piece of code looks legitimate

jakubhava: Sure -> originally in this test, we weren't starting starting H2OContext, but we were calling private constructor which does not start the backend, just created dummy instance. That works fine on master branch.

However on this branch I changed it to the normal creating of `H2OContext` as the `H2OContet` is now abstract. It started to fail as the configuration passed to H2OContext is not suitable for creating in external mode. And as the main reason we have this test is to test that H2OConf can store the configured values, I think it is fine to just create instance of H2OConf instead of H2OContext.

PR #1555, opened 2019-10-01
[SW-1655] Always require IP:Port of the external cluster in manual start mode.

jakubhava:
Currently, in manual cluster mode, it is optional to specify the H2OCluster endpoint. If it is not specified, H2O tries to find the cluster using broadcast search. In rest API based solution, we always need  to know the endpoint of the cluster.

This change accomplishes that, but also in the client based solution, not just rest api based to ensure unified behaviour. In my opinion it is better to be more explicit in manual mode. Anyway, we are planning to make Scala also rest API based at some point.

What we just need to do is to deprecate the behaviour and warn the users that endpoint will be always required. After we discuss this, I can create JIRA in our Major release which would accomplish this deprecation so we don't forget about that.


jakubhava:
Here is the jira for the deprecation https://0xdata.atlassian.net/browse/SW-1661

PR #1557, opened 2019-10-02
[SW-1656] Ensure the external cluster is killed if the client was not able to connect

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -217,7 +217,15 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
 
     H2OStarter.start(h2oClientArgs, false)
 
-    H2O.waitForCloudSize(hc.getConf.clusterSize.get.toInt, hc.getConf.cloudTimeout)
+    val expectedSize = hc.getConf.clusterSize.get.toInt
+    val discoveredSize = waitForCloudSize(expectedSize, hc.getConf.cloudTimeout)
+    if (discoveredSize < expectedSize) {
+      if (hc.getConf.isAutoClusterStartUsed) {
+        Log.err(s"Exiting! External H2O cluster was of size $discoveredSize but expected was $expectedSize!!")
+        H2O.shutdown(-1)
```
mn-mikke: what about a @H2O-3 client in manual mode? is there any?

mn-mikke: Or does this shout down the whole cluster?

jakubhava: Yup, it shuts down the whole cluster

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -217,7 +217,15 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
 
     H2OStarter.start(h2oClientArgs, false)
 
-    H2O.waitForCloudSize(hc.getConf.clusterSize.get.toInt, hc.getConf.cloudTimeout)
+    val expectedSize = hc.getConf.clusterSize.get.toInt
+    val discoveredSize = waitForCloudSize(expectedSize, hc.getConf.cloudTimeout)
+    if (discoveredSize < expectedSize) {
```
Tagar: is there is a scenario when discoveredSize > expectedSize? should it be also processed as an error?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -21,13 +21,27 @@ import org.apache.spark.SparkEnv
...
 import org.apache.spark.h2o.utils.NodeDesc
-import water.{ExternalFrameUtils, H2O}
+import water.{ExternalFrameUtils, H2O, Paxos}
 
 /**
   * Various helper methods used in the external backend
   */
 private[external] trait ExternalBackendUtils extends SharedBackendUtils {
 
+  protected def waitForCloudSize(x: Int, ms: Long): Int = {
```
Tagar: minor: x => expected_size, ms => timeout_ms?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -21,13 +21,27 @@ import org.apache.spark.SparkEnv
...
+    val start = System.currentTimeMillis()
+    while (System.currentTimeMillis() - start < ms) {
+      if (H2O.CLOUD.size() >= x && Paxos._commonKnowledge) {
```
Tagar: should it be `H2O.CLOUD.size() = x` ? or H2O.CLOUD.size() > expected_size is an okay state too?

Tagar: just for understanding - what is purpose of `Paxos._commonKnowledge` ?
thanks!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
mn-mikke: Thanks! We will fix it in a follow up PR.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
mn-mikke: Yes, I think so ->  if the manual start of the external backend is set and a user runs more H2O nodes then it's specified in the property `spark.ext.h2o.external.cluster.size`

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
mn-mikke: If the cloud is bigger than expected size, it indicates that client is able connect to H2O nodes. IMHO, I don't think it's a problem.

From `Paxos` class:
```
 Paxos detects and builds a "cloud" - a cooperating group of nodes, with mutual
 knowledge of each other.  Basically tracks all the nodes that *this* node
 has ever heard of, and when *all* of the other nodes have all heard of each
 other, declares the situation as "commonKnowledge", and a Cloud.
```

PR #1558, opened 2019-10-02
[SW-1658] Figure out better way of caching MOJOs (v1 & v2)

jakubhava:
Caching of MOJOv1 and MOJOv2 to prevent multiple loading on each executor


scoring/src/main/scala/ai/h2o/sparkling/ml/models/HasMojoDataV2.scala
```diff
@@ -0,0 +1,46 @@
...
+import ai.h2o.mojos.runtime.readers.MojoPipelineReaderBackendFactory
+import org.apache.spark.broadcast.Broadcast
+import org.apache.spark.sql.SparkSession
+
+private[models] trait HasMojoDataV2 extends HasMojoData {
+
+  private def loadMojoPipeline(mojoData: Array[Byte]): Broadcast[MojoPipeline] = {
+    val reader = MojoPipelineReaderBackendFactory.createReaderBackend(new ByteArrayInputStream(mojoData))
+    val pipeline = MojoPipeline.loadFrom(reader)
+    SparkSession.builder().getOrCreate().sparkContext.broadcast(pipeline)
```
mn-mikke: @jakubhava Please can you share details why broadcasting of `MojoPipeline` would be faster tham broadcasting of a byte array? Won't the mojo pipeline be serialized and deserialized again?

jakubhava: The issue was that we had to create MOJOPipeline instance all over again from bytes which takes time. I want to benchmark it in a bit to see whether this has some impact or not. But yup, it will be serialized and deserialized -> but all the time:

> Broadcast variables allow the programmer to keep a read-only variable cached on each machine rather than shipping a copy of it with tasks.

From now https://jaceklaskowski.gitbooks.io/mastering-apache-spark/spark-broadcast.html, that seems like exactly something we are looking for

mn-mikke: I know that you were trying multiple implementations in the past, but couldn't creation of `MOJOPipeline` instance for each thread avoided via `lazy val` on a object?

jakubhava: I believe the advantage of this approach is that we create the instance just once and transfer it to each executor just once. We do not need to even create new instance per thread. With bi g mojos, creating mojo pipeline takes long time.

I will do some benchmarks and post them here to see what impact (or not) this change has

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import org.apache.spark.expose.Logging
+import org.apache.spark.sql.SparkSession
+
+import scala.collection.mutable
+
+trait H2OMOJOBaseCache[B, M] extends Logging {
+  private object Lock
+
+  private val pipelineCache = mutable.Map.empty[String, B]
```
mmalohlava: WeakHashMap?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
```diff
@@ -0,0 +1,79 @@
...
+          val toDestroy = lastAccessMap.flatMap { case (uid, lastAccess) =>
+            val currentDiff = System.currentTimeMillis() - lastAccess
+            if (currentDiff > cleanupRetryTimeout) {
+              logDebug(s"Removing mojo $uid from cache as it has not been used for $cleanupRetryTimeout ms.")
+              Some(uid)
+            } else {
+              None
+            }
+          }
+          Lock.synchronized {
```
mmalohlava: Read and destroy is not synchronized - timestamp will be updated, but mojo will be still destroy - probably not fatal situation, will just cause reload from memory.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -32,7 +32,7 @@ import org.apache.spark.sql._
 import scala.collection.JavaConverters._
 
 class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOModel] with H2OMOJOPrediction {
-
+  H2OMOJOCache.startCleanupThread()
```
mmalohlava: should be static (=part of object) no?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
```diff
@@ -0,0 +1,79 @@
...
+  private val pipelineCache = mutable.Map.empty[String, B]
+  private val lastAccessMap = mutable.Map.empty[String, Long]
+
+  private lazy val sparkConf = SparkSession.builder().getOrCreate().sparkContext.getConf
+  private lazy val cleanupRetryTimeout = sparkConf.getInt("spark.ext.h2o.mojo.destroy.timeout", 10 * 60 * 1000)
+  private val cleanerThread = new Thread() {
+    override def run(): Unit = {
+      while (!Thread.interrupted()) {
+        try {
+          Thread.sleep(cleanupRetryTimeout)
```
mmalohlava: you can use:

```scala
val threadFactory = new ThreadFactoryBuilder().setDaemon(true).setNameFormat(threadName).build()
    val executor = new ScheduledThreadPoolExecutor(1, threadFactory)
    // By default, a cancelled task is not automatically removed from the work queue until its delay
    // elapses. We have to enable it manually.
    executor.setRemoveOnCancelPolicy(true)
    executor.scheduleAtFixedRate(new Runnable { ... })
```

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModel.scala
```diff
@@ -33,10 +33,7 @@ import scala.util.Random
 
 class H2OMOJOPipelineModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOPipelineModel] {
 
-  @transient private lazy val mojoPipeline: MojoPipeline = {
-    val reader = MojoPipelineReaderBackendFactory.createReaderBackend(new ByteArrayInputStream(getMojoData()))
-    MojoPipeline.loadFrom(reader)
-  }
+  H2OMOJOPipelineCache.startCleanupThread()
```
mmalohlava: I expect this will launch 4 threads if `--executor-cores=4`.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModel.scala
```diff
@@ -66,6 +63,7 @@ class H2OMOJOPipelineModel(override val uid: String) extends H2OMOJOModelBase[H2
   private val modelUdf = (names: Array[String]) =>
     udf[Mojo2Prediction, Row] {
       r: Row =>
+        val mojoPipeline = H2OMOJOPipelineCache.getMojoBackend(uid, getMojoData(), this)
```
mmalohlava: This is weird, what is motivation of getting pipeline inside UDF?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
mn-mikke: I really like simplicity of your proposal. But if I'm not mistaken, `WeakHashMap` evicts an entry by a GC thread if there are no strong reference to its key. IMHO, this situation could happen in the preparation period between execution of two tasks, especially when `--executor-cores=1`.

Or am I missing something in your proposal?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mn-mikke: I'm afraid that there isn't an equivalent to Java static block in Scala. A companion object is just an instance of a different JVM class.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
mmalohlava: right, good point! weakhashmap would optimize evection rate, but we need optimize hit rate (expectation is that multiple tasks will use mojo).

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mmalohlava: yes, sorry, i meant companion object.

mmalohlava: hmm, ran a test - my misunderstanding 🤦‍♂

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModel.scala
mn-mikke: Can you think of a solution that will utilize an existing thread from a shared thread pool? I'm afraid that java `Timer` or `ScheduledThreadPoolExecutor` always create a new thread...

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
mmalohlava: yes

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def startCleanupThread(): Unit = {
+    if (!cleanerThread.isAlive) {
+      cleanerThread.start()
+    }
+  }
+
+  def getMojoBackend(uid: String, bytesGetter: ()=> Array[Byte], model: M): B = Lock.synchronized {
+    if (!pipelineCache.contains(uid)) {
+      //println(s"MISS MOJO pipeline: thread=${Thread.currentThread().getName}, this=${this}")
```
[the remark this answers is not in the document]
mmalohlava: sorry, my fault. yes

PR #1559, opened 2019-10-02
[SW-1661] Deprecate multicast search for cluster in external backend in manual mode

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -388,6 +388,15 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
              |It can be set either on the configuration object or via '${SharedBackendConf.PROP_CLOUD_NAME._1}'
              |spark configuration property""".stripMargin)
       }
+
+      if (conf.h2oCluster.isEmpty) {
```
mn-mikke: What about automatic backend? Shouldn't this condition hold only for manual external backend?

jakubhava: Yup, you are right. But I think it does what we need at the moment as it is located in the else block which is executed in case of manual start mode - see line 330 where the if-else starts

mn-mikke: I see thanks!

PR #1564, opened 2019-10-04
[SW-1664] Enable Users to Specify Extra Http Headers for H2O Flow as SW Parameter

mn-mikke:


core/src/test/scala/org/apache/spark/h2o/backends/SharedBackendUtilsTestSuite.scala
```diff
@@ -0,0 +1,83 @@
+package org.apache.spark.h2o.backends
+
+import org.scalatest.{FunSuite, Matchers}
+
+class SharedBackendUtilsTestSuite extends FunSuite with Matchers {
+
+  val expectedHttpHeaderArgs = Seq(
+    "-add_http_header", "X-Frame-Options", "deny",
```
michalkurka: it might be misleading if used by an external person as an example - these headers are already built-in H2O

mn-mikke: These headers are used only for testing the parser. But no problem, I can uses different ones ;-)

michalkurka: yeah, I know - it is not an actual issue - it is just not the best example since it is redundant in actual deployment

your call - if you want to change it or not

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -158,6 +159,20 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
       .buildArgs()
   }
 
+  def parseStringToHttpHeaderArgs(headers: String): Seq[String] = {
+    val headerPattern = """^\s*([^:]+)\:\s*(.+)$""".r
+    headers.split('\n').flatMap{ header =>
```
Pscheidl: Minor: I'd suggest pre-compiling the resulting finite automaton for performance reasons, but it's not really crucial in this pase. Would personally use pre-compiled `scala.util.matching.Regex`.

mn-mikke: As per personal discussion, I will leave it as it's since regexp is used only once during the application run.

PR #1566, opened 2019-10-08
[SW-1667] Enable Users to Specify Block Size of Communication in External Backend

mn-mikke:
This PR requires https://github.com/h2oai/h2o-3/pull/3931 to be a part of a released version of H2O-3 for successful build.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
@@ -60,6 +60,8 @@ trait ExternalBackendConf extends SharedBackendConf {
   def externalH2ODriverPort = sparkConf.getOption(PROP_EXTERNAL_DRIVER_PORT._1)
   def externalH2ODriverPortRange = sparkConf.getOption(PROP_EXTERNAL_DRIVER_PORT_RANGE._1)
   def externalExtraMemoryPercent = sparkConf.getInt(PROP_EXTERNAL_EXTRA_MEMORY_PERCENT._1, PROP_EXTERNAL_EXTRA_MEMORY_PERCENT._2)
+  def externalCommunicationBlockSizeAsBytes: Long = sparkConf.getSizeAsBytes(PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._1, PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
+  def externalCommunicationBlockSize: String = sparkConf.get(PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._1, PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
 
   /** Setters */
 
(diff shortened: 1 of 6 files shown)


doc/src/site/sphinx/configuration/configuration_properties.rst
```diff
@@ -328,6 +328,10 @@ External backend configuration properties
 |                                                       |                | specifies memory for internal JVM   |
 |                                                       |                | use outside of Java heap.           |
 +-------------------------------------------------------+----------------+-------------------------------------+
+| ``spark.ext.h2o.external.communication.blockSize``    | ``1m``         | The size of blocks representing     |
```
michalkurka: imho would be good to say what is the unit - that the block size is in bytes

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -60,6 +60,8 @@ trait ExternalBackendConf extends SharedBackendConf {
   def externalH2ODriverPort = sparkConf.getOption(PROP_EXTERNAL_DRIVER_PORT._1)
   def externalH2ODriverPortRange = sparkConf.getOption(PROP_EXTERNAL_DRIVER_PORT_RANGE._1)
   def externalExtraMemoryPercent = sparkConf.getInt(PROP_EXTERNAL_EXTRA_MEMORY_PERCENT._1, PROP_EXTERNAL_EXTRA_MEMORY_PERCENT._2)
+  def externalCommunicationBlockSizeAsBytes: Long = sparkConf.getSizeAsBytes(PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._1, PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
```
michalkurka: why do we need 2 properties with the same content?

michalkurka:
@mn-mikke thanks for adding this quickly!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mn-mikke: `externalCommunicationBlockSizeAsBytes` is utilized in code to get the value. IMHO, it's easier for user to work with a string representation e.g `10m` `10k` plus I would like to be aligned with [similar Spark settings](http://spark.apache.org/docs/latest/configuration.html)

In fact,  I don't need `externalCommunicationBlockSize`. I just added this for debugging purposes to be able to quickly figure out the original string value if it's set via a config file or environment variable.

Should I leave it or remove it? WDYT?

doc/src/site/sphinx/configuration/configuration_properties.rst
mn-mikke: no problem, will add some comment. Thanks!

mn-mikke: Updated the description

PR #1569, opened 2019-10-11
[SW-1669] Improve Variable Names in the ExternalBackendUtils Class

mn-mikke:


mn-mikke:
cc @Tagar

Tagar:
thanks @mn-mikke !

PR #1572, opened 2019-10-13
[SW-1671] Expose Property for Setting Lifetime of MOJOs in Cache

mn-mikke:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOBaseCache.scala
```diff
@@ -28,8 +28,11 @@ trait H2OMOJOBaseCache[B, M] extends Logging {
   private val pipelineCache = mutable.Map.empty[String, B]
   private val lastAccessMap = mutable.Map.empty[String, Long]
 
-  private lazy val sparkConf = SparkSession.builder().getOrCreate().sparkContext.getConf
-  private lazy val cleanupRetryTimeout = sparkConf.getInt("spark.ext.h2o.mojo.destroy.timeout", 10 * 60 * 1000)
+  private lazy val cleanupRetryTimeout = {
+    val sparkConf = SparkSession.builder().getOrCreate().sparkContext.getConf
+    sparkConf.getInt("spark.ext.h2o.mojo.destroy.timeout", 10 * 60 * 1000)
```
mn-mikke: Here we can't use `H2OConf` property since the `sparkling-water-scoring` package doesn't have `sparkling-water-core` as a dependency and I guess that it's made like that on purpose.

jakubhava: yup, no H2OConf in scoring package

PR #1592, opened 2019-11-04
[SW-1607] Create gradle task to create rsparkling docker image for kubernetes

jakubhava:


kubernetes/build.gradle
```diff
@@ -15,10 +15,10 @@ ext {
 
 task buildSparkImages(dependsOn: checkSparkVersionTask) {
     doLast {
-        exec {
-            workingDir sparkHome
-            commandLine getOsSpecificCommandLine(['./bin/docker-image-tool.sh', '-t', "$sparkVersion", "build"])
-        }
+        //exec {
```
mn-mikke: Does it have to be commented out?

jakubhava: nope, good catch, thank you!

jakubhava:
@mn-mikke should be good now

PR #1595, opened 2019-11-05
[SW-1640][Separation] SW should be able to reconnect to a new H2O cluster

jakubhava:


py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -102,7 +102,6 @@ def _get_first(rdd):
 
 class H2OContext(object):
 
-    is_initialized = False
```
jakubhava: We can't have state in Python client -> we need to move this state to scala part in order to allow the re connection.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -481,13 +487,20 @@ object H2OContext extends Logging {
...
-        SharedBackendConf.PROP_RUNNING_FROM_NON_JVM_CLIENT._2)
-      val isExternalBackend = conf.runsInExternalClusterMode
-      if (isExternalBackend && isNonJVMClient) {
-        instantiatedContext.set(new H2OContextRestAPIBased(sparkSession, conf).init())
+    val isNonJVMClient = conf.getBoolean(SharedBackendConf.PROP_RUNNING_FROM_NON_JVM_CLIENT._1,
+      SharedBackendConf.PROP_RUNNING_FROM_NON_JVM_CLIENT._2)
+    val isExternalBackend = conf.runsInExternalClusterMode
+    if (isExternalBackend && isNonJVMClient) {
+      if (instantiatedContext.get() != null) {
+        if (connectingToNewCluster(instantiatedContext.get(), conf)) {
```
mn-mikke: Could happen that the `connectingToNewCluster` method is called when the external cloud is still clouding up?

jakubhava: yup, but I think it is really rare case. After H2O node is started, one of the first things it does is that it starts listening for the incoming requests. So even though the cloud is not yet fully formed, this call will succeed but instead of showing all nodes, just nodes available
 at the time of the call will be printed out. 

Which I think is fine in the manual mode as the use-case there is to have cluster running to which we want to connect

mn-mikke: Ok, perfect. Thanks!

PR #1597, opened 2019-11-05
[SW-1695][Separation] In Rest API based context, we take H2O build info from H2O object, we need to use rest api

jakubhava:
Depends on https://github.com/h2oai/h2o-3/pull/4044/files


jakubhava:
The H2O change is going to H2O 3.26.0.10, after integrating SW with this H2O, we can merge this one

PR #1598, opened 2019-11-06
[SW-1693][Separation] Ensure that we handle cluster non-availabiliy in decent fashion

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -411,10 +411,8 @@ object H2OContext extends Logging {
 
     override protected def getSelfNodeDesc(): Option[NodeDesc] = None
 
-    def getCloudInfo(): CloudV3 = getCloudInfo(conf)
```
jakubhava: should not be exposed

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -68,17 +68,28 @@ trait H2OContextRestAPIUtils extends H2OContextUtils  {
...
-    content
+  private def readURLContent(endpoint: URI, suffix: String): String = {
+    try {
+      val html = Source.fromURL(s"$endpoint/$suffix")
+      val content = html.mkString
+      html.close()
+      content
+    } catch {
+      case cause: Exception =>
+        H2OContext.get().foreach(_.stop())
```
mn-mikke: I'm wondering whether this class is right place for stopping `H2OContext`. This class looks after commonication with a rest endpoint. Do you think we could find a different place for `H2OContext.get().foreach(_.stop())`?  WDYT?

jakubhava: Yes, agree, please have a look now.

Right now we just throw exception if the node is not reachable. The only place where cluster is killed is inside the thread getting heartbeats from the cluster (might be good refactoring to rename this thread a little bit, but I wanted to keep this PR as small as possible)

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -425,9 +422,17 @@ object H2OContext extends Logging {
     }
 
     override protected def getSparklingWaterHeartBeatEvent(): SparklingWaterHeartbeatEvent = {
-      val cloudV3 = getCloudInfo()
-      val memoryInfo = cloudV3.nodes.map(node => (node.ip_port, PrettyPrint.bytes(node.free_mem)))
-      SparklingWaterHeartbeatEvent(cloudV3.cloud_healthy, System.currentTimeMillis(), memoryInfo)
+      try {
```
jakubhava: This is executed in thread which is responsible for checking the hearbeats. Right now the name is a bit confusing, I would like to refactor the thread name to something more explicit like H2OClusterHeartBeatThread but wanted to keep this change as small as possible, will do in next PR

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -213,7 +213,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
     val clusterBuildTimeout = hc.getConf.cloudTimeout
     val nodes = if (runningFromNonJVMClient(hc)) {
-      getNodes(hc._conf)
+      try {
+        getNodes(hc.getConf)
+      } catch {
+        case _: H2OClusterNodeNotReachableException =>
+          val h2oCluster = hc.getConf.h2oCluster.get
+          throw new H2OClusterNodeNotReachableException(
+          s"""
```
mn-mikke: nit: WDYT about reformatting to:
```
s"""External H2O cluster $h2oCluster - ${hc.getConf.cloudName.get} is not reachable, H2OContext has not been created.
   |Please verify that $h2oCluster is running with web enabled and retry the context creation.""".stripMargin
```

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -425,9 +422,17 @@ object H2OContext extends Logging {
...
+        val cloudV3 = getCloudInfo(conf)
+        val memoryInfo = cloudV3.nodes.map(node => (node.ip_port, PrettyPrint.bytes(node.free_mem)))
+        SparklingWaterHeartbeatEvent(cloudV3.cloud_healthy, System.currentTimeMillis(), memoryInfo)
+      } catch {
+        case e: H2OClusterNodeNotReachableException =>
+          H2OContext.get().head.stop()
+          throw new H2OClusterNodeNotReachableException(s"""
+External H2O cluster ${conf.h2oCluster.get} - ${conf.cloudName.get} is not reachable, H2OContext has been closed.
```
mn-mikke: nit: WDYT about reformatting the message?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
mn-mikke: After the client gets separated, will the heartbeat thread still exist?

mn-mikke:
LGTM

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: yes!

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Yup, this heartbeat is checking the worker nodes, not the client. Right now it is used to fill the UI with info obtained from worker nodes & in this separated situation also for stopping the H2OContext if it's not healthy. I think it just deserves a better name.

Does it make sense?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Yup, reformatted to be consistent

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
mn-mikke: Ok thanks for the explanation!

jakubhava:
Thank you so much for review @mn-mikke, very helpful ideas & catches!

PR #1599, opened 2019-11-06
[SW-1696] Remove announcement service as it is not used

jakubhava:
The goal is to simplify code as much as possible as it will help with the transition towards dynamic allocation changes. Plus right now we have one extra thread running with no use.

@bilcus adding you to review as I just want to confirm steam is not using this. This originally became part of the code as long time ago when there was no steam integration this was how steam was supposed to get flow ip and port, but I don't think it is how steam is discovering it now


jakubhava:
Perfect, thank you @bilcus !

PR #1604, opened 2019-11-07
[SW-1536] To be able to serve proxy, ensure that the worker nodes have web enabled

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -213,7 +213,9 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
     logInfo("Connecting to external H2O cluster.")
     val clusterBuildTimeout = hc.getConf.cloudTimeout
     val nodes = if (runningFromNonJVMClient(hc)) {
-      getNodes(hc._conf)
+      val nodes = getNodes(hc._conf)
```
mn-mikke: nit: shouldn't we be consistent and use `hc.getConf` everywhere?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -52,6 +52,22 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
     getNodes(cloudV3)
   }
 
+  def verifyWebOpen(nodes: Array[NodeDesc], conf: H2OConf): Unit = {
+    val nodesWithoutWeb = nodes.flatMap { node =>
+      try {
+        getCloudInfoFromNode(node, conf)
+        None
+      } catch {
+        case _: H2OClusterNodeNotReachableException => Some(node)
```
mn-mikke: If a node doesn't have web enabled, is the exception thrown immediately or there any timeout? I'm wondering about the case when a cluster of hundreds nodes have all web endpoints disabled.

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -52,6 +52,22 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
...
+      }
+    }
+    if (nodesWithoutWeb.nonEmpty) {
+      throw new H2OClusterNodeNotReachableException(
+        s"The following worker nodes are not reachable, but belong to the cluster identified by " +
+          s"${conf.h2oCluster.get}:${nodesWithoutWeb.mkString("\n")}", null)
```
mn-mikke: `.mkString("[", ", ", "]")`?

mn-mikke: Could it happen that `conf.h2oCluster` is `None`?

jakubhava: In this case no, in manual cluster mode the h2oCluster needs to be always defined

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: yup, thanks, good catch!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: `mkString("\n")` is there to print each NodeDesc on new line. What would be output of `.mkString("[", ", ", "]")`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: The exception is thrown right away after we discover the endpoint does not exist -> so I would say almost immediately. 

We could throw exception right away after we discover first node without web enabled and improve the exception message "Please, make sure that the rest of the nodes have web enabled as well".

What I like about printing the full list of non-available nodes is that we have all information at one place.

What do you think?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
mn-mikke: it would formated as an array. `[node1, node2, node3]`. I like the idea of having a node on separete line, but shouldn't we do the same with `conf.h2oCluster.get`?
Example:
```
mycluster
----------
node1
node2
node3
```

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
mn-mikke: I like as it's right now. I just wasn't sure how much time this could take. If it's immediate then it's fine ;-)

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: I see, thanks for the mkString explanation! Yup, good idea, let me fix it

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: Cool, thx!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: Screen Shot 2019-11-07 at 1 21 02 PM

Looks like this at this moment, what do you think?

mn-mikke: Looks good to me!

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -402,7 +404,7 @@ Please verify that $h2oCluster is running with web enabled and retry the context
              |  been just ${ExternalBackendConf.PROP_EXTERNAL_KERBEROS_PRINCIPAL._1}
           """.stripMargin)
       }
-
+
```
mn-mikke: nit: extra line?

jakubhava: yup, good catch!

PR #1608, opened 2019-11-07
[SW-1534] Create proxy for H2O flow as we no longer run H2O on driver side

jakubhava:


jakubhava:
@mn-mikke can you please have a look? Thanks!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+    try {
+      val serverSocket = new ServerSocket()
+
+      serverSocket.setReuseAddress(false);
+      serverSocket.bind(new InetSocketAddress(InetAddress.getByName(SparkEnv.get.blockManager.blockManagerId.host), port), 1)
+      serverSocket.close()
+      true
+    } catch {
+      case _: Exception => false
+    } finally {
```
mn-mikke: nit: extra `finally` statement

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
     }
   }
 
+  private def isTcpPortAvailable(port: Int): Boolean = {
+    try {
+      val serverSocket = new ServerSocket()
+
```
mn-mikke: nit: extra line

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+      serverSocket.setReuseAddress(false);
+      serverSocket.bind(new InetSocketAddress(InetAddress.getByName(SparkEnv.get.blockManager.blockManagerId.host), port), 1)
```
mn-mikke: nit: Would it be possible to split it into more lines due to readability?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+      serverSocket.close()
```
mn-mikke: Shouldn't `serverSocket.close()` be in the `finally` statement?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+      true
+    } catch {
+      case _: Exception => false
+    } finally {
+    }
+  }
+
+  private def findNextFreeFlowPort(conf: H2OConf): Int = {
+    if (conf.clientWebPort == -1) {
+      var port = conf.clientBasePort
```
mn-mikke: Wasn't your intention to implement something like this?
```
var port = conf.clientBasePort + conf.internalPortOffset
while (!isTcpPortAvailable(port)) {
    port = port + 1
}
port

File: core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala [hunk 6]
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+
+    }
+  }
+
+  protected def startFlowProxy(conf: H2OConf): URI = {
+    val port = findNextFreeFlowPort(conf)
+    val server = new Server(port)
+
+    val context = new ServletContextHandler(ServletContextHandler.SESSIONS);
+    context.setContextPath("/");
```
mn-mikke: Shouldn't we use context path from `H2OConf`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,60 @@ private[spark] trait H2OContextUtils extends Logging {
...
+
+    val cloudV3 = H2OContextRestAPIUtils.getCloudInfo(conf)
+    val ipPort = cloudV3.nodes(cloudV3.leader_idx).ip_port
+
+    holder.setInitParameter("proxyTo", s"${conf.getScheme()}://$ipPort")
+    holder.setInitParameter("prefix", "/")
+    context.setServletHandler(handler)
+    server.setHandler(context)
+    server.start()
+    new URI(s"${conf.getScheme()}://${SparkEnv.get.blockManager.blockManagerId.host}:$port")
```
mn-mikke: Shouldn't we include context path here?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: 👍

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Sure thing, great idea!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Yup!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Will be used at the end because of ^

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Thanks for checking!

I had a second look, but still think that the code bellow is the intended one. In H2O, the `internalPortOffset` specifies the offset between the API port (54321) and internal port(54320). Even though we don't run any H2O client here, I wanted to be consistent in port selection. Does it make sense?

mn-mikke: It makes sense what you wrote, but still don't understand why you use `offset` for  a jump if a port is already occupied `port = port + offset + 1`. Why isn't the next one tried if it's free or not?

jakubhava: I see what you're trying to say now! Yes, that is a good point!

jakubhava: After thinking more about it, I simplified it. We do not need to use offset here as we do not have any internal communication port. Therefore we just try all the port starting from the base port

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Not here, but should be specified in the `prefix` parameter

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Yup

jakubhava:
@mn-mikke I think I have applied all the changes, please have a look

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -432,14 +432,20 @@ object H2OContext extends Logging {
         case e: H2OClusterNodeNotReachableException =>
           H2OContext.get().head.stop()
           throw new H2OClusterNodeNotReachableException(
-            s"""External H2O cluster ${conf.h2oCluster.get} - ${conf.cloudName.get} is not reachable, H2OContext has been closed.
+            s"""External H2O cluster ${conf.h2oCluster.get + conf.contextPath.getOrElse("")} - ${conf.cloudName.get} is not reachable,
```
mn-mikke: Just a question: Is a delimiting `/` part of `conf.h2oCluster.get + conf.contextPath.getOrElse("")`?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -51,6 +54,64 @@ private[spark] trait H2OContextUtils extends Logging {
...
+      val socketAddress = new InetSocketAddress(InetAddress.getByName(host), port)
+      serverSocket.bind(socketAddress, 1)
+      true
+    } catch {
+      case _: Exception => false
+    } finally {
+      try {
+        serverSocket.close()
+      } catch {
+        case _: Exception =>
```
mn-mikke: nit: `scala.util.Try` or `scala.util.control.Exception.ignoring`?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Good question, yup, it is 👍

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
jakubhava: Thanks for the hint, I have used `scala.util.Try` & like it!

PR #1609, opened 2019-11-08
[SW-1703][Separation] Rest API endpoints does not work if the external backend enables any kind of authentification

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -33,6 +34,8 @@ class H2OConf(val sparkConf: SparkConf) extends Logging with InternalBackendConf
 
   H2OConf.checkDeprecatedOptions(sparkConf)
 
+  protected var nonJVMClientCreds: Option[FlowCredentials] = None
```
mn-mikke: Note: Doesn't this variable get serialized and sent to each node in internal backend during bootstrap?

jakubhava: Good catch, I think we can make it @transient, what do you think?

mn-mikke: Sounds good to me. I think it would great if we could  to move the variable somewhere else to avoid usage of reflection. Unfortunately, I can't think of a different place.

PR #1613, opened 2019-11-12
[SW-1687] Expose Offset Column in Supervised Algorithms

mn-mikke:
This PR depends on https://github.com/h2oai/h2o-3/pull/4086.


jakubhava:
@mn-mikke as we were discussing with @michalkurka, exposing offsetColumn is one thing, but we also need to modify our mojo scoring, but at this stage I do not understand the steps completely.

But let's do that as part of different PR

jakubhava:
Created JIRA for that ^ https://0xdata.atlassian.net/browse/SW-1707

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -56,6 +58,22 @@ class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOMod
...
+      flatDataFrame: DataFrame,
+      udfConstructor: Array[String] => UserDefinedFunction,
+      inputs: Array[String]): DataFrame = {
+    val relevantColumnNames = flatDataFrame.columns.intersect(inputs)
+    val args = relevantColumnNames.map(c => flatDataFrame(s"`$c`"))
+    val udf = udfConstructor(relevantColumnNames)
+    val predictWrapper = H2OMOJOCache.getMojoBackend(uid, getMojoData, this)
+    predictWrapper.getModelCategory match {
+      case ModelCategory.Binomial | ModelCategory.Regression | ModelCategory.Multinomial =>
+        flatDataFrame.withColumn(outputColumnName, udf(struct(args: _*), lit(0.0)))
```
jakubhava: I think comment would help for `lit(0.0)`

core/build.gradle
```diff
@@ -30,6 +30,8 @@ dependencies {
 
     // Generic model support
     compile("ai.h2o:h2o-genmodel:${h2oVersion}")
+    // H2O security
+    compile("ai.h2o:h2o-security:${h2oVersion}")
```
jakubhava: This should be added as part of different PR after upgrade to new H2O :)

mn-mikke: Yup, no problem. I just wanted to make it running against the current version of the `rel-yau` branch.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mn-mikke: Ok, will add something.

core/build.gradle
jakubhava: Yup

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
jakubhava: thank you!

core/build.gradle
michalkurka: @mn-mikke since this is an issue for you right now - should there be a dependency in h2o-core on h2o-security even though it doesn't actually need that

I want to avoid issues not just for SW but for other people building H2O as well

jakubhava: @michalkurka was thinking about this, on Sparkling Water side we simply fix it, but as you mentioned, it can affect other folks. I think it's actually a good idea to mark it as a dependency for maven purposes

jakubhava: But it's also a little bit hack to create such artificial dependency, not sure

mn-mikke: IMHO, there are very few customers like us referencing individual h2o projects. The vast majority will assembly jar. Maybe, it would be safer to such a change in a major release.

ml/src/test/scala/org/apache/spark/ml/spark/models/H2OSupervisedMOJOModelTestSuite.scala
```diff
@@ -0,0 +1,168 @@
...
+  test("Offset column gets propagated to MOJO model - GBM") {
+    testOffsetColumnGetsPropagatedToMOJOModel(new H2OGBM())
+  }
+
+  test("Offset column gets propagated to MOJO model - GLM") {
+    testOffsetColumnGetsPropagatedToMOJOModel(new H2OGLM())
+  }
+
+  test("Offset column gets propagated to MOJO model - DeepLearning") {
+    testOffsetColumnGetsPropagatedToMOJOModel(new H2ODeepLearning())
```
jakubhava: Is Model offset available also on XGBoost?

mn-mikke: It needs to be implemented on the H2O-3 side first. https://0xdata.atlassian.net/browse/PUBDEV-7029

PR #1614, opened 2019-11-12
[SW-1708] Simplify distribution of security files

jakubhava:
This proposal introduce a small cleanup. I thing we can keep one code path for all the cases using this solution.

What do you think? I have missed this during the review.

Maybe there is a reason for doing it like now, I just not see it :)


mn-mikke:
@jakubhava I think this is a great improvement! It seems I got lost in various scenarios. In that case I think you can remove the third parameter from property specification. E.g: https://github.com/h2oai/sparkling-water/pull/1614/files#diff-ed060ffa1e96e1fe4dc30499d7cc9a54L263

jakubhava:
Thanks @mn-mikke! I started playing with SSL communication and got this idea.

mn-mikke:
Sorry, I should have mentioned that after the third item from the property spec. The return time of [this](https://github.com/h2oai/sparkling-water/pull/1614/commits/4b4eeca0365509fb1728f3776893182e7ddba1e1#diff-ed060ffa1e96e1fe4dc30499d7cc9a54L204) needs to be modified.

jakubhava:
Ah, good catch, let me fix it

jakubhava:
Fixed, sorry, my bad

PR #1616, opened 2019-11-14
[SW-1711] Fix propagation of internal security conf in Sparkling Water

jakubhava:
Depends on https://github.com/h2oai/h2o-3/pull/4073

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
@@ -22,6 +22,8 @@ import java.io.File
 import org.apache.spark.h2o.H2OConf
 import org.apache.spark.h2o.utils.AzureDatabricksUtils
 import org.apache.spark.internal.Logging
+import org.apache.spark.network.Security
+import org.apache.spark.sql.SparkSession
 import org.apache.spark.util.Utils
 import org.apache.spark.{SparkContext, SparkEnv, SparkFiles}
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -50,14 +52,16 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Exter
...
     conf.sslConf match {
-      case Some(ssl) =>
-        val sslConfig = new Properties()
-        sslConfig.load(new FileInputStream(ssl))
-        cmdToLaunch = cmdToLaunch ++ Array("-files", sslConfig.get("h2o_ssl_jks_internal") + "," + sslConfig.get("h2o_ssl_jts"))
-        cmdToLaunch = cmdToLaunch ++ Array("-internal_security", ssl)
-        logInfo(s"Running external H2O cluster in encrypted mode with config: $ssl")
+      case Some(internalSecurityConf) =>
+      // In External Backend, auto mode we need distribute the keystore files to the H2O cluster
+      val props = new Properties()
```
mn-mikke: nit: indentation

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -60,6 +63,14 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
       AzureDatabricksUtils.setClientCheckRetryTimeout(conf)
     }
 
+    if (conf.isInternalSecureConnectionsEnabled) {
```
mn-mikke: Just a question: Is this executed only on a client? If so, can you point me to the place where it's set for workers?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
jakubhava: thanks!

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
jakubhava: Yup, it is executed only on client -> we create the configuration object on the client and then distribute the files to spark

PR #1619, opened 2019-11-15
[SW-1713] Refactor internal & external backend utils to have the same structure

jakubhava:
-> Better `conf` on H2OContext -> fixes bug where we were setting wrong conf instance on ExternalH2OBackend which was leading to error when obtaining cluster details via rest api as the clusterEndpoint was empty even though set before
-> Avoid unnecessary complexity in classes related to backends, simplify and unify them

This change does not introduce any breaking behaviour

The tests for rest api solution still fails due to missing conversions but fails as expected. Before this change we failed with exception whilst getting connection to the cluster

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -60,7 +60,7 @@ import scala.util.control.NoStackTrace
   * @param sparkSession Spark Session
   * @param conf         H2O configuration
   */
-abstract class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends Logging with H2OContextUtils {
+abstract class H2OContext private(val sparkSession: SparkSession, private val conf: H2OConf) extends Logging with H2OContextUtils {
   self =>
   val sparkContext = sparkSession.sparkContext
   val uiUpdateThread = new Thread {
@@ -92,13 +92,6 @@ abstract class H2OContext private(val sparkSession: SparkSession, conf: H2OConf)
     new InternalH2OBackend(this)
   }
 
-
-  // Check Spark and H2O environment for general arguments independent on backend used and
-  // also with regards to used backend and store the fix the state of prepared configuration
-  // so it can't be changed anymore
-  /** H2O and Spark configuration */
-  val _conf: H2OConf = backend.checkAndUpdateConf(conf).clone()
-
   protected def getFlowIp(): String
 
   protected def getFlowPort(): Int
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -17,18 +17,13 @@
...
 import org.apache.spark.h2o.H2OConf
 import org.apache.spark.h2o.backends.{ArgumentBuilder, SharedBackendUtils}
-import org.apache.spark.h2o.utils.NodeDesc
 import water.{ExternalFrameUtils, H2O, Paxos}
 
-/**
-  * Various helper methods used in the external backend
-  */
-private[external] trait ExternalBackendUtils extends SharedBackendUtils {
+private[backends] trait ExternalBackendUtils extends SharedBackendUtils {
```
jakubhava: ExternalBackendUtils, InternalBackendUtils & SharedBackendUtils, turned just into traits.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -49,42 +44,12 @@ private[external] trait ExternalBackendUtils extends SharedBackendUtils {
...
-    * In order to check the configuration for specific backend, method checkAndUpdateConf on particular backend has to be
-    * called.
-    *
-    * This method has to be called at the start of each method which override this one
-    *
-    * @param conf H2O Configuration to check
-    * @return checked and updated configuration
-    * */
-  override def checkAndUpdateConf(conf: H2OConf): H2OConf = {
-    super.checkAndUpdateConf(conf)
```
jakubhava: Move to checkAndUpdateConf on ExternalH2OBackend, this was unnecessary override. We just need to override it on ExternalH2OBackend and call super to ensure we call the method in SharedBackendUtils.

Same holds for InternalH2OBackend

jakubhava:
Just verified, after this change, the bug with H2OConf is fixed and the rest api test based now fail during conversions, not during connections to backend, which is expected

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -491,6 +484,14 @@ object H2OContext extends Logging {
     val sameNodes = hc.getH2ONodes().map(_.ipPort()).sameElements(newCloudV3.nodes.map(_.ip_port))
     !sameNodes
   }
+
+  private def checkConf(conf: H2OConf): H2OConf = {
```
mn-mikke: nit: Shouldn't the method name correspond to methods that are called. `checkAndUpdateConf`

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -44,63 +45,68 @@ class InternalH2OBackend(@transient val hc: H2OContext) extends SparklingBackend
...
+    conf.set("spark.locality.wait", "0")
+
+    if (conf.clientIp.isEmpty) {
+      conf.setClientIp(getHostname(SparkEnv.get))
+    }
+
     conf.getOption("spark.executor.instances").foreach(v => conf.set("spark.ext.h2o.cluster.size", v))
 
-    if (!conf.contains("spark.scheduler.minRegisteredResourcesRatio") && !hc.sparkContext.isLocal) {
+    if (!conf.contains("spark.scheduler.minRegisteredResourcesRatio") && SparkSession.builder().getOrCreate().sparkContext.isLocal) {
```
mn-mikke: you removed `!`. Is it on purpose?

jakubhava: Great catch! No :) Will fix

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Yup, like the proposal, thank you!

PR #1620, opened 2019-11-15
[SW-1710] Enable client mode in Sparkling Water

jakubhava:
Depends on release of H2O 3.26.0.11 with this change https://github.com/h2oai/h2o-3/pull/4056/files


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -142,6 +142,7 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
     */
   def getH2OCommonArgs(conf: H2OConf): Seq[String] = {
     new ArgumentBuilder()
+      .add("-allow_clients", "true")
```
mn-mikke: Just a question: This argument is specified in `getH2OCommonArgs()`. Do we also need it for a client?

jakubhava: Good point -> It won't hurt passing this parameter to clients, but not required. I moved the option just to worker nodes

PR #1622, opened 2019-11-17
[SW-1717] Correctness Tests for Usage of 'offsetCol' with H2OGBM

mn-mikke:
This PR depends on https://github.com/h2oai/h2o-3/pull/4086 and #1613.


py/tests/unit/with_runtime_sparkling/test_gbm.py
```diff
@@ -44,3 +50,81 @@ def testLoadAndTrainMojo(prostateDataset):
     assert len(predMojo) == len(predModel)
     for i in range(0, len(predMojo)):
         assert predMojo[i] == predModel[i]
+
+
+@pytest.fixture(scope="module")
+def dataset(spark):
+    return spark \
+        .read.csv("file://" + unit_test_utils.locate("smalldata/insurance.csv"), header=True, inferSchema=True) \
+        .withColumn("Offset", log(col("Holders")))
```
jakubhava: Just question: In H2O examples I see that they usually create offset column from the response column, what was the reason to create it from `Holders` col? Just interested :)

py/tests/unit/with_runtime_sparkling/test_gbm.py
```diff
@@ -44,3 +50,81 @@ def testLoadAndTrainMojo(prostateDataset):
...
+def testMOJOModelReturnsExpectedResultWhenOffsetColumnsIsSet(gbmModelWithOffset, dataset):
+    predictionCol = col("prediction")
+    predictionsDF = gbmModelWithOffset.transform(dataset)
+    detailsDF = predictionsDF.select(min(predictionCol).alias("min"),
+                                     max(predictionCol).alias("max"),
+                                     mean(predictionCol).alias("mean"))
+    result = detailsDF.first()
+    modelDetials = json.loads(gbmModelWithOffset.getModelDetails())
+
+    assert abs(-1.869702 - modelDetials['init_f']) < 1e-5, "Expected init_f to be {0}, but got {1}". \
```
jakubhava: And other questions, just out of curiosity -> what init_f represents?

py/tests/unit/with_runtime_sparkling/test_gbm.py
mn-mikke: I've got inspired in this test https://github.com/h2oai/h2o-3/blob/master/h2o-py/tests/testdir_algos/gbm/pyunit_offset_tweedie_gbm.py#L14 testing correctness of a binary model.

jakubhava: I see, thanks!

py/tests/unit/with_runtime_sparkling/test_gbm.py
mn-mikke: http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/gbm.html#gbm-algorithm

`fk0` is beeing initialized to some constant, so I assume `init_f` represents the constant.

PR #1628, opened 2019-11-19
[SW-1721] Do not stop the pyspark shell in case the app is stopped in REST API based approach, just stop the context

jakubhava:


py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -237,18 +239,21 @@ def __stop(self):
 
     def stop(self):
         self.__stop()
-        sys.exit()
+        self.__isStopped = True
+        # do this ugly stop only in case of non-rest api client
+        if self._conf.get("spark.ext.h2o.rest.api.based.client") == "false":
```
mn-mikke: Would it be possible to access property via a getter?

jakubhava: I was thinking about it, but I do not want to expose getter for such a property on purpose, would like to keep it internal

PR #1635, opened 2019-11-25
[SW-1730] Ensure worker nodes have always open Flow in the external Backend

jakubhava:


jakubhava:
Thanks! I have based this PR on SW-1731 as that one is doing the deprecation, going to rel & master, this is removing the deprecation and changing the behavior, but only in major

PR #1636, opened 2019-11-25
[SW-1731] Deprecate h2oNodeWebEnabled for external backend

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -204,6 +213,12 @@ trait SharedBackendConf {
 
   private[backends] def getFileProperties(): Seq[(String, _)] = Seq(PROP_JKS, PROP_LOGIN_CONF)
 
+  private def logWarningH2ONodeWebEnabledDeprecated(): Unit = {
+    if (runsInExternalClusterMode) {
```
mn-mikke: Shouldn't this hold only if the client is disabled?

jakubhava: nope, it's on purpose like this, preparing the code for the future steps

PR #1637, opened 2019-11-26
[SW-1733] Tests for Clientless Conversion from H2OFrames to DataFrames

mn-mikke:


mn-mikke:
@jakubhava Isn't this sufficient?

jakubhava:
When you run `./gradle test` yes, but we don't run tests like this in our pipeline. We just need to add right call after this line https://github.com/h2oai/sparkling-water/blob/aaedd584060851acd950714798b85df62c46e820/jenkins/sparklingWaterPipeline.groovy#L298

PR #1639, opened 2019-11-26
[SW-1732] Implement shutdown hook to ensure H2O will go down on normal stop of Spark

jakubhava:
Depends on https://github.com/h2oai/h2o-3/pull/4110 (H2O 3.26.0.12)

No need for hacky stopping, tested on YARN in both internal & external backend

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -108,6 +108,8 @@ abstract class H2OContext private(val sparkSession: SparkSession, private val co
     * otherwise it creates new H2O cluster living in Spark
     */
   def init(): H2OContext = {
+    // The lowest priority used by Spark is 25 (removing temp dirs). We need to perform cleaning up of H2O
+    // resources before Spark does as we run as embedded application inside the Spark
     ShutdownHookManager.addShutdownHook(10) { () =>
       logWarning("Spark shutdown hook called, stopping H2OContext!")
       stop(stopSparkContext = false, stopJvm = false)
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -30,6 +30,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql.{DataFrame, SparkSession}
 import water._
 import water.util.PrettyPrint
+import org.apache.spark.util.ShutdownHookManager
 
 import scala.language.{implicitConversions, postfixOps}
 import scala.reflect.ClassTag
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -92,6 +93,10 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
     * otherwise it creates new H2O cluster living in Spark
     */
   def init(): H2OContext = {
+    ShutdownHookManager.addShutdownHook(10)(() => {
```
mn-mikke: Would it be possible to add some comment explaining priority `10`?

mn-mikke: nit: WDYT about:
```
    ShutdownHookManager.addShutdownHook(10) { () =>
      logWarning("Spark shutdown hook called, stopping H2OContext!")
      stop(stopSparkContext = false, stopJvm = false)
    }
```

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -282,21 +287,34 @@ class H2OContext private(val sparkSession: SparkSession, conf: H2OConf) extends
...
       logWarning("H2OContext is already stopped, this call has no effect anymore")
     }
   }
 
+  /** Stops H2O context.
+    *
+    * @param stopSparkContext stop also spark context
+    */
+  def stop(stopSparkContext: Boolean = false): Unit = {
+    stop(stopSparkContext, stopJvm = true)
```
mn-mikke: Shouldn't we let Spark to stop JVM?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: Sure, I need to provide comment in the code. We need to have higher ( lower number) priority then spark shutdown hooks so we stop H2O first and then let Spark stop itself

jakubhava: I will make a constant for it with explanation

jakubhava: 👍

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
jakubhava: In ideal case yes :) However the current stop method worked in a way that it always tried to stop JVM.

Once we have the client-less approach we can get happily rid of this, however I did not want to change behaviour in this release.

mn-mikke: ok, np.

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -107,6 +109,10 @@ abstract class H2OContext private(val sparkSession: SparkSession, private val co
     * otherwise it creates new H2O cluster living in Spark
     */
   def init(): H2OContext = {
+    ShutdownHookManager.addShutdownHook(10) { () =>
```
jakubhava: Todo: explain why 10

jakubhava:
@mn-mikke please have a look if it looks better now

PR #1642, opened 2019-11-27
[SW-1738] Ensure benchmarks are not run as part of regular build

jakubhava:


mn-mikke:
I'm afraid that this will break compilation of  the benchmark project. Wouldn't help to just remove benchmarks from [here](https://github.com/h2oai/sparkling-water/blob/master/build.gradle#L69)?

benchmarks/build.gradle
```diff
@@ -107,7 +107,7 @@ task cleanOutput(type: Delete) {
     delete "output"
 }
```
jakubhava: shadowJar.finalizedBy substituteTerraform

benchmarks/build.gradle
```diff
@@ -122,6 +122,3 @@ task runBenchmarks(dependsOn: [build, cleanOutput]) {
         }
     }
 }
-
```
jakubhava: substituteTerraform.dependsOn build

PR #1643, opened 2019-11-28
[SW-1739] Fix target encoder multiline doc descriptions

jakubhava:
Not causing any problems, but we are adding multiple args instead of one documentation. Should be wrapped in `"""`

### DIFF
--- a/py/src/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.py
+++ b/py/src/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.py
@@ -51,26 +51,26 @@ class H2OTargetEncoderParams(Params):
         Params._dummy(),
         "holdoutStrategy",
         """A strategy deciding what records will be excluded when calculating the target average on the training dataset.
-           Options:
-            None        - All rows are considered for the calculation
-            LeaveOneOut - All rows except the row the calculation is made for
-            KFold       - Only out-of-fold data is considered (The option requires foldCol to be set.""",
+Options:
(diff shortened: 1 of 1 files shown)


py/src/ai/h2o/sparkling/ml/params/H2OTargetEncoderParams.py
```diff
@@ -61,24 +60,24 @@ class H2OTargetEncoderParams(Params):
     blendedAvgEnabled = Param(
         Params._dummy(),
         "blendedAvgEnabled",
-        "If set, the target average becomes a weighted average of the posterior average for a given "
-        "categorical level and the prior average of the target. The weight is determined by the size "
-        "of the given group that the row belongs to. By default, the blended average is disabled.",
+        """If set, the target average becomes a weighted average of the posterior average for a given
+           categorical level and the prior average of the target. The weight is determined by the size
```
mn-mikke: Wouldn't the indentation be included within the text?

jakubhava: yes, it probably will, I was following same indentation pattern as above ( keep align to first sentence)

jakubhava:
Oki, spaces will be there. Let me do a quick fix for now. I had a look and there is not a method similar to stripMargin as in scala, extra dependency needs to be done for this. I did this a quick fix without introducing a new dependency

jakubhava:
Created this JIRA https://0xdata.atlassian.net/browse/SW-1742 We can have a look when we have some free time :)

PR #1646, opened 2019-11-28
[SW-1722] Ensure external backend (rest api) is stopped in automatic …

jakubhava:
…mode if the spark app is killed ( avoid zombie clusters)


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -362,7 +362,8 @@ object SharedBackendConf {
     */
   val PROP_CLIENT_FLOW_BASEURL_OVERRIDE = ("spark.ext.h2o.client.flow.baseurl.override", None)
 
-  /** Timeout in milliseconds specifying how often we check whether the client is still connected */
+  /** Timeout in milliseconds specifying how often the H2O backend checks whether the Sparkling Water
+    * client (either H2O client or REST) is connected */
```
mn-mikke: nit: `*/` to new line

jakubhava: Good catch, thanks!

PR #1647, opened 2019-11-28
[SW-1647] Convert PySpark DataFrame to H2OFrame without Client

mn-mikke:


core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
```diff
@@ -56,69 +56,139 @@ object WriteConverterCtxUtils {
...
+      }
+
+      writerClient.finalizeFrame(keyName, res, types)
+      keyName
     }
 
-   writerClient.finalizeFrame(keyName, res, types)
   }
 
+  object RESTBasedConverter extends Converter {
```
jakubhava: I would rename this to `RestBasedExternalConvertor`, because:

I think that we at some point will get rid of the client, but I would still distinguish between internal & external rest API convertors. The external sends the data, the internal will just notify the right executor/h2o node via rest api that it should convert one partition to H2O chunk. But that is for the future.

What do you think?

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -190,6 +204,17 @@ abstract class H2OContext private(val sparkSession: SparkSession, private val co
 
   def asH2OFrame(df: DataFrame, frameName: String): H2OFrame = asH2OFrame(df, Option(frameName))
 
+  /**
+    * The method transforms RDD to H2OFrame and returns String representation of its key.
```
jakubhava: transforms RD -> transforms dataframe

mn-mikke: perfect! Good catch!

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
mn-mikke: No problem, will do ;-)

py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -290,7 +244,7 @@ def as_spark_frame(self, h2o_frame, copy_metadata=True):
             df._h2o_frame = h2o_frame
             return df
 
-    def as_h2o_frame(self, dataframe, framename=None, full_cols=100):
+    def as_h2o_frame(self, dataframe, framename=None, full_cols=-1):
```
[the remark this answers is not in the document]
mn-mikke: I've made this change for two reasons:

- In rest mode, we don't have a java H2OFrame and we can't call `.numCols()` https://github.com/h2oai/sparkling-water/pull/1647/files#diff-9143aa3a0fb6ae5a60db472c5edacf9dL68
- -1 is default value for other methods of python client. http://docs.h2o.ai/h2o/latest-stable/h2o-py/docs/frame.html#h2o.frame.H2OFrame.get_frame I've thought it would be nice if we were consistent accross Python APIs.

If you think it's crucial to have default value `100`, i could try to progate number of columns from REST objects.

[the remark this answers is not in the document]
jakubhava: I like to be consistent with the H2O. I think we have limited this to 100 in Sparkling Water case that loading really big frames with a lot of columns was really slow. But I think we are now using light endpoint which should be fine

PR #1654, opened 2019-11-29
[SW-1746] spark.ext.h2o.external.disable.version.check was false also when we needed to run rest api tests

jakubhava:


py/build.gradle
```diff
@@ -315,7 +317,7 @@ task testPythonMojoPipeline(type: PythonTask, dependsOn: distPython) {
 }
 
 task testPythonREST(type: PythonTask, dependsOn: distPython) {
-    extraArgs(*createUnitTestArgs())
+    extraArgs(*createUnitTestRESTApiArgs())
```
mn-mikke: I think this is not necessary, the option is already defined here: https://github.com/h2oai/sparkling-water/blob/master/py/tests/unit/with_runtime_clientless_sparkling/conftest.py#L36

jakubhava: Good catch, thank you!

PR #1655, opened 2019-11-29
[SW-1747] The option "-sw_ext_backend" must be enabled when REST-based client is used

mn-mikke:


jakubhava:
Thank you! It is required because the internal communication is still via REST, right?

PR #1656, opened 2019-11-29
[SW-1748] Try to fix NPE on Spark 2.1, 2.2 and 2.3 related to metadata

jakubhava:


jakubhava:
Good idea, empty array might have different meaning for these things in H2O then null - we would need to check. 

If you're ok with this, I would keep this path ( if the tests pass)

PR #1657, opened 2019-11-29
[SW-1749] Percentiles Are Not Propagated to Metadata

mn-mikke:


core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -86,7 +86,7 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
 
   def getFrame(conf: H2OConf, frameId: String): H2OFrame = {
     val endpoint = getClusterEndpoint(conf)
-    val frames = query[FramesV3](endpoint, s"3/Frames/$frameId/light", conf)
+    val frames = query[FramesV3](endpoint, s"3/Frames/$frameId/summary?row_count=0", conf)
```
jakubhava: 👍

jakubhava:
External tests failing

mn-mikke:
@jakubhava Thanks for letting me know. I've already experienced this problem with `Gson` library. I think the fix would be to try to use  json4s or something else.

mn-mikke:
@jakubhava Can you have a second look?

PR #1659, opened 2019-12-03
[SW-1744] Tests for Clientless Conversion from DataFrames to H2OFrames

mn-mikke:


py/tests/conftest.py
```diff
@@ -21,11 +21,11 @@ def pytest_addoption(parser):
     parser.addoption("--dist", action="store", default="")
     parser.addoption("--spark_conf", action="store", default="")
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
```
jakubhava: As we discussed, we need to make sure that in client-less approach we can connect to a new backenda

mn-mikke: Yup, will cover in separated tests

PR #1660, opened 2019-12-04
[SW-1737] Ensure the client has full flatfile in external backend

jakubhava:
Depends on https://github.com/h2oai/h2o-3/pull/4111

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
@@ -66,7 +66,7 @@ trait ExternalBackendConf extends SharedBackendConf {
   def externalCommunicationBlockSize: String = sparkConf.get(PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._1, PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
   def externalBackendStopTimeout: Int = sparkConf.getInt(PROP_EXTERNAL_BACKEND_STOP_TIMEOUT._1, PROP_EXTERNAL_BACKEND_STOP_TIMEOUT._2)
   private[backends] def isBackendVersionCheckDisabled() = sparkConf.getBoolean(PROP_EXTERNAL_DISABLE_VERSION_CHECK._1, PROP_EXTERNAL_DISABLE_VERSION_CHECK._2)
-
+  private[backends] def externalBackendFlatfileName() = sparkConf.getOption(PROP_EXTERNAL_FLATFILE_NAME._1)
   /** Setters */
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -65,7 +65,7 @@ trait ExternalBackendConf extends SharedBackendConf {
     PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
   def externalCommunicationBlockSize: String = sparkConf.get(PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._1, PROP_EXTERNAL_COMMUNICATION_BLOCK_SIZE._2)
   private[backends] def isBackendVersionCheckDisabled() = sparkConf.getBoolean(PROP_EXTERNAL_DISABLE_VERSION_CHECK._1, PROP_EXTERNAL_DISABLE_VERSION_CHECK._2)
-
+  private[backends] def externalBackendFlatfileName() = sparkConf.getOption(PROP_EXTERNAL_FLATFILE_NAME._1)
```
mn-mikke: nit: `FlatfileName` -> `FlatFileName`?

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
```diff
@@ -233,4 +234,7 @@ object ExternalBackendConf {
 
   /** Disable version check of external H2O backend */
   val PROP_EXTERNAL_DISABLE_VERSION_CHECK = ("spark.ext.h2o.external.disable.version.check", false)
+
+  /** Name of file where flatfile is stored in case of automatic mode of external backend */
+  val PROP_EXTERNAL_FLATFILE_NAME = ("spark.ext.h2o.external.flatfile.name", None)
```
mn-mikke: Just a question: Doesn't the preporty describe a relative path of a flat file and not just name?

jakubhava: Nope, just a name

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
jakubhava: 👍

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendConf.scala
mn-mikke: Ok, thanks!

PR #1661, opened 2019-12-04
[SW-1755] Don't need to stop worker nodes in internal backend, spark takes care of it as it shutdowns the executors

jakubhava:


py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -196,7 +196,7 @@ def getOrCreate(spark, conf=None, verbose=True, pre_create_hook=None, h2o_connec
 
     def stop(self):
```
jakubhava: this is used only for explicit stop by the user, if the user does not call this stop() method in pysparkling, the shutdown hook is called and stop() method on scala side is invoked

doc/src/site/sphinx/configuration/configuration_properties.rst
```diff
@@ -364,6 +364,14 @@ External backend configuration properties
 |                                                       |                | suffix "k", "m" or "g" (e.g.        |
 |                                                       |                | ``450k``, ``3m``)                   |
 +-------------------------------------------------------+----------------+-------------------------------------+
+| ``sspark.ext.h2o.external.backend.stop.timeout``      | ``10000ms``    | How long we wait before for proper  |
```
mn-mikke: typo: `sspark.ext.h2o.external.backend.stop.timeout`

mn-mikke: `How long we wait before for proper external backend before we give up on confirmation from worker nodes.` Would it be possible to rephrase the sentence?

jakubhava: 👍

jakubhava: Yup, done

PR #1664, opened 2019-12-04
[SW-1758] Reuse stopped field from scala backend

jakubhava:
All the information about state should be stored on the backend. The stopped field is already available on the backend & this PR makes use of it

### DIFF
--- a/py/src/ai/h2o/sparkling/H2OContext.py
+++ b/py/src/ai/h2o/sparkling/H2OContext.py
@@ -75,8 +75,6 @@ class H2OContext(object):
         self._jsql_context = self._spark_session._jwrapped
         self._jspark_session = self._spark_session._jsparkSession
         self._jvm = self._spark_session._jvm
-        self.__isStopped = None
-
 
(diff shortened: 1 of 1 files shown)


py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -162,6 +159,22 @@ def __setCreds(conf, auth):
...
+
+    def __setStopped(self):
+        hc = self._jhc.h2oContext()
+        field = self.__getStoppedField()
+        field.set(hc, True)
+
+    def __getStoppedField(self):
+        hc = self._jhc.h2oContext()
+        field = hc.getClass().getSuperclass().getDeclaredField("stopped")
+        field.setAccessible(True)
```
mn-mikke: Just a comment I don't think it's crucial: Shouldn't `__isStopped` and `__setStopped` call `field.setAccessible(False)` when a given operation is finished?

jakubhava: I don't think it's necessary as we don't store the reference to the field, but good catch!

PR #1669, opened 2019-12-05
[SW-1754] H2OContext.getOrCreate Should Create Only One Cluster in Automatic Mode

mn-mikke:


h2o-ops:
Hi @mn-mikke, your PR title "H2OContext.getOrCreate Should Create Only One Cluster in Automatic Mode" is missing JIRA issue number! Please specify it in the following form [SW-]

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -575,8 +575,10 @@ object H2OContext extends Logging {
       SharedBackendConf.PROP_REST_API_BASED_CLIENT._2)
     val isExternalBackend = conf.runsInExternalClusterMode
     if (isExternalBackend && isRestApiBasedClient) {
-      if (instantiatedContext.get() != null) {
-        if (connectingToNewCluster(instantiatedContext.get(), checkedConf)) {
+      val existingContext = instantiatedContext.get()
```
jakubhava: This would definitely work, but I'm curios why the condition in the `connectingToNewCluster(existingContext, checkedConf)` failed.

The nodes should be the same when connecting to cluster created in automatic mode shouldn't it?

mn-mikke: If the `getOrCreate` method is called for the second time with a different instance of H2OConf, information about h2o cluster might be missing, since it's set once [here](https://github.com/h2oai/sparkling-water/blob/2f916d1e80e598a93f90c78d7d46715aca309605/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala#L211). The second call than fails on [this line](https://github.com/h2oai/sparkling-water/blob/282524b96d7972471a8fb79a54634f609d77c222/core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala#L156).

jakubhava: I see, very good catch!

mn-mikke:
> We can introduce API for starting multiple clusters in automatic mode later
In this case, IMHO we will need to also think about how to track all reference to all cluster and also how gracefully kill them when are not needed anymore.

jakubhava:
Yup, this is really question of a new API, it will also open up questions about the watchdog clients etc. I will just create a JIRA for it

PR #1675, opened 2019-12-06
[SW-1768] Get Default Values of AutoML Parameters Synchronized with H2O-3

mn-mikke:
I was also thinking about referencing H2O-3 default values directly to have them specified only once. But since they must be specified also in Python API, I just changed values on the SW site.


mn-mikke:
@jakubhava According what I remember from the discussion with @seb-h2o, it's exactly as you described. `maxModels -> 0` also means an unlimited number of models. I will add note to a migration guide

 @seb-h2o please correct me if I'm wrong.

jakubhava:
Thanks for explanation!

seb-h2o:
to clarify:
- `maxRuntimeSecs = 0` means unlimited time.
- `maxModels = 0` means unlimited amount of models.
- if both are set to 0, backend will automatically change the default for `maxRuntimeSecs` to `1h`.
- if both are set to value `!= 0`, both are applied and the first limit reached will prevail.

PR #1676, opened 2019-12-06
[SW-1769] Avoid Usage of AutoML Deprecations

mn-mikke:


mn-mikke:
@jakubhava it seems that sort metric is still [defined as a string](https://github.com/h2oai/h2o-3/blob/d927ff5776ef7cdb1e20555a064b825de20e678a/h2o-automl/src/main/java/ai/h2o/automl/AutoMLBuildSpec.java#L179) And also the default value is gotten from `StoppingMetric`. @seb-h2o This sounds like a room for improvement. Thanks @jakubhava for pointing that out!

jakubhava:
No worries. Would be good to move this enum to core AutoML at some point :)

seb-h2o:
the enum for sorting metric is not implemented yet in AutoML because metrics names in H2O-3 are a bit of a mess. We're working with @erin towards standardizing those, and ideally I would like to be able to reuse `Scorekeeper.StoppingMetric` but there are still some mismatch currently.
We should be able to allow this soon though: https://0xdata.atlassian.net/browse/PUBDEV-6701

mn-mikke:
Thanks a lot @seb-h2o for clarification!

PR #1678, opened 2019-12-06
[SW-1770] Expose Monotone Constraints

mn-mikke:


py/src/ai/h2o/sparkling/ml/params/H2OTypeConverters.py
```diff
@@ -268,3 +268,25 @@ def convert(value):
...
+                map = {}
+                for k in keys:
+                    map[k] = [TypeConverters.toFloat(v) for v in value.get(k)]
+                return map
+            elif isinstance(value, dict):
+                keys = value.keys()
+                map = {}
+                for k in keys:
+                    map[k] = TypeConverters.toFloat(value[k])
+                return map
```
seb-h2o: in both case you can also generate dictionaries using comprehensions if you want, e.g.:
```
return {k: [TypeConverters.toFloat(v) for v in value.get(k)] 
             for k in in value.keySet().toArray()}
```

mn-mikke: @seb-h2o Good idea! I've tried to address it, please take a look ;-)

mn-mikke:
@jakubhava I've made some change to the PR, can you have a second look?

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasMonotoneConstraints.scala
```diff
@@ -0,0 +1,49 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package ai.h2o.sparkling.ml.params
+
+import java.util
+
+import hex.KeyValue
```
jakubhava: Do we wan to expose H2O structure in this API?

mn-mikke: `KeyValue` is used only by (this method)[https://github.com/h2oai/sparkling-water/pull/1678/files/beea788bee0125e586d15eb3e2e0c13a903a2b45#diff-dcdd1785c6fc00ccd365f4e5dce628afR39], which serves only for internal usage. The result of this method is immediately passed to H2O-3 internals. E.g.: [``XGBoost``](https://github.com/h2oai/sparkling-water/pull/1678/files/beea788bee0125e586d15eb3e2e0c13a903a2b45#diff-e98c0636636d3757ad23131bcad2b095R369)

jakubhava: 👍

py/src/ai/h2o/sparkling/ml/params/H2OTypeConverters.py
seb-h2o: I don't know the possibilities of JavaObject , but with `dict` you can compress even more:
```
{k: TypeConverters.toFloat(v) for k, v  in value.items()}
```

seb-h2o: with `value` being a `JavaObject`, it's likely (however I didn't check) that it would work this way:
```{k: TypeConverters.toFloat(value[k]) for k, v in value.__dict__.items()}```

seb-h2o: checked `JavaObject`: forget last one

mn-mikke: `JavaObject` actually represents `java.util.Map[String, Double]` that is converted to python, so the api of that object should be identical

Thanks for the hint with `{k: TypeConverters.toFloat(v) for k, v  in value.items()}`!

PR #1680, opened 2019-12-09
[SW-1774] Fix bug introduced in 3.26.11 - using flatfile for client connection in external backend

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala
```diff
@@ -47,11 +47,17 @@ private[backends] trait ExternalBackendUtils extends SharedBackendUtils {
...
-      .add("-flatfile", new File(conf.externalBackendFlatFileName().get).getAbsolutePath)
+    val builder = new ArgumentBuilder()
       .add(super.getH2OClientArgs(conf))
       .addIf("-watchdog_client", conf.isAutoClusterStartUsed)
-      .buildArgs()
+
+    if (conf.isAutoClusterStartUsed) {
+      builder.add("-flatfile", new File(conf.externalBackendFlatFileName().get).getAbsolutePath)
+    } else {
+      builder.add("-flatfile", conf.h2oCluster.map(clusterStr => saveFlatFileAsFile(clusterStr).getAbsolutePath))
```
jakubhava: we don't have the full flatfile in case of manual backend start

PR #1681, opened 2019-12-09
[SW-1775] Add security tests for the client-less approach

jakubhava:


py/tests/unit/with_runtime_clientless_sparkling/clientless_test_utils.py
```diff
@@ -19,9 +19,18 @@
 
 
 def createH2OConf(spark):
+
+    with open('login.conf', 'w') as f:
+        f.write('user:pass')
```
mn-mikke: Wouldn't this introduce collisions if H2OConf is created multiple times?

jakubhava: I think it should be fine, the file should get override.

Let's also wait for tests :)

PR #1683, opened 2019-12-09
[SW-1777] Proper stopping of PySparkling

jakubhava:


jakubhava:
@mn-mikke thanks for the suggestion for introducing `update` method. Should be better know, please have a look

jakubhava:
Thank you! I'll wait for test to finish to verify the test is also correct

PR #1685, opened 2019-12-10
[SW-1724] Create UDF for Ordinal Predictions

mn-mikke:


jakubhava:
What is the motivation for this change? I would like to avoid adding last minute features to the current release, instead would focus on stabilization

PR #1689, opened 2019-12-10
[SW-1782] Improve exception when the user is not authentification in rest api client

jakubhava:


core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -279,7 +279,12 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
         case unknown => throw new IllegalArgumentException(s"Unsupported HTTP request type $unknown")
       }
       getCredentials(conf).foreach(creds => request.setHeader(HttpHeaders.AUTHORIZATION, creds))
-      httpClient.execute(request).getEntity.getContent
+      val result = httpClient.execute(request)
+      val statusCode = result.getStatusLine.getStatusCode
+      statusCode match {
+        case 401 => throw new Exception("Unauthorized")
+      }
```
mn-mikke: Wouldn't the  pattern matching throw an exception if  `statusCode` is different from `401`.

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -279,7 +279,12 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
         case unknown => throw new IllegalArgumentException(s"Unsupported HTTP request type $unknown")
       }
       getCredentials(conf).foreach(creds => request.setHeader(HttpHeaders.AUTHORIZATION, creds))
-      httpClient.execute(request).getEntity.getContent
+      val result = httpClient.execute(request)
+      val statusCode = result.getStatusLine.getStatusCode
+      statusCode match {
+        case 401 => throw new Exception("Unauthorized")
```
mn-mikke: Unimportant: WDYT about introducing a new exception type for REST-based errors?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: Very good catch -> it will throw MatchError, thanks!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
jakubhava: yup, like it, will do as part of this change, thanks!

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -279,16 +279,28 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
         case unknown => throw new IllegalArgumentException(s"Unsupported HTTP request type $unknown")
       }
       getCredentials(conf).foreach(creds => request.setHeader(HttpHeaders.AUTHORIZATION, creds))
-      httpClient.execute(request).getEntity.getContent
+      val result = httpClient.execute(request)
+      val statusCode = result.getStatusLine.getStatusCode
+      statusCode match {
+        case 401 => throw new RestApiException("Unauthorized")
```
mn-mikke: WDYT about adding `statusCode` to the error message?

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextRestAPIUtils.scala
```diff
@@ -279,16 +279,28 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
...
+        case _ =>
+      }
+      result.getEntity.getContent
     } catch {
+      case _: RestApiException => throw new H2OClusterNodeNotReachableException(
```
mn-mikke: WDYT about passing the original `RestApiException` to `H2OClusterNodeNotReachableException` as the second argument?

jakubhava:
@mn-mikke I have tried to restructure the exception structure a little bit, please have a look

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
```diff
@@ -180,55 +181,55 @@ trait H2OContextRestAPIUtils extends H2OContextUtils {
...
     *                      consists of the class containing the field and the field name.
     * @tparam ResultType A type that the result will be deserialized to
     * @return A deserialized object
     */
-  private def query[ResultType : ClassTag](
-       endpoint: URI,
-       suffix: String,
-       conf: H2OConf,
-       skippedFields: Seq[(Class[_], String)] = Seq.empty): ResultType = {
+  private def query[ResultType: ClassTag](
```
jakubhava: I have nothing against the original formatting, I just like to use the feature of InteliJ to auto-format. If we have any automatic tool to format it for us, I'm fine with it

core/src/main/scala/org/apache/spark/h2o/utils/H2OContextUtils.scala
```diff
@@ -54,58 +48,6 @@ private[spark] trait H2OContextUtils extends Logging {
     }
   }
 
-  private def isTcpPortAvailable(port: Int): Boolean = {
```
jakubhava: Move out to separate class, as it is know used only in external backend

PR #1690, opened 2019-12-11
[SW-1783] Unify passing of authentication information

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -262,9 +264,12 @@ object SharedBackendConf {
   /** Login configuration file. */
   val PROP_LOGIN_CONF = ("spark.ext.h2o.login.conf", None)
 
-  /** Override user name for cluster. */
+  /** User name for cluster and the client authentication. */
   val PROP_USER_NAME = ("spark.ext.h2o.user.name", None)
 
+  /** Password for the client authentication. */
+  val PROP_PASSWORD = ("spark.ext.h2o.user.name", None)
```
mn-mikke: Shouldn't the property be different from `"spark.ext.h2o.user.name"`?

doc/src/site/sphinx/tutorials/kerberos_auth.rst
```diff
@@ -67,24 +68,25 @@ You can pass the required properties directly as Spark properties, such as:
...
 
 
 Or, you can also use setters available on ``H2OConf`` as:
 
 .. code:: python
 
     from pysparkling import *
-    conf = H2OConf(spark).set_login_conf("kerberos.conf").set_user_name("username").set_kerberos_login_enabled()
-    hc = H2OContext.getOrCreate(spark, conf, auth=("username", "password"))
+    conf = H2OConf(spark).set_login_conf("kerberos.conf").set_user_name("username").set_kerberos_login_enabled().setUserName("username").setPassword("password")
```
mn-mikke: Username seems to be specified twice `.set_user_name("username")` and `.setUserName("username")`

doc/src/site/sphinx/tutorials/ldap.rst
```diff
@@ -72,21 +73,22 @@ You can pass the required properties directly as Spark properties, such as:
...
 
 
 Or, you can also use setters available on ``H2OConf`` as:
 
 .. code:: python
 
     from pysparkling import *
-    conf = H2OConf(spark).set_login_conf("ldap.conf").set_user_name("username").set_ldap_login_enabled()
-    hc = H2OContext.getOrCreate(spark, conf, auth=("username", "password"))
+    conf = H2OConf(spark).set_login_conf("ldap.conf").set_user_name("username").set_ldap_login_enabled().setUserName("username").setPassword("password")
```
mn-mikke: Username seems to be specified twice `.set_user_name("username")` and `.setUserName("username")`

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
jakubhava: As we have requirement that the field `username` needs to be the same as the username passed to the backend, I think it would make sense to reuse this option to limit number of options

doc/src/site/sphinx/tutorials/ldap.rst
jakubhava: 👍

PR #1692, opened 2019-12-11
[SW-1788] Lock cloud in case of rest api client, auto mode

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalH2OBackend.scala
```diff
@@ -223,6 +223,7 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Loggi
     logInfo("Connecting to external H2O cluster.")
     val clusterBuildTimeout = conf.cloudTimeout
     val nodes = if (isRestApiBasedClient(hc)) {
+      lockCloud(conf)
```
mn-mikke: Shouldn't `lockCloud(conf)` be in the below `try` statement in order to handle REST connectivity issues properly?

PR #1711, opened 2019-12-16
[SW-1803] Sparkling Water does not stop automaticaly in client-less mode when hc.stop is not defined

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ProxyStarter.scala
```diff
@@ -22,17 +22,26 @@ import java.net._
 import org.apache.spark.SparkEnv
 import org.apache.spark.expose.Logging
 import org.apache.spark.h2o.H2OConf
-import org.eclipse.jetty.proxy.ProxyServlet
-import org.eclipse.jetty.server.Server
-import org.eclipse.jetty.servlet.{ServletContextHandler, ServletHandler}
+import org.spark_project.jetty.client.HttpClient
```
jakubhava: Use spark jetty -> to avoid collisions

PR #1735, opened 2020-01-15
[SW-1840] Read Chunks via REST API

mn-mikke:
Related PR: https://github.com/h2oai/h2o-3/pull/4217


core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -49,6 +49,14 @@ trait SharedBackendConf {
   def loginConf = sparkConf.getOption(PROP_LOGIN_CONF._1)
   def userName: Option[String] = sparkConf.getOption(PROP_USER_NAME._1)
   def password: Option[String] = sparkConf.getOption(PROP_PASSWORD._1)
+  private[h2o] def credentials: Option[String] = {
```
jakubhava: I think we have similar code to this in Rest API Utils, could it be reused?

mn-mikke: Will take a look, thanks!

PR #1737, opened 2020-01-21
[SW-1843] The Second Call of H2OContext.getOrCreate Throws an Exception

mn-mikke:


bilcus:
Looks good to me, will test the build once you have it ready.

core/src/main/scala/org/apache/spark/network/Security.scala
```diff
@@ -16,41 +16,32 @@
...
   def enableSSL(spark: SparkSession, conf: SparkConf): Unit = {
-    val sslPair = SecurityUtils.generateSSLPair()
-    val config = SecurityUtils.generateSSLConfig(sslPair)
+    val sslPair = SparklingWaterSecurityUtils.generateSSLPair()
+    val config = SparklingWaterSecurityUtils.generateSSLConfig(sslPair)
     conf.set(SharedBackendConf.PROP_SSL_CONF._1, config)
-    val files = Seq(sslPair.jks.getLocation, sslPair.jts.getLocation)
-    files.foreach(SparkSession.builder().getOrCreate().sparkContext.addFile(_))
+    spark.sparkContext.addFile(sslPair.jks.getLocation)
+    if (sslPair.jks.getLocation != sslPair.jts.getLocation) {
```
jakubhava: Good catch!

PR #1743, opened 2020-01-27
[SW-1790] RSparkling in cran should be dummy code to point to our rsparkling in custom repo

jakubhava:
This is the change as we discussed via e-mail @ledell. 

It should ensure that whenever people use the rsparkling from cran, they get the message specified in the file https://github.com/h2oai/sparkling-water/blob/78fe95259f1ae95dcb8ec01f7822fb9513877df8/r-cran/src/R/package.R. It will always point to the latest released sparkling water.

After this is approved I will build it and upload to cran instead of current one year old rsparkling


r-cran/src/DESCRIPTION
```diff
@@ -0,0 +1,36 @@
...
+    Navdeep Gill [aut],
+    Erin LeDell [aut],
+    Michal Malohlava [aut],
+    JJ Allaire [aut],
+    H2O.ai [cph],
+    RStudio [cph]
+Maintainer: Jakub Hava <jakub@h2o.ai>
+Imports:
+    utils,
+    sparklyr (>= 0.3),
```
mn-mikke: Should we be more restrictive about versions of dependant packages? sparklyr 0.3 is 4+ years old.

r-cran/src/DESCRIPTION
```diff
@@ -0,0 +1,36 @@
...
+    h2o (>= 3.8.3.3)
+Suggests: dplyr, testthat
+Description: An extension package for 'sparklyr' that provides an R interface to
+    H2O Sparkling Water machine learning library (see <https://github.com/h2oai/sparkling-water> for more information).
+License: Apache License 2.0 | file LICENSE
+Encoding: UTF-8
+LazyData: true
+NeedsCompilation: no
+SystemRequirements: Java (>6)
```
mn-mikke: Java 8+?

r-cran/src/DESCRIPTION
jakubhava: I just copied this from the main RSparkling package -> but agree, we should specify newer version here and in the main rsparkling package

r-cran/src/DESCRIPTION
jakubhava: similar to above ^ -> just copied, but agree

jakubhava: I was thinking that it might be actually better to do these in another JIRA, WDYT?

r-cran/src/DESCRIPTION
jakubhava: I was thinking that it might be actually better to do these in another JIRA, WDYT?

PR #1746, opened 2020-01-27
[SW-1828] Ensure H2OContext in RSparkling is a class so we don't have to pass sc to methods asH2OFrame and asDataFrame]

jakubhava:
Another step towards unified API between R/Py/Scala

This can also serve as prototype how we can handle other cases, like H2OMOJOModel.createFromMojo etc


r/src/R/H2OContext.R
```diff
@@ -0,0 +1,70 @@
...
+H2OContext.getOrCreate <- function(sc, conf = NULL) {
+  if (is.null(conf)) {
+    conf <- H2OConf(sc)
+  }
+
+  jhc <- invoke_static(sc, "org.apache.spark.h2o.H2OContext", "getOrCreate", spark_context(sc), conf$jconf)
+  hc <- H2OContext(jhc)
+  returnedConf <- invoke(jhc, "getConf")
+  # Because of checks in Sparkling Water, we are sure context path starts with one slash
+  context_path_with_slash <- invoke(returnedConf, "get", "spark.ext.h2o.context.path", "")
```
mn-mikke: nit: camel case?

r/src/R/h2o_context.R
```diff
@@ -67,7 +69,8 @@ h2o_context.spark_jobj <- function(sc, conf = NULL, username = NA_character_, pa
 #' @param sc Object of type \code{spark_connection}.
 #' @export
 h2o_flow <- function(sc) {
-  flowURL <- invoke(h2o_context(sc), "flowURL")
+  print("Method h2o_flow is deprecated and will be deleted in major release 3.30. First create instance of H2OContext as hc <- H2OContext.getOrCreate(sc) and then call hc$openFlow().")
+  flowURL <- invoke(h2o_context(sc), "flowUR     gsresL")
```
mn-mikke: typo?:`"flowUR     gsresL"`

r/src/R/H2OContext.R
```diff
@@ -0,0 +1,70 @@
+getClientConnectedField <- function(jhc) {
```
jakubhava: These are not in the class on purpose as we do not want them to be visible by the user

r/src/R/H2OContext.R
jakubhava: Good catch, fixed :)

r/src/R/h2o_context.R
jakubhava: Ah, yes, thanks!!

PR #1752, opened 2020-01-28
[SW-1854] Expose Individual Chunks via REST API

mn-mikke:


extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/SerializationUtils.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  def sendTimestamp(ab: AutoBuffer, time: Long): Unit = {
+    sendLong(ab, time)
+  }
+
+  def sendTimestamp(ab: AutoBuffer, data: Timestamp): Unit = {
+    sendLong(ab, data.getTime)
+  }
+
+  private[serde] def sendNA(ab: AutoBuffer, expectedType: Byte): Unit = expectedType match {
```
jakubhava: Maybe typo? private[serde]

jakubhava:
I'm excited that we will have full control over external backend in SW repo 💯

jakubhava:
Note: We also need to automatically add the jar to `-libjars` in case of automatic mode of external backend and mention this in the documentation

extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/SerializationUtils.scala
mn-mikke: Would you like me to rename `serde` package?

jakubhava: Maybe something a little bit more explanatory would help? ( Unless this is a know term which just I don't know, in that case it's ok :D )

PR #1762, opened 2020-02-02
[SW-1863] Mention in documentation that High Currency clusters are not yet supported

jakubhava:


jakubhava:
Perfect idea, thanks!

PR #1764, opened 2020-02-02
[SW-1867] Add Sparkling Water UI tab only in case the UI is enabled

jakubhava:


jakubhava:
Bug fix -> we get NPE when we try to attach Sparkling Water UI tab in case UI is disabled

PR #1775, opened 2020-02-04
[SW-1870][FollowUp] Don't automatically apply internal secure connections in manual mode of external backend

jakubhava:


doc/src/site/sphinx/configuration/configuration_properties.rst
```diff
@@ -155,7 +155,7 @@ Configuration properties independent of selected backend
 |                                                    |                | special characters when passing        |
 |                                                    |                | the parameter from a command line.     |
 +----------------------------------------------------+----------------+----------------------------------------+
-| ``spark.ext.h2o.internal_secure_connections``      | ``true``       | Enables secure communications among    |
+| ``spark.ext.h2o.internal_secure_connections``      | ``false``      | Enables secure communications among    |
```
mn-mikke: Shouldn't the default value reflect the conditions? manual start -> `false`, otherwise `true`

jakubhava: Yup, will fix, thanks

jakubhava:
Agree, this is not exactly straightforward.

PR #1776, opened 2020-02-04
[SW-1878] Replace H2O_EXTENDED_JAR with H2O_DRIVER_JAR in all cases

mn-mikke:


doc/src/site/sphinx/deployment/backends.rst
```diff
@@ -101,40 +101,13 @@ lead (as in the previous mode) to h2o cluster being killed as well.
...
- .. code:: bash
-
-    ./bin/get-extended-h2o.sh
-
-The script downloads the jar to the current directory and prints the absolute path to the downloaded jar.
+.. code:: bash
 
-**Note**: If you want to get an extended H2O jar for Sparkling Water and H2O versions that have not yet been
-released, you need to extend the JAR manually. This is explained in the following tutorial: :ref:`extend_jar_manually`.
+    H2O_DRIVER_JAR=/path/to/h2o-driver.jar
```
jakubhava: Should this section still mention how to obtain the JAR -> `./bin/get-h2o-driver.sh`?

doc/src/site/sphinx/deployment/backends.rst
```diff
@@ -144,11 +117,11 @@ environment at the moment. We recommend this approach, as it is easier to deploy
 and it is also more suitable for production environments. When the H2O cluster is started on YARN, it is started
 as a map reduce job, and it always uses the flatfile approach for nodes to cloud up.
 
-Get extended H2O driver, for example, for cdh 5.8:
+Get H2O driver, for example, for cdh 5.8:
 
 .. code:: bash
 
-    H2O_EXTENDED_JAR=$(./bin/get-extended-h2o.sh cdh5.8)
+    H2O_DRIVER_JAR=$(./bin/get-h2o-driver.sh cdh5.8 | tail -n 1 | awk 'NF>1{print $NF})
```
jakubhava: Is this part necessary? `| tail -n 1 | awk 'NF>1{print $NF})` I would prefer fixing the `./bin/get-h2o-driver.sh` and keep the user-facing command simple, WDYT?

doc/src/site/sphinx/deployment/backends.rst
mn-mikke: Good idea, will change it!

doc/src/site/sphinx/deployment/backends.rst
mn-mikke: Np, will do.

jakubhava:
@mn-mikke one last place we need to change a little bit of doc -> in dist/src/index.html (our download page), in the section `h2>Kluster</h2>`

mn-mikke:
@jakubhava will do in a followup PR if you don't mind...

jakubhava:
yup, no problem, my bad I realized it just now :)

What do you think to change that doc on the download page to use automatic cluster start instead? And we can point people to documentation for info on manual start.

PR #1779, opened 2020-02-05
[SW-1881] Fix typo in RestCommunication

jakubhava:
I think we should log these in `logDebug` as in normal app there will be huge amount of requests, WDYT?


core/src/main/scala/org/apache/spark/h2o/backends/external/RestCommunication.scala
```diff
@@ -134,9 +134,9 @@ trait RestCommunication extends Logging {
         connection.getResponseCode()
       }
       statusCode match {
-        case HttpURLConnection.HTTP_OK => logInfo(
+        case HttpURLConnection.HTTP_OK => logDebug(
```
mn-mikke: Maybe I would leave there `logInfo` for a while and change it `logDebug` when logs become polluted with these messages.

PR #1780, opened 2020-02-05
[SW-1875] Ensure method prepareDatasetForFitting works in Rest API based mode

jakubhava:
TODO: Start testing client-less approach in Scala(H2OContext can already be created in REST-API env) and test changes introduced in this change ( so far just the methods as the whole thing is still not ready for rest-api)


core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
```diff
@@ -19,22 +19,53 @@ package org.apache.spark.h2o.backends.external
...
+    val fr = getFrame(conf, frameId)
+    val columns = fr.columns.filter(_.dataType == H2OColumnType.string).map(_.name)
+    convertColumnsToCategorical(conf, frameId, columns)
+  }
+
+  def convertColumnsToCategorical(conf: H2OConf, frameId: String, columns: Array[String]): Unit = {
+    val endpoint = getClusterEndpoint(conf)
+    columns.foldLeft(frameId){ case (frameId, name) =>
+      val params = Map(
+        "ast" -> MessageFormat.format("(assign {0} (:= {0} (as.factor (cols_py {0} {1})) {1} []))", frameId, name)
```
jakubhava: Note to me: `cols_py` looks suspicios

core/src/main/scala/ai/h2o/sparkling/frame/H2OFrame.scala
```diff
@@ -44,4 +44,21 @@ case class H2OColumn(
   def nullable: Boolean = numberOfMissingElements > 0
 }
 
+object H2OColumnType extends Enumeration {
+  val enum, string, int, real, time, uuid = Value
+
+  def fromString(dataType: String): Value = {
+    dataType match {
```
mn-mikke: What about? :
```values.find(_.toString == dataType).getOrElse(throw new RuntimeException(s"Unknown H2O's Data type $dataType"))```

File: core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala [hunk 3]
```diff
@@ -19,22 +19,53 @@ package org.apache.spark.h2o.backends.external
...
+import org.apache.spark.h2o.{H2OConf, H2OContext}
 import org.apache.spark.h2o.utils.NodeDesc
 import water.api.schemas3.FrameChunksV3.FrameChunkV3
 import water.api.schemas3.FrameV3.ColV3
 import water.api.schemas3._
 
 trait RestApiUtils extends RestCommunication {
 
+  def isRestAPIBased(hc: Option[H2OContext] = None): Boolean = {
+    hc.getOrElse(H2OContext.ensure()).getConf.get("spark.ext.h2o.rest.api.based.client", "false") == "true"
```
mn-mikke: Maybe we could make `H2OContext.isRestAPIBased` public or package private and reuse it here.

jakubhava: Good idea, I have removed the method on H2OContext and used the one from RestAPIUtils

core/src/main/scala/ai/h2o/sparkling/frame/H2OFrame.scala
jakubhava: That's how you do it 🚀  Thanks!

jakubhava:
@mn-mikke ready for the final review :)

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
```diff
@@ -38,6 +38,39 @@ trait RestApiUtils extends RestCommunication {
     hc.getOrElse(H2OContext.ensure()).getConf.get("spark.ext.h2o.rest.api.based.client", "false") == "true"
   }
 
+  def convertAllStringVecToCategorical(conf: H2OConf, frameId: String): H2OFrame = {
```
mn-mikke: Maybe these methods could be extracted to a decicate class/trait

core/src/main/scala/org/apache/spark/h2o/backends/external/RestCommunication.scala
```diff
@@ -218,6 +194,31 @@ trait RestCommunication extends Logging {
   }
 }
 
+object RestCommunication {
```
mn-mikke: What is the reason of moving the `decodeSimpleParam`, `decodeParams` and `decodeArray` to an object?

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
jakubhava: Yup, agree, I think it is time that we split the RestAPI Utils to several helper classes or traits

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
jakubhava: Removed

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
jakubhava: I was thinking about it and I actually think that we should put these classes in the new package as well, something like`ai.h2o.sparkling.rest` and then create classes like `Frame`, `Model` etc
I have created JIRA for this task https://0xdata.atlassian.net/browse/SW-1937 ( I would like to do the split and moving after we merge your change to avoid unnecessary rebase conflicts)

I will still move these new methods to the new package. Thanks!

jakubhava: Done 👍

core/src/main/scala/org/apache/spark/h2o/backends/external/RestCommunication.scala
jakubhava: Originally I wanted to reuse these methods outside this class to encode array I need to pass to params, but at the end I have found a cleaner approach. Thanks!

core/src/main/scala/org/apache/spark/h2o/backends/external/RestApiUtils.scala
jakubhava: I have made the frame relates classes available on `ai.h2o.sparkling.H2OFrame` so we can slowly start creating our Scala H2OFrame API via Rest

mn-mikke: That sounds great!

PR #1783, opened 2020-02-05
[SW-1883] Write Individual Chunks via REST API

mn-mikke:


extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
```diff
@@ -17,113 +17,131 @@
...
-  * - expected_type - byte array encoded in Base64 encoding. The types corresponds to the `selected_columns` parameter
-  * - selected_columns - selected columns indices encoded int Base64 encoding.
-  * The result is represented as a stream of binary data. Data are encoded to AutoBuffer row by row.
-  * The data stream starts with the integer representing the number of rows.
+  * This servlet class handles GET and PUT requests for the path /3/Chunk
   */
 final class ChunkServlet extends HttpServlet {
 
-  private case class RequestParameters(
+  private case class GetRequestParameters(
```
jakubhava: Does this class represent parameters for the `GET` request? Shouldn't the case class be named `GETRequestParameters` instead?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
```diff
@@ -133,10 +151,69 @@ final class ChunkServlet extends HttpServlet {
...
         }
       }
-    } catch {
-      case e: Exception => ServletUtils.sendErrorResponse(response, e, uri)
-    } finally {
-      ServletUtils.logRequest(request.getMethod, request, response)
+    }
+  }
+
+  private case class PutRequestParameters(
```
jakubhava: Same as above -> `PUTRequestParameters` ?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
```diff
@@ -17,113 +17,131 @@
...
-      selectedColumnIndices: Array[Int])
+      selectedColumnIndices: Array[Int]) {
 
-  private def getParameterAsString(request: HttpServletRequest, parameterName: String): String = {
-    val result = request.getParameter(parameterName)
-    if (result == null) {
-      throw new RuntimeException(s"Cannot find value for the parameter '$parameterName'")
+    def validate(): Unit = {
+      val frame = DKV.getGet[Frame](this.frameName)
+      if (frame == null) throw new RuntimeException(s"A frame with name '$frameName")
```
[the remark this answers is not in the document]
jakubhava: Also just an idea: Would it be possible to return the exception message to the client in case something went wrong? We could store the exception into the `response` and just create a new schema type like H2OExceptionV3.

It would be nice if all errors on the backend via rest api are propagated to the client as well

core/src/main/scala/org/apache/spark/h2o/backends/external/RestCommunication.scala
```diff
@@ -152,56 +183,65 @@ trait RestCommunication extends Logging {
...
+  private def resolveUrl(endpoint: URI, suffix: String): URL = {
     val suffixWithDelimiter = if (suffix.startsWith("/")) suffix else s"/$suffix"
+    endpoint.resolve(suffixWithDelimiter).toURL
+  }
 
-    val suffixWithParams = if (params.nonEmpty && requestType == "GET") {
-      s"$suffixWithDelimiter?${decodeParams(params)}"
-    } else {
-      suffixWithDelimiter
+  private def setHeaders(connection: HttpURLConnection, conf: H2OConf, requestType: String, params: Map[String, Any]): Unit = {
```
jakubhava: 👍

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
```diff
@@ -17,81 +17,131 @@
...
   }
 
+  override def numOfRows(): Int = currentRowIdx
+
   override def startRow(rowIdx: Int): Unit = {}
 
   override def finishRow(): Unit = {}
-}
 
+  private def increaseCounters(): Unit = {
```
jakubhava: We can remove `increaseCounters` from the code. I think we do not need it anymore as we upload the data via streams. This was originally used on the backend to see when H2O should stop listenting for the incomming data.

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
```diff
@@ -17,81 +17,131 @@
...
-import water.fvec.Frame
+import ai.h2o.sparkling.extensions.serde.ChunkAutoBufferWriter
 
-class ExternalWriteConverterCtx(nodeDesc: NodeDesc, writeTimeout: Int, driverTimeStamp: Short, blockSize: Long) extends WriteConverterCtx {
+class ExternalWriteConverterCtx(conf: H2OConf, nodeDesc: NodeDesc)
+  extends WriteConverterCtx {
 
-  private val externalFrameWriter = ExternalFrameWriterClient.create(nodeDesc.hostname, nodeDesc.port, driverTimeStamp, writeTimeout, blockSize)
+  private var expectedTypes: Array[Byte] = null
+  private var currentColIdx: Int = 0
```
jakubhava: I belive 
```

  private var currentColIdx: Int = 0
  private var currentRowIdx: Int = 0
``` can be removed, see the comment in the `increaseCounter` method

File: core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala (hunk 5 above)
Comment (jakubhava) [reply core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala]: We can edit this method https://github.com/h2oai/sparkling-water/blob/ad063d605e976d6770730c41024170dcae73a0e7/core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtx.scala#L33 to accept dataSize as `Int` and pass there size of the partition instead in both, internal & external cluster case

File: extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/ChunkAutoBufferWriter.scala [hunk 7]
```diff
@@ -17,50 +17,48 @@
...
+final class ChunkAutoBufferWriter(val outputStream: OutputStream) extends Closeable with ChunkSerdeConstants {
 
   private val buffer = new AutoBuffer(outputStream, false)
 
   def writeChunk(frameName: String, chunkId: Int, expectedTypes: Array[Byte], selectedColumnIndices: Array[Int]): Unit = {
     val frame = DKV.getGet[Frame](frameName)
     val chunks = ChunkUtils.getChunks(frame, chunkId)
     val numberOfRows = chunks(0)._len
-    sendInt(buffer, numberOfRows)
+    writeInt(numberOfRows)
```
jakubhava: I think that in the future we can remove this from the protocol as we are streaming the data now, but I wouldn't do it as part of this PR as that would be bigger change

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/SparklingWaterRestApiHandlersRegister.scala
```diff
@@ -0,0 +1,41 @@
...
+      classOf[ImportFrameHandler],
+      "initialize",
+      "Initializes a new frame before pushing chunks to the server")
+
+    context.registerEndpoint(
+      "sw_frame_finalize",
+      "POST " + Paths.FINALIZE_FRAME,
+      classOf[ImportFrameHandler],
+      "finalize",
+      "Performs finalizing procedures after the date chunks were delivered to the server")
```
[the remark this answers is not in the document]
mn-mikke: Good catch, thanks!

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
```diff
@@ -99,14 +98,12 @@ object WriteConverterCtxUtils {
       */
     def convert[T: ClassTag : TypeTag](hc: H2OContext, rddInput: RDD[T], keyName: String, colNames: Array[String], expectedTypes: Array[Byte],
                                        maxVecSizes: Array[Int], sparse: Array[Boolean], func: ConversionFunction[T]): String = {
-      val writeTimeout = hc.getConf.externalWriteConfirmationTimeout
-
       val writerClient = if (hc.getConf.runsInInternalClusterMode) {
         new InternalWriteConverterCtx()
       } else {
         val leader = H2O.CLOUD.leader()
```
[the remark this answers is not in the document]
mn-mikke: Is this relevant if the method is defined in [`ClientBasedConverter`](https://github.com/h2oai/sparkling-water/pull/1783/files/8ae85fe91cd91424eb3dd89c03af43f5b64dd651#diff-efd5dbabfc9f9f8000ebc70807fe3532L84)?

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
```diff
@@ -179,7 +175,7 @@ object WriteConverterCtxUtils {
       // prepare required metadata
       val uploadPlan = Some(ExternalWriteConverterCtx.scheduleUpload(nonEmptyPartitions.size))
 
-      val operation: SparkJob[T] = func(keyName, expectedTypes, uploadPlan, writeTimeout, -1, sparse, nonEmptyPartitions)
+      val operation: SparkJob[T] = func(keyName, expectedTypes, uploadPlan, -1, sparse, nonEmptyPartitions)
```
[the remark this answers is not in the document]
mn-mikke: Good idea!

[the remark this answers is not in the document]
mn-mikke: In fact, `driverTimeStamp` is not needed anymore

core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalWriteConverterCtx.scala
mn-mikke: Ok, will do.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
mn-mikke: Ok, no problem

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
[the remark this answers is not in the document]
mn-mikke: IMHO, the errors will be propagated to the client anyway: https://github.com/h2oai/sparkling-water/pull/1783/files/8ae85fe91cd91424eb3dd89c03af43f5b64dd651#diff-77ccca4e517cc75d148a17f89f920188R124

I hope I don't miss anything, need to test that.

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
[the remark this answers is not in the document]
jakubhava: We are still using `H2O.CLOUD.leader()` which is using runtime property of running H2O cluster -> in rest api approach we don't have any node available on the driver.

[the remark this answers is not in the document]
jakubhava: Good catch, sorry about that!

But this actually leads me to idea that in this change we should remove the `ClientBasedConverter` as we are going to do all the conversions via RestAPI.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ChunkServlet.scala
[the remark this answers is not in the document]
jakubhava: Perfect!

core/src/main/scala/org/apache/spark/h2o/converters/WriteConverterCtxUtils.scala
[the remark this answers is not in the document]
mn-mikke: Ok, will do.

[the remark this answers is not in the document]
jakubhava: Perfect, thanks a lot!

jakubhava:
Just realized one more place which will need update: https://github.com/h2oai/sparkling-water/blob/488485b47e6f1dde1cfe32b148095f0c273360e3/core/src/main/scala/org/apache/spark/h2o/backends/external/ExternalBackendUtils.scala#L28 This method still references the `water.ExternalFrameUtils` defined on the H2O side. Maybe we can even move this method to `serde` as it is somewhat related to that?

core/src/main/scala/ai/h2o/sparkling/frame/H2OFrame.scala
```diff
@@ -120,4 +123,24 @@ object H2OFrame extends RestCommunication {
       location = clusterNodes(sourceChunk.node_idx))
   }
 
+  def initializeFrame(conf: H2OConf, frameId: String, columns: Array[String]): InitializeFrameV3 = {
```
jakubhava: NIT (I wouldn't do it know): It would be nice for these methods to be private/ package-private from some point so `H2OFrame` can be user-facing class, but we can do any cleanups after this change is done

mn-mikke:
>This method still references the water.ExternalFrameUtils defined on the H2O side. Maybe we can even move this method to serde as it is somewhat related to that?

done

core/src/main/scala/ai/h2o/sparkling/frame/H2OFrame.scala
mn-mikke: Good point!

mn-mikke: Will do it in a different PR, It would require to move `ExternalWriteConverterCtx` to the same package as `H2OFrame`.

core/src/main/scala/org/apache/spark/h2o/converters/SparkDataFrameConverter.scala
```diff
@@ -138,7 +138,7 @@ private[h2o] object SparkDataFrameConverter extends Logging {
     con.closeChunks(localRowIdx)
 
     // Return H2O chunk index and number of rows in this chunk
-    (chunkIdx, con.numOfRows())
+    (chunkIdx, dataSize)
```
mn-mikke: @jakubhava Need to revert this line to fix tests on internal backend, since `dataSize` is `-1` for internal tests. if you don't mind, we can proceed with refactoring after the PR is merged.

jakubhava: Yup, no worries, we'll clean this up after this PR is done, there will be more places we can simplify actually

jakubhava:
And we are green!! Huge thanks for this change @mn-mikke !

PR #1791, opened 2020-02-06
[SW-1899] Fix intermittent bug in rest api client tests

jakubhava:


core/src/main/scala/org/apache/spark/h2o/backends/external/ProxyStarter.scala
```diff
@@ -30,9 +30,9 @@ import org.spark_project.jetty.util.thread.{QueuedThreadPool, ScheduledExecutorS
 
 object ProxyStarter extends Logging {
   def startFlowProxy(conf: H2OConf): URI = {
+    var port = findNextFreeFlowPort(conf.clientWebPort, conf.clientBasePort)
     while (true) {
       try {
-        val port = findNextFreeFlowPort(conf)
```
jakubhava: we should always start with next port if next iteration if we failed during `server.start()`

core/src/main/scala/org/apache/spark/h2o/backends/external/ProxyStarter.scala
```diff
@@ -46,7 +46,8 @@ object ProxyStarter extends Logging {
         server.start()
         return new URI(s"${conf.getScheme()}://${SparkEnv.get.blockManager.blockManagerId.host}:$port${conf.contextPath.getOrElse("")}")
       } catch {
-        case _: BindException =>
+        case _: BindException => port = findNextFreeFlowPort(conf.clientWebPort, port)
+
```
mn-mikke: nit: extra line

core/src/main/scala/org/apache/spark/h2o/backends/external/ProxyStarter.scala
```diff
@@ -30,9 +30,9 @@ import org.spark_project.jetty.util.thread.{QueuedThreadPool, ScheduledExecutorS
 
 object ProxyStarter extends Logging {
   def startFlowProxy(conf: H2OConf): URI = {
+    var port = findNextFreeFlowPort(conf.clientWebPort, conf.clientBasePort)
```
mn-mikke: if the first try to get a free port fails, it will cause failure of the whole process. Is it what we want?

jakubhava: Good catch, nope, thanks!

core/src/main/scala/org/apache/spark/h2o/backends/external/ProxyStarter.scala
jakubhava: thanks!

PR #1796, opened 2020-02-06
[SW-1904] Remove extended jar from the codebase

jakubhava:
This should also speed up our regular test & release pipeline


jenkins/sparklingWaterPipeline.groovy
```diff
@@ -115,17 +101,21 @@ def withSharedSetup(sparkMajorVersion, config,  shouldCheckout, code) {
 
                 config.put("sparkVersion", getSparkVersion(config))
 
+                if (config.buildAgainstH2OBranch.toBoolean()) {
+                    config.put("driverJarPath", "${env.WORKSPACE}/h2o-3/h2o-hadoop-2/h2o-${config.driverHadoopVersion}-assembly/build/libs/h2odriver.jar")
+                } else {
+                    def majorVersionLine = readFile("gradle.properties").split("\n").find() { line -> line.startsWith('h2oMajorVersion') }
```
mn-mikke: Isn't there any way how to parse `gradle.properties` to  a map and then extract a given values from the map?

jenkins/sparklingWaterPipeline.groovy
```diff
@@ -257,20 +242,24 @@ def prepareSparkEnvironment() {
...
-                     fi
-                fi
-                """
+            if (config.backendMode.toString() == "external") {
+                if (config.buildAgainstH2OBranch.toBoolean()) {
+                    sh """
+                        git clone https://github.com/h2oai/h2o-3.git
+                        cd h2o-3
+                        git checkout ${config.h2oBranch}
+                        . /envs/h2o_env_python2.7/bin/activate
```
mn-mikke: Shouldn't the same be also done for `. /envs/h2o_env_python3.6/bin/activate`?

jakubhava: Nope, we just need any python, but we don't later depend on any specific python version as we just copy the python sources directly

jenkins/sparklingWaterPipeline.groovy
jakubhava: Yes, that would be possible

jakubhava: Will do it right after this PR, have JIRA for it https://0xdata.atlassian.net/browse/SW-1909

It will be nice clean-up as well, but don't want to make this change even bigger. Is it ok with you ? :)

mn-mikke: ok, no problem

PR #1798, opened 2020-02-07
[SW-1908] Ignore SVM Tests

mn-mikke:


jakubhava:
Is it related to the latest hangs on external backend? I think that for H2OSVM spark wrapper to work we need to add sparkling-water-ml to our extensions jar.

mn-mikke:
Unfortunately, for SVM we need more sparkling-water-core, and spark-ml. The current SVM implementation is incompatible with the concept of extension assembly.

jakubhava:
I see. I would like to remove this H2O wrapper around Spark SVM anyway and expose H2OSVM in the same way we have for rest of the users. We can mention in the doc that people who would like the previous functionality can use Spark SVM directly.

I don't think many people use this and if we first deprecate & document in migration guide, we should be good

PR #1803, opened 2020-02-09
[SW-1914] Copy extension jar to jars folder in distribution archive

jakubhava:
I have discovered that all documentation mentions that location of extension assembly jar is in the root of distribution archive, however the file is being copied into `sparkling-water-extensions/build/lib` which is a bug.

I have tried to fix the doc + copy the file into newly created `jars` folder to avoid complicated directory structure inside the sw distribution. I would like to put there also the original assembly jar, but created another JIRA for it https://0xdata.atlassian.net/browse/SW-1915

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -147,10 +147,10 @@ task copyFilesForZipDistribution {
         }
 
         copySingle("r/build/", "rsparkling_${version}.tar.gz", "$zipDir")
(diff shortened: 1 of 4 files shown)


h2o-ops:
Hi @jakubhava, your PR title "[1914] Copy extension jar to jars folder in distribution archive" is missing JIRA issue number! Please specify it in the following form [SW-]

mn-mikke:
Oh, I see `sparkling-water-assembly_xxx.jar` is moved to as well. https://github.com/h2oai/sparkling-water/pull/1804

PR #1808, opened 2020-02-10
[SW-1859] Support Azure Data Lake Storage Gen2

Pscheidl:
https://0xdata.atlassian.net/browse/PUBDEV-6917

## Problem and solution
A customer of ours wants Sparkling Water to be able to import files from ABFS(S).  On normal hadoop, this can be done by modifying HADOOP config and adding https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-azure in version > 3.20.0, which contains support for ABFS. Putting credentials into `hdfs-config.xml;` Such a configuration file might look as follows:

```xml
<?xml version="1.0"?>
<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
<configuration>
<property>
  <name>fs.azure.account.auth.type.h2ogen2test.dfs.core.windows.net</name>
  <value>SharedKey</value>
  <description>
  </description>
</property>
<property>
  <name>fs.azure.account.key.h2ogen2test.dfs.core.windows.net</name>
  <value>SECRET</value>
  <description>
  The secret password. Never share these.
  </description>
</property>
<property>
    <name>fs.abfs.impl</name>
    <value>org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem</value>
</property>
<property>
    <name>fs.abfss.impl</name>
    <value>org.apache.hadoop.fs.azurebfs.SecureAzureBlobFileSystem</value>
</property>
</configuration>
```

However, on platforms like **DataBricks**, the configuration of classes to use for ABFS is already pre-set and looks exactly as follows:

```xml
<property>
    <name>fs.abfs.impl</name>
    <value>shaded.databricks.v20180920_b33d810.org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem</value>
</property>
<property>
    <name>fs.abfss.impl</name>
    <value>shaded.databricks.v20180920_b33d810.org.apache.hadoop.fs.azurebfs.SecureAzureBlobFileSystem</value>
</property>
```

The **problem** for the user is to provide such configuration into H2O, so that H2O's `PersistHdfs` class can read from it and set-up a proper configuration.

One **generic** way to do that is to expose `hadoop-config` parameter in H2O, enabling Sparkling Water and possibly other clients to pass whole hadoop config (`sparkContext.hadoopConfiguration` to be pricise when it comes to SW) to H2O. Different H2O components then may use parts of it where necessary. **This enables us to provide all sorts of workarounds for users without the need of publishing new versions.**

### Implicit configuration handover

By default, SW will pass `sc.hadoopConfiguration`, unless the user specifies the configuration to pass manually. User is only 

**Starting Sparkling Water**

**ABFS works in H2O**

Screenshot from 2020-02-10 11-03-31

### Explicit configuration passed by the user

User is able to set the precise `org.hadoop.Configuration` via a setter.

**Staring Sparkling Water with explicitly set Configuration**

**Works in H2O as well**
Screenshot from 2020-02-10 10-53-51

### Explicit configuration with configuration file passed
This is an alternative to the previous setter - but an overloaded setter is used, setting the path to an XML configuration file redable by `org.hadoop.Configuration` directly.

### Notes

H2O already has `hdfs-config` property which is considered to be specialized and is always preferred in `PersistHdfs` to `hadoopconfig`.


h2o-ops:
Hi @Pscheidl, your PR title "PUBDEV-6917 - Support Azure Data Lake Storage Gen2 by exposing whole …" is missing JIRA issue number! Please specify it in the following form [SW-]

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -164,15 +169,29 @@ trait SharedBackendConf {
 
   def setNodeBasePort(port: Int) = set(PROP_NODE_PORT_BASE._1, port.toString)
 
+  def setHadoopConf(path: String) = set(PROP_HADOOP_CONF._1, path)
+
+  def setHadoopConf(conf: Configuration): H2OConf = {
+    val hadoopConfigTempFile: File = File.createTempFile("hadoop_config_all", ".xml")
```
mn-mikke: Shouldn't we delete the file after the driver terminates? 
```hadoopConfigTempFile.deleteOnExit()```

File: core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala [hunk 2]
```diff
@@ -164,15 +169,29 @@ trait SharedBackendConf {
...
+    val fileWriter: FileWriter = new FileWriter(hadoopConfigTempFile)
+    try {
```
mn-mikke: nit: you can use `withResource` statement after you import `ai.h2o.sparkling.utils.ScalaUtils._`

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -380,4 +399,7 @@ object SharedBackendConf {
 
   /** Extra properties passed to H2O client during startup. */
   val PROP_CLIENT_EXTRA_PROPERTIES = ("spark.ext.h2o.client.extra", None)
+
+  /** Path to Java KeyStore file used for the internal SSL communication. */
+  val PROP_HADOOP_CONF = ("spark.ext.h2o.hadoof_conf", None)
```
mn-mikke: nit:  I'm not sure whether we should use camel case here. `spark.ext.h2o.hadoopConf` It's rather a question to think about. The conventions seems to be inconsistent.

mn-mikke: typo: `hadoof` -> `hadoop`

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
```diff
@@ -380,4 +399,7 @@ object SharedBackendConf {
 
   /** Extra properties passed to H2O client during startup. */
   val PROP_CLIENT_EXTRA_PROPERTIES = ("spark.ext.h2o.client.extra", None)
+
+  /** Path to Java KeyStore file used for the internal SSL communication. */
```
mn-mikke: The comment seems to be wrong.

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
Pscheidl: There is `spark.ext.h2o.internal_security_conf` or `spark.ext.h2o.internal_secure_connections"` and many others. I'm happy to follow any convention. No problem here.

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
Pscheidl: Fixed. Thank you.

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
Pscheidl: Thank you, corrected.

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
Pscheidl: Done ! Great ! Thanks !

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
Pscheidl: Yes, we should. Changed. Gave it a try after this change once again on the cluster - works. Thank you.

mn-mikke:
@Pscheidl Could you add a note about new property to `doc/src/site/sphinx/configuration/configuration_properties.rst`? Thanks!

Pscheidl:
Done @mn-mikke . Thanks for reminding me.

mn-mikke:
@Pscheidl I've realized the last thing, `getH2OCommonArgs` is not called from external backend. Can you also pass the property to external backend? If `-hadoop_config` is not exposed in H2O driver, prefix key and value with `-J` arguments.

Pscheidl:
Will do. We have no electricity currently. Will fix asap.

michalkurka:
> @Pscheidl I've realized the last thing, getH2OCommonArgs is not called from external backend. Can you also pass the property to external backend? If -hadoop_config is not exposed in H2O driver, prefix key and value with -J arguments.

Isn't this the wrong thing to do? You can have Spark running on one environment (k8s) and connecting to an external cluster running on Hadoop. You probably don't want to send the configuration over to Hadoop.

IMHO this change can only work well in the internal mode.

mn-mikke:
@michalkurka That's a good point! Haven't thought about such a scenario.

mn-mikke:
@michalkurka if Spark is running in a different environment, doesn't it mean that h2o cluster has to be started manually? If we automatically start H2O cluster with external backend, doesn't imply that Spark and H2O are on the same Hadoop cluster? Just asking whether I'm not missing something. Thanks!

michalkurka:
> @michalkurka if Spark is running in a different environment, doesn't it mean that h2o cluster has to be started manually? If we automatically start H2O cluster with external backend, doesn't imply that Spark and H2O are on the same Hadoop cluster? Just asking whether I'm not missing something. Thanks!

Hmm, I think you have a point here. My thinking was more about external cluster not started automatically.

I still think it is problematic to just automatically transfer the configuration to a Hadoop job just because the classpath can be different and the filesystem implementation classes might be missing (for example). Can we test it somehow?

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendConf.scala
jakubhava: Hm, I would use the variant with `_` for now even though it is not the best, but at least consistent

core/src/main/scala/org/apache/spark/h2o/backends/SharedBackendUtils.scala
```diff
@@ -113,6 +113,10 @@ private[backends] trait SharedBackendUtils extends Logging with Serializable {
       conf.setH2OClientLogLevel(incLogLevel(conf.h2oClientLogLevel, "INFO"))
     }
 
+    if(!conf.hadoopConf.isDefined){
```
jakubhava: I would apply the automatic hadoop configuration just in case of Internal backend for now. It is fine to just move this piece of code to to https://github.com/h2oai/sparkling-water/blob/ad063d605e976d6770730c41024170dcae73a0e7/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala#L59

jakubhava:
After reading this discussion, I think I agree with @michalkurka regarding the fact that the classes in the hadoop configuration might be missing in case of external cluster (manual/automatic).

I would for now expose this option just for the internal backend and remain external backend unchanged.

Pscheidl:
Thank you @jakubhava .

1. Moved the configuration into internal (scala) - in commit https://github.com/h2oai/sparkling-water/pull/1808/commits/a8164fbb6af3b822ab5c2c7a1b47510cd4474efe
1. Added the pySparkling getter & setter. Not sure whether the Python/JVM bridge will use the appropriate setter on Scala backend.

py/src/ai/h2o/sparkling/SharedBackendConf.py
```diff
@@ -832,11 +834,15 @@ def set_client_check_retry_timeout(self, timeout):
...
         warnings.warn("Method 'set_client_extra_properties' is deprecated and will be removed in the next major release. Please use 'setClientExtraProperties'.")
         return self.setClientExtraProperties(extraProperties)
 
     def setClientExtraProperties(self, extraProperties):
         self._jconf.setClientExtraProperties(extraProperties)
         return self
+
+    def setHdfsConfig(self, hdfsConfig):
+        self._jconf.setHdfsConfig(hdfsConfig)
+        return self
```
jakubhava: Nil: Missing new line

Pscheidl: Thank you. Added. https://github.com/h2oai/sparkling-water/pull/1808/commits/ee0a5aa7776e510721f4b9a3b91fe2dac63d0364

PR #1812, opened 2020-02-10
[SW-1589] Remove H2OSVM from the codebase

jakubhava:


jakubhava:
@mn-mikke I will go ahead and merge this one, but added you as review so you can have a look :)

Great thing is that this will also clean-up an unify a lots of code in the sparkling-water-ml package ( everything is finally under ai.h2o.sparkling)

mn-mikke:
LGTM ;-)

PR #1816, opened 2020-02-11
[SW-1897] In prediction which do classification, be more explicit about relations about class and probabilities

jakubhava:
Depends on https://github.com/h2oai/sparkling-water/pull/1817/files

### DIFF
--- a/ml/src/test/scala/ai/h2o/sparkling/ml/algos/MultinomialPredictionTestSuite.scala
+++ b/ml/src/test/scala/ai/h2o/sparkling/ml/algos/MultinomialPredictionTestSuite.scala
@@ -24,19 +24,16 @@ import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.{FunSuite, Matchers}
 import water.api.TestUtils
+
 @RunWith(classOf[JUnitRunner])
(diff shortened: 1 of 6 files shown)


ml/src/test/scala/ai/h2o/sparkling/ml/algos/BinomialPredictionTestSuite.scala
```diff
@@ -87,10 +89,10 @@ class BinomialPredictionTestSuite extends FunSuite with Matchers with SharedH2OT
 
     val datasetFields = dataset.schema.fields
     val labelField = StructField("label", StringType, nullable = false)
-    val propFields = Seq("p0", "p1").map(StructField(_, DoubleType, nullable = false))
+    val probabilitiesField = StructField("probabilities", MapType(StringType, DoubleType, valueContainsNull = false), nullable = false)
```
mn-mikke: `MapType` uses a linear key search. Do we know how many classes we can have in real world?

jakubhava: Good to know! I don't think that in real life use-cases there are that many classes. Plus this is still part of the detailed prediction column whilst people who care about performance will probably still use just the regular `prediction` column

PR #1834, opened 2020-02-14
[SW-1946] Move H2OFrame to Sparkling Water

jakubhava:
There is a quite a lot of changed files, but this change is fairly small.

I have copied two classes from H2O: H2OFrame & FrameOps ( needed to rename our FrameUtils to avoid conflict with FrameUtils on H2O side)

The reason why there are more changed files is that I wanted to do this properly and don't put the H2OFrame class into `extensions` package. I believe the right place for this class is `core`.

However `asDataFrame` and `asRDD` transformations were working on both `H2OFrame` and `Frame` which in case we were converting `H2OFrame` leaded to no class found exception on the backend. 

I have therefore ensured that we always use H2O's class `water.fvec.Frame` in the conversion back to Spark. This means that there is no need to have the H2OFrame class in extensions as the backend will never see it. Which I hope is right approach -> we don't want to leak classes from SW


core/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,138 @@
...
+      // Get only indexes of selected vectors
+      selector(name, vec)
+    }).map(idx => {
+      val name = names(idx)
+      val vec = vecs(idx)
+      (idx, transformation(name, vec))
+    }).foreach { case (idx: Int, newVec: Vec) => {
+      val v = self.replace(idx, newVec)
+      if (removeVec) v.remove(futures)
+    }
```
mn-mikke: nit: indentation seems to be broken

core/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,138 @@
...
+  def apply(transformation: VecTransformation, selector: VecSelector, removeVec: Boolean = true): H2OFrame = {
+    val vecs = self.vecs()
+    val names = self.names()
+    val futures = new Futures()
+    vecs.indices.filter(idx => {
+      val name = names(idx)
+      val vec = vecs(idx)
+      // Get only indexes of selected vectors
+      selector(name, vec)
+    }).map(idx => {
```
mn-mikke: Could we merge content of `map` and `foreach`, in order to avoid materialization of the output from map?

mn-mikke: Maybe replace the whole statement with one "for comprehension" statement

core/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,138 @@
...
+    futures.blockForPending()
+    self
+  }
+
+  /**
+   * Transform columns in enum columns
+   *
+   * @param cols : Array[ String ] containing all the names of enum columns
+   */
+  def colToEnum(cols: Array[String]): Unit = {
```
mn-mikke: Would it be possible to simplify the logic of this method?

jakubhava: Yup, we can replace completely by H2OFrameSupport.convertColumnsToCategorical

core/src/main/scala/water/fvec/FrameOps.scala
jakubhava: Thanks, should be easier to read now :)

jakubhava:
Note: We can actually keep water.fvec.H2OFrame just in core, it needs to be in extensions project, for now.

It is because that calls in Scala like `new H2OFrame(uri)` on the external backend requires the class `H2OFrame` to be available on the worker nodes. After we move everything to Rest API in external backend, we can move this class to `core`

core/src/main/scala/water/fvec/H2OFrameUtils.scala
```diff
@@ -0,0 +1,27 @@
...
+* limitations under the License.
+*/
+
+package water.fvec
+
+
+/**
+ * This is a simple bridge to access package-private/protected members.
+ */
+object H2OFrameUtils {
```
jakubhava: Method `getChunksLocations` was unused and all other are already defined on ChunkUtils except the `createNewChunks` with `sparse` argument. Created PR on H2O side and SW side we can remove this duplication

extensions/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+    // Block for all fired deletes
+    futures.blockForPending()
+    self
+  }
+
+  /**
+   * Transform columns in enum columns
+   *
+   * @param cols : Array[ String ] containing all the names of enum columns
```
mn-mikke: nit:  `Array[ String ]` ->  `Array[String]`

extensions/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,130 @@
...
+      this.update()
+    } else {
+      throw new IllegalArgumentException("One or several columns are not present in your DataFrame")
+    }
+  }
+
+  /**
+   * Transform columns in enum columns
+   *
+   * @param cols : Array[ Int ] containing all the indexes of enum columns
```
mn-mikke: nit: `Array[ Int ]` ->  `Array[Int]`

extensions/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,130 @@
...
+   */
+  def colToEnum(cols: Array[Int]): Unit = {
+    Try(cols.map(i => this.name(i))) match {
+      case Success(s) => colToEnum(s)
+      case Failure(t) => println(t)
```
mn-mikke: `println(t)` -> shouldn't an error be logged properly?

extensions/src/main/scala/water/fvec/FrameOps.scala
```diff
@@ -0,0 +1,130 @@
...
+   * Rename a column of your DataFrame
+   *
+   * @param index   : Index of the column to rename
+   * @param newName : New name
+   */
+  def rename(index: Int, newName: String): Unit = {
+    val tmp = this.names
+    Try(tmp(index) = newName) match {
+      case Success(_) => this._names = tmp
+      case Failure(t) => println(t)
```
mn-mikke: `println(t)` -> shouldn't an error be logged properly?

extensions/src/main/scala/water/fvec/H2OFrame.scala
```diff
@@ -0,0 +1,180 @@
...
+  /** Create a new H2OFrame based on existing Java Frame.
+   *
+   * Simple field copy, so the Frames share
+   * underlying arrays.  Recommended that the input Java Frame be dead after
+   * this call.
+   *
+   * @param fr Java frame
+   * @return new H2O frame with parsed data
+   */
+  def this(fr: Frame) = this(if (fr._key != null) fr._key else Key.make("dframe" + Key.rand()).asInstanceOf[Key[Frame]], fr._names, fr.vecs())
```
mn-mikke: nit: Can we split this expression into more lines?

extensions/src/main/scala/water/fvec/FrameOps.scala
jakubhava: 👍

extensions/src/main/scala/water/fvec/FrameOps.scala
jakubhava: 👍

extensions/src/main/scala/water/fvec/FrameOps.scala
jakubhava: Yup, I just copied the class from H2O-3 and kept it as it was, will fix :)

extensions/src/main/scala/water/fvec/FrameOps.scala
jakubhava: 👍

extensions/src/main/scala/water/fvec/H2OFrame.scala
jakubhava: I think it won't be possible as it is in the constructor and `this(` call needs to be first in the method definiton

jakubhava: But I tried to put each argument on a new line for better readability

jakubhava:
I hope I have addressed all the review comments :) Please have a final look. Thanks a lot!

PR #1841, opened 2020-02-17
[SW-1868] Remove deprecated option externalWriteConfirmationTimeout

jakubhava:


jakubhava:
Just simple removal removals of deprecated methods

PR #1844, opened 2020-02-17
[SW-1919] Make benchmarks up-to-date with removal of extended h2o jar

mn-mikke:


mn-mikke:
The bechmarks.jar is on the classpath of each H2O node, because of [`DKVCleaner`](https://github.com/h2oai/sparkling-water/blob/a129222756951d2358f9ed64135c0a06ec8a1c04/benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/DKVCleaner.scala) In the original version, the h2o extended jar had to be enhanced with benchmark classes.

mn-mikke:
@jakubhava I've also removed generation of a dedicated assembly jar for benchmarks. The benchmarks will utilize the generic assembly.

PR #1845, opened 2020-02-17
[SW-1958] Property for Passing Extra Jars to External Backend

mn-mikke:


jakubhava:
Ah I see, it is used in the benchmarks, is it necessary though?

mn-mikke:
Yes, the motivation is coming from benchmarks https://github.com/h2oai/sparkling-water/pull/1844#issuecomment-587420137 . IMHO, this property might be helpful in future anyway. WDYT?

jakubhava:
Yup, I would keep it exposed, just wanted to ask for the motivation in case of benchmarks :) Thanks!

jakubhava:
Thank you @mn-mikke!

PR #1848, opened 2020-02-18
[SW-1965] Test PRs only on Spark 2.1 and Spark 2.4

jakubhava:


jakubhava:
Thanks, it's defined here https://github.com/h2oai/h2o-jenkins-pipeline-lib/blob/master/vars/isPrJob.groovy. Our jenkins server defines this repo as utilities library and we can can use methods defined in here

PR #1852, opened 2020-02-19
[SW-1963] Fix execution of tests using local-cluster

jakubhava:
The reason was that Spark in local-cluster mode tried to create directory $SPARK_HOME/work but it did not have correct rights, that is why the tests in local-cluster were failing.


jakubhava:
Just note: Failed on unrelated intermittent failure of R tests. I till try to see if I can figure out we these R tests sometimes fail. These are the only intermittent failures we have in our tests

PR #1856, opened 2020-02-20
[SW-1975] Move spark.ext.h2o.hdfs_conf Property among Properties of Internal Backend

mn-mikke:
I propose to move this property since it's used only by the internal backend.


jakubhava:
I think we should still keep it exposes as it is as people in external backend might want to point to specific HDFS configuration manually, WDYT?

mn-mikke:
@jakubhava I thought that the outcome of your discussion with @michalk was that this property will be supported only for the internal backend? Has anything changed?

mn-mikke:
I'm happy with any solution, but current situation seems to me inconsistent.

jakubhava:
Yup, agree, I just confused myself :) Feel free to merge, thank you!

PR #1858, opened 2020-02-20
[SW-1846] Be able to compile Sparkling Water with Scala 2.12

jakubhava:
Enable users to build Sparkling Water with Scala 2.12. 

Manually tested that it compiles and tests unit pass fine with Scala 2.12

Just updated code so it can build on Scala 2.11 and Scala 2.12. 

Note: We do not publish artifacts for non-default Scala for given Spark at this moment. So no tests on purpose at this stage. Just to give developers the ability to start building SW with Scala 2.12 and give us feedback.

Also upgrade to Spark. 3.0 will be easier as we now know that our code is agnostic to Scala version (at least when considering 2.11 and 2.12). We can just specify the scalaBaseVersion on sparkProperties level, not on the global gradle.properties level.

### DIFF
--- a/extensions/src/main/scala/ai/h2o/sparkling/extensions/stacktrace/StackTraceCollector.scala
+++ b/extensions/src/main/scala/ai/h2o/sparkling/extensions/stacktrace/StackTraceCollector.scala
@@ -27,7 +27,6 @@ class StackTraceCollector extends AbstractH2OExtension {
   private var interval = -1 // -1 means disabled
 
   override def getExtensionName = "StackTraceCollector"
-
   override def printHelp(): Unit = {
     System.out.println(
       "\nStack trace collector extension:\n" +
@@ -61,16 +60,15 @@ class StackTraceCollector extends AbstractH2OExtension {
 
   private class StackTraceCollectorThread extends Thread("StackTraceCollectorThread") {
     setDaemon(true)
-
+    import scala.collection.JavaConverters._
     override def run(): Unit = {
       while (true) {
         try {
           Log.info("Taking stacktrace at time: " + new Date)
-          val allStackTraces = Thread.getAllStackTraces
-          import scala.collection.JavaConversions._
-          for (e <- allStackTraces.entrySet) {
-            Log.info("Taking stacktrace for thread: " + e.getKey)
-            for (st <- e.getValue) {
+          val allStackTraces = Thread.getAllStackTraces.asScala
+          for (e <- allStackTraces) {
+            Log.info("Taking stacktrace for thread: " + e._1)
+            for (st <- e._2) {
               Log.info("\t" + st.toString)
             }
           }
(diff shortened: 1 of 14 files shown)


gradle.properties
```diff
@@ -7,7 +7,7 @@ h2oMajorName=yu
 # H2O Build version, defined here to be overriden by -P option
 h2oBuild=4
 # Scala version for Sparkling Water. By default, for 2.11 we use 2.11.12.
-scalaBaseVersion=2.11
+scalaBaseVersion=2.12
```
jakubhava: just for testing, revert to 2.11 before merging

repl/src/main/scala/ai/h2o/sparkling/repl/H2OInterpreter.scala
```diff
@@ -71,13 +71,6 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
 
     settings
   }
-
-
-  override def valueOfTerm(term: String): Option[Any] = {
-    try Some(intp.eval(term)) catch {
```
jakubhava: Ugly hack + preventing us to use it in Scala 2.12

extensions/src/main/scala/ai/h2o/sparkling/extensions/stacktrace/StackTraceCollector.scala
```diff
@@ -61,16 +60,15 @@ class StackTraceCollector extends AbstractH2OExtension {
...
       while (true) {
         try {
           Log.info("Taking stacktrace at time: " + new Date)
-          val allStackTraces = Thread.getAllStackTraces
-          import scala.collection.JavaConversions._
-          for (e <- allStackTraces.entrySet) {
-            Log.info("Taking stacktrace for thread: " + e.getKey)
-            for (st <- e.getValue) {
+          val allStackTraces = Thread.getAllStackTraces.asScala
+          for (e <- allStackTraces) {
```
mn-mikke: What about `for ((key, value) <- allStackTraces) {` ?

repl/src/main/scala/ai/h2o/sparkling/repl/PatchUtils.scala
```diff
@@ -56,7 +56,11 @@ private[repl] object PatchUtils {
     val patcher: Patcher = (obj: AnyRef, clz: Class[_]) => {
       val f = clz.getDeclaredField("REPLClass")
       f.setAccessible(true)
-      f.set(obj, OUTER_SCOPE_REPL_REGEX)
+      try {
+        f.set(obj, OUTER_SCOPE_REPL_REGEX)
+      } catch {
+        case _: Throwable => // we have already patched once
```
mn-mikke: Shouldn't we be more specific with the exception type?

jakubhava: Yup, will improve

extensions/src/main/scala/ai/h2o/sparkling/extensions/stacktrace/StackTraceCollector.scala
jakubhava: Great idea!

PR #1862, opened 2020-02-21
[SW-1935] Expose getLeaderboard() Methond on AutoML

mn-mikke:


jakubhava:
Note: Can you please mention this on our AutoML doc page?

mn-mikke:
Note & note 2: Np, will do both. ;-)

py/src/ai/h2o/sparkling/ml/algos/H2OAutoML.py
```diff
@@ -68,8 +71,16 @@ def __init__(self,
         self._set(**kwargs)
 
     def leaderboard(self):
+        warnings.warn("The 'leaderboard' method is deprecated and will be removed in the SW version 3.30. "
+                      "Use the 'getLeaderboard' method instead!")
         leaderboard_java = self._java_obj.leaderboard()
         if leaderboard_java.isDefined():
-            return DataFrame(leaderboard_java.get(), self._hc._sql_context)
+            return DataFrame(leaderboard_java.get(), SparkSession._instantiatedSession._wrapped)
```
mn-mikke: Fixing the current `leaderboard` implementation

mn-mikke:
I changed my mind and decided to introduce a proper `getLeaderboard()` method.  and deprecated current `leaderboard` I think this will give users a better flexibility to choose correct extra columns. The training process in AutoML might be very time consuming and users might forget to set the parameter before training.

sebhrusen:
@mn-mikke Although I don't have anything with your first approach, I think that a `getLeaderboard` method as you're now suggesting is more convenient as we plan to add more features through this method in other clients.

Also, some future columns in the extended leaderboard may be time-consuming, that's why it's separated from training by default, although the backend is trying to optimize this and may cache some values if they are already available during training.

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -124,17 +127,28 @@ class H2OAutoML(override val uid: String) extends Estimator[H2OMOJOModel]
...
+
+  def getLeaderboard(extraColumns: String*): DataFrame = getLeaderboard(extraColumns.toArray)
+
+  def getLeaderboard(extraColumns: java.util.ArrayList[String]) : DataFrame = {
+    getLeaderboard(extraColumns.asScala.toArray)
+  }
+
+  def getLeaderboard(extraColumns: Array[String]): DataFrame = amlOption match {
+    case Some(aml) => leaderboardAsSparkFrame(aml, extraColumns)
+    case None => throw new RuntimeException("The 'fit' method must be called at first!")
```
sebhrusen: are `String*` and `Array[String]` really different? I thought the compiler would automatically translate the first into the second: even Java does that.

py/src/ai/h2o/sparkling/ml/algos/H2OAutoML.py
```diff
@@ -68,8 +71,16 @@ def __init__(self,
...
         else:
             return None
+
+    def getLeaderboard(self, *extraColumns):
```
jakubhava: Thank you, this also fits better into the API!

py/src/ai/h2o/sparkling/ml/algos/H2OAutoML.py
jakubhava: 👍

jakubhava: I'm going to merge this change, just will change ``SparkSession._instantiatedSession`` to ``SparkSession.builder.getOrCreate()``. It is possible that user is using just spark context and the session does not exist.

It should work in a same way, if you have any comments, please let me know :)

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
mn-mikke: @sebhrusen variable parameters in Scala are represented by seqence. If you want to pass collection to variable parameters you must use a special operator for that. `getLeaderboard(collection: _*)`. The method overloading with `java.util.ArrayList[String]` must be there because of Py4J and Python client. 

The overloading with `Array[String]` is unnecessary, you are right!

PR #1878, opened 2020-02-25
[SW-1872] Remove deprecated nEstimators field and related methods on H2OXGBoost

jakubhava:


jakubhava:
I have already removed them in the deprecation stage, so they are gone :)

PR #1881, opened 2020-02-25
[SW-1998] Add Setter Methods of H2OConf to Documentation of Configuration Properties

mn-mikke:


mn-mikke:
@jakubhava It seems that I've fixed the problem with alternating color table. Instead of using subcells, I maneged to break lines by empty new line. 
```
+----------------------------------------------------+----------------+-------------------------------------------------+----------------------------------------+
| ``spark.ext.h2o.flow.scala.cell.async``            | ``false``      | ``setFlowScalaCellAsyncEnabled()``              | Decide whether the Scala cells in      |
|                                                    |                |                                                 | H2O Flow will run synchronously or     |
|                                                    |                | ``setFlowScalaCellAsyncDisabled()``             | Asynchronously. Default is             |
|                                                    |                |                                                 | synchronously.                         |
+----------------------------------------------------+----------------+-------------------------------------------------+----------------------------------------+
```


This approach was adding extra CSS `margin-bottom`, so I decided to modify CSS a bit and set `margin-bottom: 0px` for `th p` and `td p`. I hope it won't break anything.

Please let me know whether it's ok for you?

jakubhava:
Perfect!

PR #1882, opened 2020-02-25
[SW-1872][FOLLOWUP] Remove deprecated nEstimators getter and setter H2OXGBoost Python API

mn-mikke:
I would like to follow on the comment https://github.com/h2oai/sparkling-water/pull/1878#pullrequestreview-364090922 and complete the removal.


jakubhava:
Ah, sorry, I missed that comment. Thank you!

PR #1891, opened 2020-02-26
[SW-2008] Fix compile after removing code for external backend from H2O

jakubhava:
We were still referencing one class from H2O

### DIFF
--- a/extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/SerdeUtils.scala
+++ b/extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/SerdeUtils.scala
@@ -19,7 +19,7 @@ package ai.h2o.sparkling.extensions.serde
 
 import water.fvec.Vec
 
-object Utils extends ChunkSerdeConstants {
+object SerdeUtils extends ChunkSerdeConstants {
   private[sparkling] def expectedTypesToVecTypes(expectedTypes: Array[Byte], vecElemSizes: Array[Int]): Array[Byte] = {
     var vecCount = 0
     expectedTypes.flatMap {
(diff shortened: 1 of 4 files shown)


extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/Utils.scala
```diff
@@ -0,0 +1,36 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package ai.h2o.sparkling.extensions.serde
+
+import water.fvec.Vec
+
+object Utils extends ChunkSerdeConstants {
```
mn-mikke: note: `Utils` sounds a bit generic. I'm wondering whether we shouldn't go for anything more specific to serialization/deserialization although the class is in the `ai.h2o.sparkling.extensions.serde` namespace. 

This note is  really really minor and you can leave it as it's.

jakubhava: Good point, what about `SerdeUtils`?

mn-mikke: Sounds good to me ;-)

extensions/src/main/scala/ai/h2o/sparkling/extensions/serde/SerdeUtils.scala
```diff
@@ -0,0 +1,36 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+package ai.h2o.sparkling.extensions.serde
+
+import water.fvec.Vec
+
+object SerdeUtils extends ChunkSerdeConstants {
```
mn-mikke: I also wondering what is better? inheritance or `import ai.h2o.sparkling.extensions.serde. ChunkSerdeConstants._` WDYT?

jakubhava: I like `import ChunkSerdeConstants._` also, good catch

PR #1892, opened 2020-02-26
[SW-2009] No need to search for client ip in rest api mode

jakubhava:
The client ip configuration has now only effect when we run in the original client approach. 

This change is reflecting that. We do not need to do the search for client ip in rest API mode as we don't use client anyway.

Also this way we keep the code related to client connection close and separated from rest of the code

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalBackendUtils.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalBackendUtils.scala
@@ -21,7 +21,8 @@ import java.net.{InetAddress, NetworkInterface}
 
 import ai.h2o.sparkling.backend.shared.SharedBackendUtils
 import ai.h2o.sparkling.extensions.serde.ChunkSerdeConstants
-import org.apache.spark.h2o.H2OContext
+import org.apache.spark.SparkEnv
+import org.apache.spark.h2o.{H2OConf, H2OContext}
 import org.apache.spark.h2o.utils.NodeDesc
 import water.api.RestAPIManager
 import water.init.{HostnameGuesser, NetworkBridge}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalBackendUtils.scala
```diff
@@ -59,9 +60,8 @@ private[backend] trait ExternalBackendUtils extends SharedBackendUtils {
     }
   }
 
-
-  protected[backend] def startH2OClient(hc: H2OContext, nodes: Array[NodeDesc]): Unit = {
-    val conf = hc.getConf
+  protected[backend] def startH2OClient(hc: H2OContext, conf: H2OConf, nodes: Array[NodeDesc]): Unit = {
```
mn-mikke: Why do we need to pass `H2OConf` separately?

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
```diff
@@ -38,40 +38,24 @@ import scala.io.Source
...
+      ExternalH2OBackend.startH2OClient(hc, conf, nodes)
     }
     nodes
   }
 
   override def backendUIInfo: Seq[(String, String)] = {
     Seq(
       ("External backend YARN AppID", yarnAppId),
-      ("External IP", externalIP)
+      ("External IP", hc.getConf.h2oClusterHost)
```
mn-mikke: it seems to me that `externalIP` https://github.com/h2oai/sparkling-water/pull/1892/files#diff-f5f6b5c818efd05c837b2a930123d96eL118 used to contain also a port. Shouldn't `hc.getConf.h2oCluster` be used here?

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalBackendUtils.scala
jakubhava: Good question -> `hc.getConf` returns clone of `conf`, but the `conf` passed here is still accepts changes and is stored inside H2OContext after all the initialization

jakubhava: So we still need to update the right conf. Ideal would be to have val conf on `H2OContext` visible just in `ai.h2o.sparkling` but that we can do after we have all classes in the same package

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
jakubhava: Good catch, thanks!

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalBackendUtils.scala
mn-mikke: Ok, cool. Thanks for your explanation!

mn-mikke:
LGTM

PR #1893, opened 2020-02-26
[SW-1996] Benchmarks are Failing on Exception "VpcLimitExceeded"

mn-mikke:
I had to create a new VPC with a subnet, the current one didn't work


jakubhava:
Perfect, at least we have separated vpc for our benchmarks :)

Thank you!

PR #1897, opened 2020-02-27
[SW-2015] Simplify asH2OFrame in PySparkling

jakubhava:
This change makes conversion method in PySparkling more simple & robust:

- Avoid using JavaH2OContext, but use directly H2OContext ( non-necessary indirection, would like to get rid of JavaH2OContext at some point)
- On the back-end, RDD conversions are done in a way that we first convert them to DataFrame. We can do this directly on PySparkling side and avoid multiple conversion methods
- Because of previous point -> we don't need to have the special conversion functions for double & long in JavaH2OContext -> possible speedup of these conversions
- Allow conversion of big numbers -> treat them as strings
- Simpler and more maintainable code
- Single conversion path for all data (as we have on the back-end)


py/tests/unit/with_runtime_sparkling/test_conversions.py
```diff
@@ -78,15 +79,16 @@ def testFloatRDDToH2OFrame(spark, hc):
...
+    assert h2o_frame[2, 0] == 178
     unit_test_utils.asert_h2o_frame(h2o_frame, rdd)
 
 
 def testDoubleRDDToH2OFrame(spark, hc):
     rdd = spark.sparkContext.parallelize([0.5, 1.3333333333, 178])
     h2o_frame = hc.asH2OFrame(rdd)
     assert h2o_frame[0, 0] == 0.5
-    assert h2o_frame[1, 0] == 1.3333333333
+    assert pytest.approx(h2o_frame[1, 0]) == 1.3333333333
```
mn-mikke: What is the difference caused by this change?

py/tests/unit/with_runtime_sparkling/test_conversions.py
```diff
@@ -116,9 +118,11 @@ def testNumericRDDtoH2OFrameWithValueTooBig(spark, hc):
     min = _jvm().Long.MIN_VALUE - 1
     max = _jvm().Long.MAX_VALUE + 1
     rdd = spark.sparkContext.parallelize([1, min, max])
-    with pytest.raises(ValueError):
-        hc.asH2OFrame(rdd)
-
+    h2o_frame = hc.asH2OFrame(rdd)
+    assert h2o_frame[0, 0] == str(1)
+    assert h2o_frame[1, 0] == str(min)
```
mn-mikke: Are we casting unbounded Python int type to string during this conversion?

jakubhava: During conversion we check the range of integeres. If it fits into int, we create Dataframe with schema containing one integer, if it fits into long, we do the same for long.

If it's bigger, we just convert the number to string and this is the case.

py/tests/unit/with_runtime_sparkling/test_conversions.py
jakubhava: Haven't look into that, just fixed the test. I would say it's because we convert Spark RDD to DF on Python side and there might be some type conversion underneath.

The value is really same, but because it's double we can't compare it anymore exactly

jakubhava:
Merging to major only

py/tests/unit/with_runtime_sparkling/test_conversions.py
mn-mikke: ok, thanks!

py/tests/unit/with_runtime_sparkling/test_conversions.py
jakubhava: Previously we failed on exception that the number is too big, but I think we can handle it like this

mn-mikke: WDYT about DecimalType in Spark?

jakubhava: Ah, didn't know about this one. Will create JIRA for improvement for 3.30, WDYT?

mn-mikke: Yes, of course. I'm not pushing for that. Just mentioning as an option to think about.

jakubhava: Yup, good idea

PR #1898, opened 2020-02-27
[SW-2016] Expose getDomainValues on H2OMojoModel

jakubhava:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -46,6 +46,21 @@ class H2OMOJOModel(override val uid: String) extends H2OMOJOModelBase[H2OMOJOMod
 
   def getModelDetails(): String = $(modelDetails)
 
+  def getDomainValues(): Array[Array[String]] = {
```
mn-mikke: Shouldn't we return `Map[String, Array[String]]` where key is a column name instead?

ml/src/test/scala/ai/h2o/sparkling/ml/algos/MultinomialPredictionTestSuite.scala
```diff
@@ -113,4 +113,15 @@ class MultinomialPredictionTestSuite extends FunSuite with Matchers with SharedH
     assert(schema == expectedSchema)
     assert(schema == expectedSchemaByTransform)
   }
+
+  test("getDomainValues") {
```
mn-mikke: In fact, this test is testing `getDomainValuesForCol` method. Shouldn't we change name of this test and introduce another one for  `getDomainValues`?

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mn-mikke: Do Python and R converters support nested collections? Shouldn't we cover it with tests?

ml/src/test/scala/ai/h2o/sparkling/ml/algos/MultinomialPredictionTestSuite.scala
jakubhava: 👍  Good catch, I should introduce more tests anyway

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
jakubhava: Yup, better idea!

jakubhava: Good idea, will test

jakubhava: With this I think we should keep only one method `def getDomainValues(): Map[String, Array[String]]`, users can extract the specific column themselves

mn-mikke: Yup, makes sense!

jakubhava: Fixed Python & added test for Python & R

PR #1899, opened 2020-02-27
[SW-2018] Don't use allow_client flag in rest api mode

jakubhava:


jakubhava:
Just fyi: @michalkurka

PR #1902, opened 2020-02-28
[SW-2022] Start H2OContext on python side if the user didn't explicitly ask for it

jakubhava:
The motivation for this started when I was testing the algo api, I was running code:

```
from pyspark.sql.functions import log, col, min, max, mean, lit
dataset = spark.read.csv("file:///Users/kuba/devel/repos/sparkling-water/examples/smalldata/insurance.csv", header=True, inferSchema=True).withColumn("Offset", log(col("Holders")))

from pysparkling.ml import H2OGBM
gbm = H2OGBM(ntrees=2, seed=42, distribution="bernoulli", labelCol="capsule")
model = gbm.fit(dataset)
```

You can see that I do not call `H2OContext.getOrCreate()`. H2OContext is in this case created automatically inside the `fit` method on Scala. 

In Python & R implementation of H2OContext, we are now setting option `spark.ext.h2o.rest.api.based.client=true` (https://github.com/h2oai/sparkling-water/blob/14cfdcdb7712d9386444e79e81906e20961cac0d/py/src/ai/h2o/sparkling/H2OContext.py#L93), but not on the Scala backend. Due to this laziness ``H2OContext.getOrCreate`` hasn't been called in python & we do not set this option. 
I have ensured that we call, lazily when required, ``H2OContext.getOrCreate`` automatically on python side. Can be done by overriding the fit method on our algo wrappers.

### DIFF
--- a/doc/src/site/sphinx/migration_guide.rst
+++ b/doc/src/site/sphinx/migration_guide.rst
@@ -18,8 +18,8 @@ From 3.30 to 3.32
   been removed. Also, ``H2OContext`` can be created as ``H2OContext.getOrCreate()`` or ``H2OContext.getOrCreate(conf)``.
   The other variants of this method have been removed.
 
-- It is now required to explicitly create ``H2OContext`` before you run any of our exposed algorithm. Previously,
-  the algorithms would create the H2OContext on demand.
+- It is now required to explicitly create ``H2OContext`` before you run any of our exposed algorithms. Previously,
+  the algorithm would create the H2OContext on demand.
 
 From 3.28.1 to 3.30
 -------------------
--- a/doc/src/site/sphinx/migration_guide.rst
+++ b/doc/src/site/sphinx/migration_guide.rst
@@ -18,7 +18,7 @@ From 3.30 to 3.32
   been removed. Also, ``H2OContext`` can be created as ``H2OContext.getOrCreate()`` or ``H2OContext.getOrCreate(conf)``.
   The other variants of this method have been removed.
 
-- It is now required to explicitly created ``H2OContext`` before you run any of our exposed algorithm. Previously,
+- It is now required to explicitly create ``H2OContext`` before you run any of our exposed algorithm. Previously,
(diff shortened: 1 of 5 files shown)


mn-mikke:
I'm just wondering whether we shouldn't use `H2OContext.get()` on Scala side and fail if `H2OContext` doesn't exist. IMHO, we could experience more cases like this.

jakubhava:
Yeah, that is the other approach I was thinking about. Not sure which one is better though.

What I actually on this approach is you don't have to call H2OContext.getOrCreate and you can focus just on the algorithms.

jakubhava:
Yeah, that is the other approach I was thinking about. Not sure which one is better though.

What I actually on this approach is you don't have to call H2OContext.getOrCreate and you can focus just on the algorithms.

The approach you're proposing would be more explicit though.

jakubhava:
Maybe it's better to be explicit, it's just one line anyway

jakubhava:
@mn-mikke please have another look, at the end I think the explicit approach might be cleaner

mn-mikke:
I prefer to be explicit, once we enable users to stop adn create H2oContext mulitple times, the tracking of what configuration has been applied will be much more easier.

jakubhava:
I think we should add this just to major branch 3.30, I will back-port the bit from migration guide to rel-3.28.1 so it's in that doc as well

doc/src/site/sphinx/migration_guide.rst
```diff
@@ -18,6 +18,9 @@ From 3.30 to 3.32
   been removed. Also, ``H2OContext`` can be created as ``H2OContext.getOrCreate()`` or ``H2OContext.getOrCreate(conf)``.
   The other variants of this method have been removed.
 
+- It is now required to explicitly create ``H2OContext`` before you run any of our exposed algorithm. Previously,
```
mn-mikke: ``exposed algorithm`` -> ``exposed algorithms``

PR #1903, opened 2020-02-28
[SW-2023] Avoid repeated log messages

jakubhava:
Calling H2OContext.getOrCreate should provide all the warnings, such as:

```
20/02/28 16:11:55 WARN ExternalH2OBackend: To avoid non-deterministic behavior of Spark broadcast-based joins,
we recommend to set `spark.sql.autoBroadcastJoinThreshold` property of SparkSession to -1.
E.g. spark.conf.set("spark.sql.autoBroadcastJoinThreshold", -1)
We also recommend to avoid using broadcast hints in your Spark SQL code.
20/02/28 16:11:55 WARN ExternalH2OBackend: Increasing 'spark.locality.wait' to value 30000
20/02/28 16:11:55 WARN ExternalH2OBackend: To avoid non-deterministic behavior of Spark broadcast-based joins,
we recommend to set `spark.sql.autoBroadcastJoinThreshold` property of SparkSession to -1.
E.g. spark.conf.set("spark.sql.autoBroadcastJoinThreshold", -1)
We also recommend to avoid using broadcast hints in your Spark SQL code.
```

But additional calls of H2OContext.getOrCreate() should be already silent

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -598,6 +598,9 @@ object H2OContext extends Logging {
       val existingContext = instantiatedContext.get()
       if (existingContext != null) {
         val startedManually = existingContext.conf.isManualClusterStartUsed
+        if (conf.h2oCluster.isEmpty) {
+          throw new IllegalArgumentException("H2O Cluster endpoint has to be specified!")
+        }
         if (startedManually && connectingToNewCluster(existingContext, conf)) {
           val checkedConf = checkAndUpdateConf(conf)
           instantiatedContext.set(new H2OContextRestAPIBased(checkedConf).init())
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -591,7 +591,6 @@ object H2OContext extends Logging {
    * @return H2O Context
    */
   def getOrCreate(conf: H2OConf): H2OContext = synchronized {
-    val checkedConf = checkAndUpdateConf(conf)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -591,24 +591,26 @@ object H2OContext extends Logging {
...
-    val checkedConf = checkAndUpdateConf(conf)
     val isRestApiBasedClient = conf.getBoolean(SharedBackendConf.PROP_REST_API_BASED_CLIENT._1,
       SharedBackendConf.PROP_REST_API_BASED_CLIENT._2)
     val isExternalBackend = conf.runsInExternalClusterMode
     if (isExternalBackend && isRestApiBasedClient) {
       val existingContext = instantiatedContext.get()
       if (existingContext != null) {
         val startedManually = existingContext.conf.isManualClusterStartUsed
-        if (startedManually && connectingToNewCluster(existingContext, checkedConf)) {
+        if (startedManually && connectingToNewCluster(existingContext, conf)) {
```
mn-mikke: Don't we need to pass `checkedConf` to `connectingToNewCluster`. Dosn't `checkAndUpdateConf` contain any procedure that needs to be performed before `connectingToNewCluster ` is called?

jakubhava: Was thinking about it and actually no. The `checkAndUpdateConf` checks a lot of things, but we need only h2oCluster port and ip. The method just checks the field is not empty ( that I should improve), but otherwise we are good in this context

jakubhava: pushed the fix

mn-mikke: ok, thanks!

jakubhava: Thanks for asking, would forget about the empty check

PR #1904, opened 2020-02-28
[SW-2025] Extend REST Errors with Details from Server

mn-mikke:


core/src/test/scala/ai/h2o/sparkling/backend/external/RestApiUtilsTestSuite.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.spark.h2o.utils.SharedH2OTestContext
+import org.junit.runner.RunWith
+import org.scalatest.FunSuite
+import org.scalatest.junit.JUnitRunner
+
+@RunWith(classOf[JUnitRunner])
+class RestApiUtilsTestSuite extends FunSuite with SharedH2OTestContext {
+  override val createSparkContext: SparkContext = new SparkContext("local[*]", getClass.getSimpleName, defaultSparkConf)
+
+  if (createSparkContext.getConf.get("spark.ext.h2o.backend.cluster.mode") == "external") {
```
jakubhava: Could we do something like https://github.com/h2oai/sparkling-water/blob/9bb1b4a79b3ed610d3979fec97fb0f6ccdaa7d0f/core/build.gradle#L135 instead?

mn-mikke: Yes, will do. Great idea, thanks!

PR #1905, opened 2020-02-28
[SW-2023]FollowUp Avoid repeated log messages (#1903)

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -598,14 +597,17 @@ object H2OContext extends Logging {
...
-        if (conf.h2oCluster.isEmpty) {
-          throw new IllegalArgumentException("H2O Cluster endpoint has to be specified!")
-        }
-        if (startedManually && connectingToNewCluster(existingContext, conf)) {
-          val checkedConf = checkAndUpdateConf(conf)
-          instantiatedContext.set(new H2OContextRestAPIBased(checkedConf).init())
-          logWarning(s"Connecting to a new external H2O cluster : ${checkedConf.h2oCluster.get}")
+        if (startedManually) {
+          if (conf.h2oCluster.isEmpty) {
+            throw new IllegalArgumentException("H2O Cluster endpoint has to be specified!")
```
jakubhava: The problem was that we just need to check in case the external cluster was started manually. This option is always missing in automatic mode

PR #1906, opened 2020-02-28
[SW-2028] Cleanup: Remove extra repl classes for different spark versions

jakubhava:
Using my free time for some cleanup :) I believe we can simplify this code in this way. We have covered it with tests in scripts tests so as long as the tests are green, we should be good.

Upgrading to new Spark versions will be easier if we have less spark-specific code blocks


repl/src/main/scala/ai/h2o/sparkling/repl/H2OInterpreter.scala
```diff
@@ -71,6 +74,15 @@ class H2OInterpreter(sparkContext: SparkContext, sessionId: Int) extends BaseH2O
 
     settings
   }
+
+  private def getJarsForShell(conf: SparkConf): String = {
+    val localJars = conf.getOption("spark.repl.local.jars")
+    localJars.map(_.split(",")).map(_.filter(_.nonEmpty)).toSeq.flatten
```
mn-mikke: Would it be possible to tune this piece of code a bit? Here we have a sequence of `map`, `map`, `flatten`, `map`. Would it be possible to use just `flatmap` with 0-1 `map` operations?

jakubhava: Will do :)

PR #1908, opened 2020-03-02
[SW-2026] Expose H2OConf Getters and Setters in R

mn-mikke:


py/src/ai/h2o/sparkling/ExternalBackendConf.py
```diff
@@ -192,5 +192,5 @@ def setExternalExtraJars(self, paths):
         return self
 
     def setExternalCommunicationCompression(self, compression):
-        self._jconf.setExternalCommunicationCompression(self, compression)
+        self._jconf.setExternalCommunicationCompression(compression)
```
jakubhava: Good catch, thanks!

jakubhava: It would be good to write tests for all configurations so we avoid these mistakes. I think we already have jira for it, but that is not scope of this PR :)

mn-mikke:
@jakubhava I've manually tested all properties and made some changes, can you take a look one more time?

PR #1913, opened 2020-03-03
[SW-2039] Upgrade to Gradle 6.2.1

jakubhava:
New gradle fix release is out and Gradle community suggest to upgrade


jakubhava:
There should be nothing affecting us, but upgrading just to keep up with the latest version

PR #1916, opened 2020-03-04
[SW-2043] Deprecate get_grid_models, get_grid_models_params and get_grid_models_metrics params on H2OGridSearch

jakubhava:
This should also be last place where we had inconsistencies between Scala & Python client.

All methods should be in sync now ( except in Python & R we have method ``hc.asSparkFrame`` but in Scala we have ``hc.asDataFrame``) -> we need to rename the method on Scala side, but that is scope of another PR https://0xdata.atlassian.net/browse/SW-2044)

### DIFF
--- a/doc/src/site/sphinx/migration_guide.rst
+++ b/doc/src/site/sphinx/migration_guide.rst
@@ -228,7 +228,7 @@ Removal of Deprecated Methods and Classes
   ``setClusterInfoFile`` on ``H2OConf``.
 
 - In ``H2OGridSearch`` Python API, the methods: ``get_grid_models``, ``get_grid_models_params`` and `` get_grid_models_metrics``
-  are removed and replaced by ``getGridModels``, ``getGridModelsParams`` and `` getGridModelsMetrics``
+  are removed and replaced by ``getGridModels``, ``getGridModelsParams`` and `` getGridModelsMetrics``.
 
 
 From 3.28.0 to 3.28.1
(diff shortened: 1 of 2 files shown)


doc/src/site/sphinx/migration_guide.rst
```diff
@@ -224,6 +224,9 @@ Removal of Deprecated Methods and Classes
 
 - In ``H2OAutoML`` Python and Scala API, the member ``leaderboard()``/``leaderboard`` is replaced by the method ``getLeaderboard()``.
 
+- In ``H2OGridSearch`` Python API, the methods: ``get_grid_models``, ``get_grid_models_params`` and `` get_grid_models_metrics``
+  are removed and replaced by ``getGridModels``, ``getGridModelsParams`` and `` getGridModelsMetrics``
```
mn-mikke: nit: missing dot at EOL.

jakubhava:
Good point! I think that`getGridModelParams` sounds better to me as well, but originally the motivation was to stress out we are showing parameters for all models discovered by grid.

To be honest, not sure. WDYT?

mn-mikke:
Ok, let's leave as it's for now and minimize breaking changes. The change from plural would require changes in Scala API as well.

doc/src/site/sphinx/migration_guide.rst
jakubhava: thanks, fixed

PR #1917, opened 2020-03-04
[SW-2037] Remove deprecated setClusterConfigFile from H2OConf

mn-mikke:


jakubhava:
Ah, missed this one, luckily infra captured that. We need to update ExternalH2OBackend.scala with the new configuration

mn-mikke:
@jakubhava I've created separate PR https://github.com/h2oai/sparkling-water/pull/1920 to fix to problem to propagate it also to `rel` branch.

jakubhava:
Ups, one more: There is a python test still using the old method setClusterConfigFile

mn-mikke:
@jakubhava here is the fix https://github.com/h2oai/sparkling-water/pull/1933

PR #1923, opened 2020-03-04
[SW-2047] Deprecate options for disabling or enabling REST api on H2O worker nodes. It needs to be on because of REST client

jakubhava:
I have checked and actually these options are not used anywhere in the Sparkling Water, so wen user was specifying them, they did not have any effect & REST API was always open. Nevertheless, we should deprecate properly and remove in 3.30

### DIFF
--- a/py/src/ai/h2o/sparkling/InternalBackendConf.py
+++ b/py/src/ai/h2o/sparkling/InternalBackendConf.py
@@ -79,7 +79,7 @@ class InternalBackendConf(SharedBackendConfUtils):
         return self
 
     def setH2ONodeWebDisabled(self):
-        warnings.warn("Method 'setH2ONodeWebEnabled' is deprecated and will be removed in the next major release 3.30'.")
+        warnings.warn("Method 'setH2ONodeWebDisabled' is deprecated and will be removed in the next major release 3.30'.")
         self._jconf.setH2ONodeWebDisabled()
         return self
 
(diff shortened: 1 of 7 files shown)


r/src/R/ai/h2o/sparkling/InternalBackendConf.R
```diff
@@ -33,8 +33,6 @@ InternalBackendConf <- setRefClass("InternalBackendConf", methods = list(
 
     subseqTries = function() { invoke(jconf, "subseqTries") },
 
-    h2oNodeWebEnabled = function() { invoke(jconf, "h2oNodeWebEnabled") },
```
jakubhava: We don't need to deprecate as this was introduced in the same version as where we are deprecating  this option. Was never exposed to users

py/src/ai/h2o/sparkling/InternalBackendConf.py
```diff
@@ -73,10 +74,12 @@ def setSubseqTries(self, subseqTriesNum):
         return self
 
     def setH2ONodeWebEnabled(self):
+        warnings.warn("Method 'setH2ONodeWebEnabled' is deprecated and will be removed in the next major release 3.30'.")
         self._jconf.setH2ONodeWebEnabled()
         return self
 
     def setH2ONodeWebDisabled(self):
+        warnings.warn("Method 'setH2ONodeWebEnabled' is deprecated and will be removed in the next major release 3.30'.")
```
mn-mikke: `setH2ONodeWebEnabled` -> `setH2ONodeWebDisabled`

jakubhava: 👍

PR #1925, opened 2020-03-05
[SW-2050] Fix timezone handling

jakubhava:
Still need to write tests

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/converters/SparkDataFrameConverter.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/converters/SparkDataFrameConverter.scala
@@ -17,6 +17,8 @@
 
 package ai.h2o.sparkling.backend.converters
 
+import java.time.LocalDate
+
 import ai.h2o.sparkling.backend.external.{ExternalBackendConverter, ExternalBackendH2OFrameRelation, ExternalH2OBackend}
 import ai.h2o.sparkling.backend.internal.InternalBackendH2OFrameRelation
 import ai.h2o.sparkling.backend.shared.{Converter, Writer}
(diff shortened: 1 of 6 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/converters/SparkDataFrameConverter.scala
```diff
@@ -170,8 +173,8 @@ object SparkDataFrameConverter extends Logging {
           case _: DecimalType => con.put(idxH2O, row.getDecimal(idxField).doubleValue())
           case DoubleType => con.put(idxH2O, row.getDouble(idxField))
           case StringType => con.put(idxH2O, row.getString(idxField))
-          case TimestampType => con.put(idxH2O, row.getAs[java.sql.Timestamp](idxField))
-          case DateType => con.put(idxH2O, row.getAs[java.sql.Date](idxField))
+          case TimestampType => con.put(idxH2O, row.getLong(idxField))
+          case DateType => con.put(idxH2O, DateTimeUtils.daysToMillis(row.getInt(idxField)))
```
jakubhava: This call is using default timezone ( which we set in H2OContext.init)

jakubhava:
@mn-mikke should I close this PR?

mn-mikke:
I was about proceed with this PR, but the problem is hidden in H2O Python client, all timestamps and dates are displayed as UTC regardless of the time zone of the origin. We will have to fix the H2O client or convert dates and timestamps into UTC before sending it. Now, we are waiting for @st-pasha to shad some light on the problem with H2O Python client. IMHO, the original implementation on our side is fine. We can continue in this PR if we decide to implement a workaround on SW side, but we can close it for now, it's up to you.

jakubhava:
Oki thanks for explanation! Let's keep this one open then. I just thought you have done your work in different PR, wanted to avoid stalled PRs

PR #1931, opened 2020-03-05
[SW-2053] Deprecate clientWebEnabled and related setters (web needs to be open)

jakubhava:


py/src/ai/h2o/sparkling/SharedBackendConf.py
```diff
@@ -157,6 +157,7 @@ def ignoreSparkPublicDNS(self):
         return self._jconf.ignoreSparkPublicDNS()
 
     def clientWebEnabled(self):
+        warnings.warn("Method 'clientWebEnabled' is deprecated and will be removed in the next major release 3.30.")
         return self._jconf.clientWebEnabled()
```
mn-mikke: Can we return `true` directly?

py/src/ai/h2o/sparkling/SharedBackendConf.py
```diff
@@ -410,10 +411,12 @@ def setIgnoreSparkPublicDNSDisabled(self):
         return self
 
     def setClientWebEnabled(self):
+        warnings.warn("Method 'setClientWebEnabled' is deprecated and will be removed in the next major release 3.30.")
         self._jconf.setClientWebEnabled()
```
mn-mikke: Do we have to call the  scala property here?

py/src/ai/h2o/sparkling/SharedBackendConf.py
```diff
@@ -410,10 +411,12 @@ def setIgnoreSparkPublicDNSDisabled(self):
...
         return self
 
     def setClientWebDisabled(self):
+        warnings.warn("Method 'setClientWebDisabled' is deprecated and will be removed in the next major release 3.30.")
         self._jconf.setClientWebDisabled()
```
mn-mikke: Do we have to call the  scala property here?

doc/src/site/sphinx/migration_guide.rst
```diff
@@ -24,6 +24,11 @@ From 3.28.1 to 3.30
 - It is now required to explicitly create ``H2OContext`` before you run any of our exposed algorithms. Previously,
   the algorithm would create the H2OContext on demand.
 
+- It is no longer possible to disable web (REST API endpoints) on the client node as we require the Rest API
```
mn-mikke: This could be mentioned in the section 3.28 -> 3.28.1 and a notion about removal would stay?

core/src/main/scala/ai/h2o/sparkling/backend/shared/SharedBackendConf.scala
```diff
@@ -122,6 +123,7 @@ trait SharedBackendConf {
 
   def ignoreSparkPublicDNS: Boolean = sparkConf.getBoolean(PROP_CLIENT_IGNORE_SPARK_PUBLIC_DNS._1, PROP_CLIENT_IGNORE_SPARK_PUBLIC_DNS._2)
 
+  @DeprecatedMethod(version="3.30")
   def clientWebEnabled: Boolean = sparkConf.getBoolean(PROP_CLIENT_ENABLE_WEB._1, PROP_CLIENT_ENABLE_WEB._2)
```
mn-mikke: Can't we return always true and remove the property?

jakubhava: Yup!

jakubhava: Yup!

py/src/ai/h2o/sparkling/SharedBackendConf.py
jakubhava: 👍 nope

py/src/ai/h2o/sparkling/SharedBackendConf.py
jakubhava: Yup

doc/src/site/sphinx/migration_guide.rst
jakubhava: I think that we are deprecating in 3.28.1 & that option will be completely removed in 3.30

I believe it should be in section 3.28.1 -> 3.30

mn-mikke: ok, np.

PR #1942, opened 2020-03-09
[SW-2067] Propagate Timezone Settings from Spark to H2O

mn-mikke:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -170,6 +173,23 @@ abstract class H2OContext private(private val conf: H2OConf) extends Logging wit
     this
   }
 
+  private def setTimeZone(): Unit = {
+    if (conf.isSparkTimeZoneFollowed) {
+      val sparkTimeZone = sparkSession
+        .sparkContext
+        .getConf
+        .getOption("spark.sql.session.timeZone")
```
mn-mikke: This option was introduced in Spark 2.2

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -151,6 +153,7 @@ abstract class H2OContext private(private val conf: H2OConf) extends Logging wit
     // Init the H2O Context in a way provided by used backend and return the list of H2O nodes in case of external
     // backend or list of spark executors on which H2O runs in case of internal backend
     initBackend()
+    setTimeZone()
```
jakubhava: This fix is targeted for the fix release. I think that in that case we need to be also able to set timezone directly (no REST).

So ideally, I would keep this PR as it is for master (3.30) but on release branch, set timezone directly if rest api mode is not enabled

mn-mikke: Ok, I've introduced two implementations via https://github.com/h2oai/sparkling-water/pull/1942/files#diff-a7201e955bc1543aea7b96100cdfafe0R194

The PR https://github.com/h2oai/sparkling-water/pull/1943 unifies implementation for the major version 3.30

PR #1944, opened 2020-03-11
[SW-2071] InternalH2OBackend Shouldn't Call setH2OCluster

mn-mikke:
`setH2OCluster()` method on h2oConf has a side effect and internally calls `setExternalClusterMode()`


jakubhava:
I would probably remove this side effect in the future, but the behavior needs to be deprecated first

mn-mikke:
@jakubhava I share your opinion.

PR #1947, opened 2020-03-12
[SW-2073] Deprecate Implicit Switch to External Backend when H2OConf setH2OCluster is Called

mn-mikke:


jakubhava:
I'm just thinking whether we can't deprecate it already for release 3.30?

mn-mikke:
I asked myself the same question. I'm not sure how many fix releases will go out before 3.30. My motivation for 3.32 is to get user more time to adopt this before hitting it directly as breaking change. If you think that it's ok to do this change in 3.30 and I will alter the PR.

jakubhava:
Yup, good thinking, there may not bee that many fix release before 3.30. Let's keep it for 3.32. Thanks for explanation!

PR #1948, opened 2020-03-13
[SW-2068] Remove REST API & client h2ocontext, make rest api the default one for conversions

jakubhava:
IgnoreSparkPublicDNS no longer required -> will deprecated in 3.28.1.2 https://0xdata.atlassian.net/browse/SW-2075. This is because we will always use leader node as the endpoint for the py/r clients to connect to ( we mimic what was in rest api mode already), we won't use h2o client for this ( I would like to avoid using H2O client in our code completely even though in non-rest api mode)

This change switches some operations to rest api as in 3.30 we are sure the web on all nodes is always enabled. 

There are more changes to be done, such as switch setting up logs via rest, etc. At this stage this change just removes different h2o context implementations and applies conversion via rest api globally. Hopefully simplifying code and ensuring we have just one single path how we convert data.

The ultimate goal is to use rest API for everything in both internal & external mode and just keep the H2O client available ( without using it internally at all) in Scala for people using the mixed api.


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -106,77 +105,42 @@ abstract class H2OContext private(private val conf: H2OConf) extends Logging wit
   }
   backendHeartbeatThread.setDaemon(true)
 
-  /** IP of H2O client */
-  private var localClientIp: String = _
```
jakubhava: avoid using vars

core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -106,77 +105,42 @@ abstract class H2OContext private(private val conf: H2OConf) extends Logging wit
...
   }
+  private val leaderNode = RestApiUtils.getLeaderNode(conf)
+  if (sparkContext.ui.isDefined) {
+    SparkSpecificUtils.addSparklingWaterTab(sparkContext)
+  }
+  logInfo(s"Sparkling Water ${BuildInfo.SWVersion} started, status of context: $this ")
+  updateUIAfterStart() // updates the spark UI
+  backendHeartbeatThread.start() // start backend heartbeats
+  private val (flowIp, flowPort) = {
+    val uri = ProxyStarter.startFlowProxy(conf)
```
jakubhava: Always start flow proxy so we have consistent behaviour in both rest api mode and non rest api.

jakubhava:
still wip

core/src/main/scala/ai/h2o/sparkling/backend/Converter.scala
```diff
@@ -0,0 +1,216 @@
...
+        }
+      }
+    }
+    rowIdx + 1
+  }
+
+  private def scheduleUpload[T](numPartitions: Int, rdd: H2OAwareRDD[T]): UploadPlan = {
+    val hc = H2OContext.ensure("H2OContext needs to be running")
+    val nodes = hc.getH2ONodes()
+    if (hc.getConf.runsInInternalClusterMode) {
```
jakubhava: On current master, all algorithms are using rest api and also rest conversions, even in internal backend. But the upload plan was always the same. If we use the information about the partition location, we get significant speedup in internal backend on rest api ( we transfer the data on the same node)

jakubhava:
@mn-mikke should be good to review :) Really sorry this change got this big

jakubhava:
@mn-mikke yup, I'll wait with this PR, no worries

core/src/main/scala/ai/h2o/sparkling/backend/H2ODataFrame.scala
```diff
@@ -54,21 +52,40 @@ private[backend] class ExternalBackendH2ODataFrame(val frame: H2OFrame, val requ
...
   override def compute(split: Partition, context: TaskContext): Iterator[InternalRow] = {
-    new H2ODataFrameIterator {
+    new H2OChunkIterator[InternalRow] {
       private val chnk = frame.chunks.find(_.index == split.index).head
-      override val reader: Reader = new ExternalBackendReader(frameKeyName, split.index, chnk.numberOfRows,
-        chnk.location, expectedTypes.get, selectedColumnIndices, h2oConf)
+      override val reader: Reader = new Reader(frameKeyName, split.index, chnk.numberOfRows,
+        chnk.location, expectedTypes, selectedColumnIndices, h2oConf)
+
+      private lazy val columnIndicesWithTypes: Array[(Int, SimpleType[_])] = selectedColumnIndices map (i => (i, bySparkType(types(i))))
```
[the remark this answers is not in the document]
jakubhava: Yup, agree. I just copied this code from original place which was created by former Sparkling Water developer :) Will update!

core/src/main/scala/ai/h2o/sparkling/backend/Converter.scala
```diff
@@ -0,0 +1,217 @@
...
+        ChunkSerdeConstants.EXPECTED_TIMESTAMP
+      } else if (clazz == classOf[org.apache.spark.ml.linalg.Vector]) {
+        ChunkSerdeConstants.EXPECTED_VECTOR
+      } else {
+        throw new RuntimeException("Unsupported class: " + clazz)
+      }
+    }
+  }
+
+  private def perDataFramePartition(conf: H2OConf,
```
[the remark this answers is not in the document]
jakubhava: Agree, will do!

core/src/main/scala/ai/h2o/sparkling/backend/Reader.scala
```diff
@@ -15,45 +15,69 @@
...
-import org.apache.spark.h2o.utils.ReflectionUtils._
-import org.apache.spark.h2o.utils.SupportedTypes
-import org.apache.spark.h2o.utils.SupportedTypes._
+package ai.h2o.sparkling.backend
+
+import ai.h2o.sparkling.extensions.serde.ChunkAutoBufferReader
+import ai.h2o.sparkling.frame.H2OChunk
+import org.apache.spark.h2o.H2OConf
+import org.apache.spark.h2o.utils.ReflectionUtils.NameOfType
+import org.apache.spark.h2o.utils.SupportedTypes.{Boolean, Byte, Date, Double, Float, Integer, Long, OptionalType, Short, SimpleType, String, Timestamp, UTF8, byBaseType}
```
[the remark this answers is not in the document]
jakubhava: 👍

core/src/main/scala/ai/h2o/sparkling/backend/H2ODataFrame.scala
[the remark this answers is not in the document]
jakubhava: Done

core/src/main/scala/ai/h2o/sparkling/backend/H2ODataFrame.scala
```diff
@@ -54,21 +53,48 @@ private[backend] class ExternalBackendH2ODataFrame(val frame: H2OFrame, val requ
...
+    ConversionUtils.expectedTypesFromClasses(javaClasses)
   }
 
   override def compute(split: Partition, context: TaskContext): Iterator[InternalRow] = {
-    new H2ODataFrameIterator {
+    new H2OChunkIterator[InternalRow] {
       private val chnk = frame.chunks.find(_.index == split.index).head
-      override val reader: Reader = new ExternalBackendReader(frameKeyName, split.index, chnk.numberOfRows,
-        chnk.location, expectedTypes.get, selectedColumnIndices, h2oConf)
+      override val reader: Reader = new Reader(frameId, split.index, chnk.numberOfRows,
```
mn-mikke: question to think about: Is reader used straight after it' creation? if not, shouldn't it be lazy val? It seems that the constructor of `Reader` creates connection to the REST server (https://github.com/h2oai/sparkling-water/pull/1948/files#diff-1a7545e7f6937e88d3f808a2d607e875R39).

core/src/main/scala/ai/h2o/sparkling/backend/H2ORDD.scala
```diff
@@ -15,25 +15,58 @@
...
+  private val h2oConf = hc.getConf
+
+  // Get product type before building an RDD
+  def this(@transient frame: H2OFrame)
+          (@transient hc: H2OContext) = this(frame, ProductType.create[A])(hc)
+
+  override def compute(split: Partition, context: TaskContext): Iterator[A] = {
+    new H2ORDDIterator {
+      private val chnk = frame.chunks.find(_.index == split.index).head
+      override val reader: Reader = new Reader(frameId, split.index, chnk.numberOfRows,
```
mn-mikke: question to think about: ditto

core/src/main/scala/ai/h2o/sparkling/backend/Writer.scala
```diff
@@ -0,0 +1,175 @@
...
+          case _: DecimalType => con.put(row.getDecimal(idxField).doubleValue())
+          case DoubleType => con.put(row.getDouble(idxField))
+          case StringType => con.put(row.getString(idxField))
+          case TimestampType => con.put(row.getAs[java.sql.Timestamp](idxField))
+          case DateType => con.put(row.getAs[java.sql.Date](idxField))
+          case v if ExposeUtils.isMLVectorUDT(v) => con.putVector(row.getAs[ml.linalg.Vector](idxField))
+          case _: mllib.linalg.VectorUDT => con.putVector(row.getAs[mllib.linalg.Vector](idxField))
+          case udt if ExposeUtils.isUDT(udt) => throw new UnsupportedOperationException(s"User defined type is not supported: ${udt.getClass}")
+          case unsupported => throw new UnsupportedOperationException(s"Data of type ${unsupported.getClass} are not supported for the conversion" +
+            s"to H2OFrame.")
```
mn-mikke: nit: `s"to H2OFrame."` -> `"to H2OFrame."`

core/src/main/scala/ai/h2o/sparkling/backend/Writer.scala
```diff
@@ -0,0 +1,175 @@
...
+  def putSparseVector(vector: ml.linalg.SparseVector): Unit = chunkWriter.writeSparseVector(vector.indices, vector.values)
+
+  def putDenseVector(vector: ml.linalg.DenseVector): Unit = chunkWriter.writeDenseVector(vector.values)
+
+  def putVector(vector: mllib.linalg.Vector): Unit = putVector(vector.asML)
+
+  def putVector(vector: ml.linalg.Vector): Unit = {
+    vector match {
+      case sparseVector: ml.linalg.SparseVector =>
+        putSparseVector(sparseVector)
```
mn-mikke: nit: `putSparseVector(sparseVector)` and `putDenseVector(denseVector)` can be on one the same line with condition

core/src/main/scala/ai/h2o/sparkling/backend/utils/RestCommunication.scala
```diff
@@ -15,11 +15,12 @@
 * limitations under the License.
 */
 
-package ai.h2o.sparkling.backend.external
+package ai.h2o.sparkling.backend.utils
 
 import java.io._
 import java.net.{HttpURLConnection, URI, URL}
 
+import ai.h2o.sparkling.backend.exceptions.{RestApiCommunicationException, RestApiException, RestApiNotReachableException, RestApiUnauthorisedException}
```
mn-mikke: nit: `ai.h2o.sparkling.backend.exceptions._`

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -0,0 +1,62 @@
...
+package ai.h2o.sparkling.backend.utils
+
+import java.io.File
+
+import org.apache.spark.h2o.H2OContext
+import water.api.ImportHiveTableHandler
+import water.api.ImportHiveTableHandler.HiveTableImporter
+import water.fvec.Frame
+
+trait H2OContextExtensions extends H2OContextUtils {
```
mn-mikke: Is the only place where `H2OContextUtils` are used. if so, couldn't they be merged with this class

core/src/main/scala/ai/h2o/sparkling/backend/H2OAwareRDD.scala
```diff
@@ -14,18 +14,23 @@
...
 import org.apache.spark.{Partition, TaskContext}
 
 import scala.reflect.ClassTag
 
-private[internal] class H2OAwareRDD[U: ClassTag](nodes: Array[NodeDesc], prev: RDD[U]) extends RDD[U](prev: RDD[U]) {
+private[backend] class H2OAwareRDD[U: ClassTag](nodes: Array[NodeDesc], prev: RDD[U]) extends RDD[U](prev: RDD[U]) {
 
-  override def getPreferredLocations(split: Partition): Seq[String] = nodes.map { nodeDesc =>
-    s"executor_${nodeDesc.hostname}_${nodeDesc.nodeId}"
+  override def getPreferredLocations(split: Partition): Seq[String] = {
```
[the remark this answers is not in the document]
jakubhava: Good idea, done! ( I couldn't do trait, but did abstract class instead)

core/src/main/scala/ai/h2o/sparkling/backend/H2OFrameRelation.scala
```diff
@@ -42,10 +42,10 @@ private[backend] case class ExternalBackendH2OFrameRelation(frame: H2OFrame, cop
   override val schema: StructType = createSchema(frame, copyMetadata)
 
   override def buildScan(): RDD[Row] =
-    new ExternalBackendH2ODataFrame(frame)(h2oContext).asInstanceOf[RDD[Row]]
+    new H2ODataFrame(frame)(hc).asInstanceOf[RDD[Row]]
```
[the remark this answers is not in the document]
jakubhava: 👍

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
jakubhava: The motivation for the separation was to keep public methods in extensions which are user facing, and in utils, just utilities. Bud I guess at this moment this is not necessary. Merging together 👍

core/src/main/scala/ai/h2o/sparkling/backend/H2ODataFrame.scala
jakubhava: Great catch!

Reader should be used whenever we run any action on the obtained dataframe ( after h2o frame -> spark frame conversion). So therefore, the reader should be lazy.

Thanks!

PR #1950, opened 2020-03-13
[SW-2050] Fix Timezone Handling via Conversions to UTC

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendReader.scala
```diff
@@ -18,13 +18,14 @@
...
 import ai.h2o.sparkling.backend.shared.Reader
+import org.apache.spark.h2o.H2OConf
 import water.fvec.{Chunk, Frame, Vec}
 import water.parser.BufferedString
 import water.{DKV, Key}
 
 import scala.language.postfixOps
 
-class InternalBackendReader(override val keyName: String, override val chunkIdx: Int) extends Reader {
+class InternalBackendReader(override val keyName: String, override val chunkIdx: Int, val conf: H2OConf) extends Reader {
```
jakubhava: Is the conf argument required?

mn-mikke: it's reqired by https://github.com/h2oai/sparkling-water/pull/1950/files#diff-14e9f8743d500fef4414bd86722b3e63R27 since `InternalBackendReader` transitively  inherits from Timezone conversions.

jakubhava: 👍 oki, thanks for explanation!

core/src/main/scala/ai/h2o/sparkling/backend/converters/TimeZoneConversions.scala
```diff
@@ -0,0 +1,45 @@
...
+    val defaultTimeZone = DateTimeUtils.defaultTimeZone().getID
+    conf.get("spark.sql.session.timeZone", defaultTimeZone)
+  }
+
+  def fromSparkTimeZoneToUTC(timestamp: Long): Long = DateTimeUtils.fromUTCTime(timestamp, sparkTimeZone())
+
+  def fromSparkTimeZoneToUTC(timestamp: Timestamp): Long = fromSparkTimeZoneToUTC(timestamp.getTime * 1000) / 1000
+
+  def fromSparkTimeZoneToUTC(date: Date): Long = {
+    DateTimeUtils.fromUTCTime(date.getTime * 1000, DateTimeUtils.defaultTimeZone().getID) / 1000
```
mn-mikke: Spark `DateType` is based on a default time zone and not on `spark.sql.session.timeZone`.

mn-mikke:
@jakubhava Tests are finally passing, please can you take a look one more time.

core/src/main/scala_spark_others/ai/h2o/sparkling/SparkTimeZone.scala
```diff
@@ -0,0 +1,25 @@
...
+* limitations under the License.
+*/
+
+package ai.h2o.sparkling
+
+import java.util.TimeZone
+import ai.h2o.sparkling.utils.SparkSessionUtils
+
+object SparkTimeZone {
+  def current(): TimeZone = TimeZone.getTimeZone(SparkSessionUtils.active.sessionState.conf.sessionLocalTimeZone)
```
jakubhava: SparkTimeZone is same for Spark 2.2, 2.3 and spark_others. In this case we can just define spark_others

mn-mikke: Ok, will do.

jakubhava: Ah, sorry, my bad. We need to have those files there. As long the specific directory for some spark exists, we need to have the classes there. Will create the followup fix

PR #1951, opened 2020-03-16
[SW-2075] Deprecate ignorePublicDNS option as in 3.30 it is no longer required

jakubhava:


r/src/R/ai/h2o/sparkling/SharedBackendConf.R
```diff
@@ -111,7 +111,10 @@ SharedBackendConf <- setRefClass("SharedBackendConf", methods = list(
 
     clientNetworkMask = function() { ConfUtils.getOption(invoke(jconf, "clientNetworkMask")) },
 
-    ignoreSparkPublicDNS = function() { invoke(jconf, "ignoreSparkPublicDNS") },
+    ignoreSparkPublicDNS = function() {
+      print("Method 'ignoreSparkPublicDNS' is deprecated in release 3.30 without replacement.")
```
mn-mikke: I'm not sure how to generate warnings in R properly, but it seems that there is a method for that https://stat.ethz.ch/R-manual/R-devel/library/base/html/warning.html

py/src/ai/h2o/sparkling/SharedBackendConf.py
```diff
@@ -314,6 +314,7 @@ def ignore_spark_public_dns(self):
         return self.ignoreSparkPublicDNS()
 
     def ignoreSparkPublicDNS(self):
+        warnings.warn("Method 'ignoreSparkPublicDNS' is deprecated in release 3.30 without replacement.")
```
mn-mikke: `"Method 'ignoreSparkPublicDNS' is deprecated in release 3.30 without replacement."` -> `"Method 'ignoreSparkPublicDNS' is deprecated. It will be removed in the release 3.30 without replacement."`?

jakubhava: 👍

PR #1957, opened 2020-03-18
[SW-2050][FollowUp] Fix Timezone Handling via Conversions to UTC (#1950)

jakubhava:


core/src/main/scala/ai/h2o/sparkling/backend/converters/TimeZoneConversions.scala
```diff
@@ -25,15 +25,15 @@ import org.apache.spark.sql.catalyst.util.DateTimeUtils
 trait TimeZoneConversions {
   def sparkTimeZone: TimeZone
 
-  def fromSparkTimeZoneToUTC(timestamp: Long): Long = DateTimeUtils.fromUTCTime(timestamp, sparkTimeZone.getID)
+  def fromSparkTimeZoneToUTC(timestamp: Long): Long = DateTimeUtils.toUTCTime(timestamp, sparkTimeZone.getID)
```
jakubhava: I believe we were using opposite methods in this class

core/src/test/scala/org/apache/spark/h2o/converters/DataFrameConverterTest.scala
```diff
@@ -102,20 +103,33 @@ class DataFrameConverterTest extends FunSuite with SharedH2OTestContext {
...
+    assert(df.collect().sameElements(rddBack.collect()))
+    h2oFrame.delete()
+  }
+
+  test("H2OFrame[T_TIME] UTC to DataFrame[TimestampType]") {
     val fname: String = "testTime.hex"
     val colNames: Array[String] = Array("C0")
     val chunkLayout: Array[Long] = Array(2L, 2L)
     val data: Array[Array[Long]] = Array(Array(1428517563L, 1428517564L), Array(1428517565L, 1428517566L))
+      .map(_.map(t => DateTimeUtils.toUTCTime(t * 1000, SparkTimeZone.current().getID)))
```
jakubhava: We need to simulate H2O import ( treat time as in UTC)

core/src/main/scala/ai/h2o/sparkling/backend/converters/TimeZoneConversions.scala
mn-mikke: This was made like this on purpose. It seems that Spark internal methods has opposite meaning.

mn-mikke: The python tests I added should show us the correct behavior

PR #1964, opened 2020-03-19
[SW-2086] Move classes from org.apache.spark.h2o.utils to ai.h2o.sparkling

jakubhava:
The reason why I'm moving the classes to ai.h2o.sparkling package is to have all internal classes in the `backend` package and so we can restrict access to all of them for the users. Since we will start creating Scala API I would like to have clear distinction.

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/utils/SupportedTypes.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/utils/SupportedTypes.scala
@@ -21,7 +21,6 @@ import org.apache.spark.sql.types._
 import org.apache.spark.unsafe.types.UTF8String
 import water.fvec.Vec
 
-import scala.language.{implicitConversions, postfixOps}
 import scala.reflect.runtime.universe._
 
 /**
@@ -74,8 +73,6 @@ private[backend] object SupportedTypes extends Enumeration {
 
   import java.{lang => jl, sql => js}
 
-  private val ZeroTime = new js.Timestamp(0L)
-
   private def onNAthrow(what: String) = {
     throw new IllegalArgumentException(s"$what value missing")
   }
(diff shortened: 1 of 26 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/utils/SupportedTypes.scala
```diff
@@ -82,21 +82,18 @@ object SupportedTypes extends Enumeration {
...
-  val Float     = SimpleType[scala.Float  ] (Vec.T_NUM,  FloatType,     classOf[jl.Float    ], onNAreturn(scala.Float.NaN), typeOf[Float])
-  val Double    = SimpleType[scala.Double ] (Vec.T_NUM,  DoubleType,    classOf[jl.Double   ], onNAreturn(scala.Double.NaN), typeOf[Double])
-  val Timestamp = SimpleType[js.Timestamp ] (Vec.T_TIME, TimestampType, classOf[js.Timestamp], onNAthrow)
-  val Date      = SimpleType[js.Date      ] (Vec.T_TIME, DateType,      classOf[js.Date     ], onNAthrow)
-  val String    = SimpleType[String       ] (Vec.T_STR,  StringType,    classOf[String],       onNAreturn(null), typeOf[String])
-  // TODO(vlad): figure it out, why
-  val UTF8      = SimpleType[UTF8String   ] (Vec.T_STR,  StringType,    classOf[String],       onNAreturn(null), typeOf[UTF8String])
-  // scalastyle:on
-  
+  val Boolean = SimpleType[scala.Boolean](Vec.T_NUM, BooleanType, classOf[jl.Boolean], onNAreturn(false), typeOf[Boolean])
```
mn-mikke: Isn't the old formatting more legible?

jakubhava: Oki, can revert this part back

jakubhava: Actually more I think about it, it makes more sense to me to keep the new version. At least we can really apply automatic formatting without thinking about it.

PR #1965, opened 2020-03-19
[SW-2082] Deprecate hex.ModelUtils.classify in favour of SW Algo API

jakubhava:


doc/src/site/sphinx/migration_guide.rst
```diff
@@ -21,6 +21,9 @@ From 3.30 to 3.32
 - The ``setH2OCluster(ip, port)`` method on ``H2OConf`` in all APIs doesn't implicitly set the external backend anymore.
   The method ``setExternalClusterMode()`` must be called explicitly.
 
+- The method ``classify`` in the ``hex.ModelUtils`` object is removed. Please use Sparkling Water algorithm API to train
```
mn-mikke: Maybe, it would worth to mention that this affects only scala API?

jakubhava:
Nope, it is not documented anywhere

doc/src/site/sphinx/migration_guide.rst
jakubhava: will do, thanks!

PR #1966, opened 2020-03-19
[SW-2084] Deprecate DeepLearningSupport and GBMSupport in favor of SW Algo API

jakubhava:


doc/src/site/sphinx/migration_guide.rst
```diff
@@ -21,6 +21,10 @@ From 3.30 to 3.32
 - The ``setH2OCluster(ip, port)`` method on ``H2OConf`` in all APIs doesn't implicitly set the external backend anymore.
   The method ``setExternalClusterMode()`` must be called explicitly.
 
+- The method ``DLModel`` in ``water.support.DeepLearningSupport`` is removed. Please use Sparkling Water algorithm API
```
mn-mikke: I would maybe mention that it's related only to Scala API.

core/src/main/scala/water/support/DeepLearningSupport.scala
```diff
@@ -42,6 +44,7 @@ trait DeepLearningSupport {
     * @tparam T H2O Frame Type
     * @return Deep Learning Model
     */
+  @DeprecatedMethod("Sparkling Water Algorithm API to train and score H2O models", "3.32")
```
mn-mikke: Maybe, I would add reference to the `H2ODeepLearning` class.

core/src/main/scala/water/support/GBMSupport.scala
```diff
@@ -39,6 +41,7 @@ trait GBMSupport {
     * @tparam T H2O Frame Type
     * @return Gradient Boosting Model
     */
+  @DeprecatedMethod("Sparkling Water Algorithm API to train and score H2O models", "3.32")
```
mn-mikke: Maybe, I would add reference to the `H2OGBM` class.

doc/src/site/sphinx/migration_guide.rst
jakubhava: 👍

core/src/main/scala/water/support/DeepLearningSupport.scala
jakubhava: 👍

core/src/main/scala/water/support/GBMSupport.scala
jakubhava: 👍

PR #1967, opened 2020-03-19
[SW-2089] Expose only H2OFrame, hide remaining internal API

jakubhava:
I'm sorry for another mass moving (this time only moving), hopefully last one.

The reason for this is that allmost all classes in `job`, `frame` and `model` packages in `ai.h2o.sparkling` should not be user facing. Currently the are public and for example, the `H2OModel` being exposed could really confuse the users.

I have therefore moved all the classes which should be internal to `backend` package and left just `H2OFrame` in `ai.h2o.sparkling` ( As this will be user facing class a lot, I think it should be located conveniently in ai.h2o.sparkling directly).

Also we can afford to hide these now as this code hasn't been released so far.

Also I have moved (now private) H2OModel to ml package as it it should be there anyways

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/backend/H2OModel.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/backend/H2OModel.scala
@@ -15,7 +15,7 @@
 * limitations under the License.
 */
 
-package ai.h2o.sparkling.model
+package ai.h2o.sparkling.ml.backend
 
 import java.io.File
 import java.nio.file.Files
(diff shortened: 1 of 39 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/backend/H2OModel.scala
```diff
@@ -15,7 +15,7 @@
 * limitations under the License.
 */
 
-package ai.h2o.sparkling.model
+package ai.h2o.sparkling.ml.backend
```
mn-mikke: the name `ai.h2o.sparkling.ml.backend` is very similar to `ai.h2o.sparkling.backend`. Would it be possible to use a more distinctive name? E.g. : `ai.h2o.sparkling.ml.internals` WDYT?

jakubhava: Yup, good idea, like it 👍

PR #1968, opened 2020-03-19
[SW-2061] Improve and test getCurrentMetrics on H2OModel

jakubhava:
While ago I have promised to write tests for getCurrentMetrics. I have tried to improve it a bit as well ( this code hasn't been released, so we can afford it)

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
@@ -179,7 +179,7 @@ class H2OGridSearch(override val uid: String) extends Estimator[H2OMOJOModel]
     }
 
     val modelMetricPair = gridModels.map { model =>
-      (model, model.getCurrentMetrics(getNfolds(), getSplitRatio()).find(_._1 == metric).get._2)
+      (model, model.currentMetrics.find(_._1 == metric).get._2)
     }
 
     val ordering = if (metric.higherTheBetter) Ordering.Double.reverse else Ordering.Double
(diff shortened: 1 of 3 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/backend/H2OModel.scala
```diff
@@ -107,9 +113,13 @@ private[sparkling] object H2OModel extends RestCommunication {
         val stringElements = v.asScala.flatMap(stringifyJSON)
         val arrayAsString = stringElements.mkString("[", ", ", "]")
         Some(arrayAsString)
-      case _ =>
-        // don't put more complex type to output yet
-        None
+      case _: JsonNull => None
+      case v: JsonObject =>
+        if (v.has("name")) {
```
jakubhava: To ensure that also complex values ( such as frame names, key names etc are return to the list) Mainly needed to discover validation frame

ml/src/main/scala/ai/h2o/sparkling/ml/backend/H2OMetricsHolder.scala
```diff
@@ -17,6 +17,20 @@
...
+private[sparkling] case class H2OMetricsHolder(trainingMetrics: TrainingMetrics,
+                                               validationMetrics: ValidationMetrics,
+                                               crossValidationMetrics: CrossValidationMetrics) {
+}
+
+class H2OMetrics(previousMap: Map[H2OMetric, Double]) extends mutable.HashMap[H2OMetric, Double] {
+  this ++= previousMap.toSeq
+}
+
+class TrainingMetrics(previousMap: Map[H2OMetric, Double]) extends H2OMetrics(previousMap)
```
mn-mikke: What is the purpose of these subclasses? Why `Map[H2OMetric, Double]` wasn't sufficient?

jakubhava: Introduced mainly because of tests to verify that the current metrics return correct object. If you have different idea, please let me know :)

mn-mikke: I assume traning metrics, validation metrics and cross validation metrics are different. I would be possible to test them based on content and according to the implementation of `currentMetrics` it seems that it would be possible bases on references?

jakubhava: Good point! I was so focused on this that I even didn't realize this is an option!

jakubhava: I'm not really sure why I haven't about this at first:

``assert(model.metrics.trainingMetrics == model.currentMetrics)``

Thank you!

PR #1976, opened 2020-03-22
[SW-2095] Remove and replace remaining reference to H2O algo API by SW Algo API ( ensure all SW tests are running without H2O client)

jakubhava:
Documentation updated.

TODO: update integ & examples tests scala tests


core/src/test/scala/org/apache/spark/h2o/ConfigurationPropertiesTestSuite.scala
```diff
@@ -158,14 +158,6 @@ class ConfigurationPropertiesTestSuite_SetNotifyLocalViaClientExtraProperties_Lo
   }
 }
 
-@RunWith(classOf[JUnitRunner])
```
jakubhava: this test has no effect when running without the client ( but we still test when running in local mode which is covering very similar scenario, I think it is just fine to remove this test)

jakubhava:
Will do several smaller PRs instead of one big

PR #1977, opened 2020-03-23
[SW-2107] Ensure all scala unit tests can run without client

jakubhava:
The removed test fails as there is no H2O client running.

We still test when running in local mode which is covering very similar scenario, I think it is just fine to remove this test


jakubhava:
Requires more work

PR #1979, opened 2020-03-23
[SW-2109] Ensure AirlinesDemo runs without the client

jakubhava:
I will probably create jira for each integ test separately to keep the changes small. The goal is to switch the code so it uses SW algo api and can run without the client. Once all integ tests are switched, I will switch rest api mode on scala tests by default.

Also when changing the tests I will also try to clean them up and ensure we do not have duplicate tests in our codebase to speedup the tests.

Also when looking at integ tests, almost all scripts tests are duplicating code in integTests. I will move towards moving code to integration tests and removing scripts tests as they are hard to maintain ( and in all except one just duplicates the integ tests)


examples/src/scriptsTest/scala/ai/h2o/sparkling/ScriptsTestSuite.scala
```diff
@@ -157,21 +157,6 @@ class CityBikeSharing extends ScriptsTestHelper {
   }
 }
 
-@RunWith(classOf[JUnitRunner])
```
jakubhava: This script is almost identical to AirlinesWithWeatherDemo -> no point in keeping duplicate code and testing the same code twice ( it just predicts arr delay, the demo predicts dep delay)

examples/src/main/scala/ai/h2o/sparkling/examples/AirlinesWithWeatherDemo2.scala
```diff
@@ -1,175 +0,0 @@
-/*
-* Licensed to the Apache Software Foundation (ASF) under one or more
```
jakubhava: This is almost identical to AirlinesWithWeatherDemo except it runs GBM at the end. Added GBM to AirlinesWithWeatherDemo to avoid duplicated tests

examples/src/integTest/scala/ai/h2o/sparkling/local/AirlinesWithWeatherDemoSuite.scala
```diff
@@ -20,9 +37,3 @@ class AirlinesWithWeatherDemoSuite extends FunSuite with IntegTestHelper {
     )
   }
 }
-
-object AirlinesWithWeatherDemoTest extends IntegTestStopper {
```
jakubhava: unused

PR #1982, opened 2020-03-23
[SW-2112] Update getOrCreate method after enabling rest api in py/r as well

jakubhava:
Also internal backend can run in rest api mode, ensure this is reflected in this method

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
@@ -477,7 +477,7 @@ object H2OContext extends Logging {
         val startedManually = existingContext.conf.isManualClusterStartUsed
         if (isExternalBackend && startedManually) {
           if (conf.h2oCluster.isEmpty) {
-            throw new IllegalArgumentException("H2O Cluster endpoint has to be specified!")
+            throw new IllegalArgumentException("H2O cluster endpoint has to be specified!")
           }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -457,12 +457,25 @@ object H2OContext extends Logging {
    * @return H2O Context
    */
   def getOrCreate(conf: H2OConf): H2OContext = synchronized {
-    val isExternalBackend = conf.runsInExternalClusterMode
-    if (isExternalBackend && !H2OClientUtils.isH2OClientBased(conf)) {
+    if (H2OClientUtils.isH2OClientBased(conf)) {
+      if (instantiatedContext.get() == null)
```
mn-mikke: qq: Where has the context been created?

jakubhava: The context is being created in case of client here:

```
      if (instantiatedContext.get() == null)
        if (H2O.API_PORT == 0) { // api port different than 0 means that client is already running
          val checkedConf = checkAndUpdateConf(conf)
          instantiatedContext.set(new H2OContext(checkedConf))
``` 
The last line

mn-mikke: oh, ok thanks! I'm blind.

PR #1984, opened 2020-03-24
[SW-2114] Make Prostate example use SW API

jakubhava:


examples/src/main/scala/ai/h2o/sparkling/examples/AmazonFineFood.scala
```diff
@@ -1,124 +0,0 @@
-/*
```
jakubhava: Remove as complete example is available in h2o-tutorials repo

PR #1989, opened 2020-03-25
[SW-2119] Make SW Compatible with Older Versions of Steam

mn-mikke:


mn-mikke:
@jakubhava  the problem is not with the current version of steam, but due to this change in the stracture the older versions of steam are not able run newer SW versions. Not all customers can upgrade steam easily.

dist/build.gradle
```diff
@@ -174,6 +175,10 @@ task copyFilesForZipDistribution {
         copyRecursive("bin/", "$zipDir/bin")
         copyRecursive("doc/build/site/", "$zipDir/doc/build/site")
         copyRecursive("docker/", "$zipDir/docker")
+
+        // TODO: The below lines should be removed in future. For now, they ensure compatibility with older versions of Steam.
```
jakubhava: @mn-mikke please create JIRA fro removal

jakubhava:
Yup, realized it's on older versions -> approved

dist/build.gradle
mn-mikke: https://0xdata.atlassian.net/browse/SW-2121

bilcus:
> I would prefer if Steam handles this -> it should be matter of checking whether the file exists on the given path right? @bilcus

I will change Steam to be more flexible about the location of the files.

jakubhava:
Thank you @bilcus!

PR #1998, opened 2020-03-26
[SW-2127] Fix compile & running of micro benchmarks

jakubhava:
A while ago I discovered that the micro benchmarks have wrong path


core/build.gradle
```diff
@@ -146,6 +146,7 @@ processResources.dependsOn ':sparkling-water-assembly-extensions:embeddedAssembl
 
 bench {
     systemProperty "spark.ext.h2o.backend.cluster.mode", detectBackendClusterMode()
+    systemProperty "spark.ext.h2o.rest.api.based.client", "false"
```
jakubhava: We first need to switch H2OContext methods asH2OFrame etc to rest api before we can allow client mode in sw core tests

PR #1999, opened 2020-03-26
[SW-2125] Introduce Update Plan Reflecting the Final Layout of H2O Chunks

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/backend/Writer.scala
```diff
@@ -176,4 +158,11 @@ private[backend] object Writer {
   private def getNonEmptyPartitions(partitionSizes: Map[Int, Int]): Seq[Int] = {
     partitionSizes.keys.toSeq.sorted
   }
+
+  private def getUploadPlan(conf: H2OConf, numberOfPartitions: Int): UploadPlan = {
+    val endpoint = getClusterEndpoint(conf)
+    val parameters = Map("number_of_chunks" -> numberOfPartitions)
+    val rawPlan = query[UploadPlanV3](endpoint, Paths.UPLOAD_PLAN, conf, parameters)
+    rawPlan.layout.map(ca => ca.chunk_id -> NodeDesc(ca.node_idx.toString, ca.ip, ca.port)).toMap
```
jakubhava: NIT: why `ca`?

mn-mikke: `ca` -> `chunkAssignment`. I will use a full name, so it's clear.

PR #2001, opened 2020-03-27
[SW-2131] Speed up of integration tests by sharing the context

jakubhava:


core/src/integTest/scala/ai/h2o/sparkling/IntegrationTestSuite.scala
```diff
@@ -0,0 +1,153 @@
...
+    val updatedFrame = h2oFrame.add(h2oFrame)
+
+    val convertedDf = hc.asDataFrame(updatedFrame)
+    convertedDf.collect()
+
+    assert(convertedDf.count() == h2oFrame.numRows())
+    assert(convertedDf.columns.length == h2oFrame.names().length)
+  }
+
+  test("Task killed but frame still converted successfully") {
```
mn-mikke: The tests were moved from different test suites with names identifying the scope of testing. Shouldn't the test names also contain information defined by the former suite names? ("H2OFrameToDataFrameHA")

PR #2002, opened 2020-03-28
[SW-1391] Switch to Scala formatter

jakubhava:
Weekend project :) Finally found a good way how to apply scalafmt https://github.com/diffplug/spotless/tree/master/plugin-gradle

Also this plugin is great in a way that it can easily format all other code types.

We no longer need to depend on idea and have unified formatting in code. The checks depends on build so whenever you build sw these check are run automatically and if any violation is found, you can run `gradlew spotlessApply` to fix those.

In the first commit of this change I'm applying this plugin so you can see the plugin code.

 In the second commit I'm applying the formatting fixes ( all automatic based on specified scalafmt). Please have a look as you have more experience with Spark formatting if it looks as you like. We can additionally improve the scala format


benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/BenchmarkBase.scala
```diff
@@ -158,8 +166,10 @@ abstract class BenchmarkBase[TInput](context: BenchmarkContext) {
   }
 }
 
-abstract class AlgorithmBenchmarkBase[TInput](context: BenchmarkContext, algorithm: AlgorithmBundle)
-  extends BenchmarkBase[TInput](context) {
+abstract class AlgorithmBenchmarkBase[TInput](
+    context: BenchmarkContext,
```
mn-mikke: I like this indentation

jakubhava:
Agree it's a big change, however with this plugin it's not possible to do this iteratively. On the other hand it's still just formatting.

I'm not pushing for this change that much, but I promised several times that we will give try to scala formatter :))

benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/BenchmarkBase.scala
```diff
@@ -80,23 +79,28 @@ abstract class BenchmarkBase[TInput](context: BenchmarkContext) {
...
-    context.datasetDetails.labelCol +: (1 until numberOfColumns).map(i => "feature_" + i)
+    context.datasetDetails.labelCol +: (1 until numberOfColumns).map(i =>
+      "feature_" + i)
   }
 
   private def loadVirtualDataFrame(): DataFrame = {
     val columns = generateVirtualColumns()
-    val minValue: Long = context.datasetDetails.minValue.getOrElse[Int](Int.MinValue)
-    val maxValue: Long = context.datasetDetails.maxValue.getOrElse[Int](Int.MaxValue)
+    val minValue: Long =
```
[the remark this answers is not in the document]
jakubhava: 👍

benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/BenchmarkBase.scala
jakubhava: 👍

benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/BenchmarkBase.scala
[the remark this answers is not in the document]
jakubhava: changed to 160

benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/Runner.scala
```diff
@@ -121,10 +121,10 @@ object Runner {
   }
 
   private def filterCollection[T](
-                                   entity: String,
-                                   filter: Option[String],
-                                   collection: Seq[T],
-                                   nameGetter: T => String): Seq[T] = filter match {
+      entity: String,
```
jakubhava: It will take me some time to use to it, but I'm fine with it as long it's typical for Scala and used with Spark. I like consistency :)

benchmarks/src/main/scala/ai/h2o/sparkling/benchmarks/BenchmarkBase.scala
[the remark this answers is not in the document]
mn-mikke: Perfect sounds good to me, in Spark they use 100, but it seems to me too little.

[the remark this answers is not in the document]
jakubhava: Perfect1

jakubhava:
@mn-mikke I will wait for tests to turn green just to be sure. Is it ok to merge this PR from your perspective?

jakubhava:
Oki, tests are green ( unrelated failure in R tests -> the timeouts again. I have improved the speed of tests by reducing some sparklyr timeouts but I guess it was too much)

core/src/main/scala/ai/h2o/sparkling/backend/H2OFrameRelation.scala
```diff
@@ -26,13 +26,16 @@ import org.apache.spark.sql.types.{Metadata, MetadataBuilder, StructField, Struc
 import org.apache.spark.sql.{Row, SQLContext}
 
 /**
- * H2O relation implementing column filter operation.
- */
+  * H2O relation implementing column filter operation.
+  */
 case class H2OFrameRelation(frame: H2OFrame, copyMetadata: Boolean)(@transient val sqlContext: SQLContext)
-  extends BaseRelation with TableScan with PrunedScan {
```
mn-mikke: If I'm the original indentation is used in Spark as well (two spaces)

jakubhava: It I'm not wrong scalafmt has 2 spaces by default, do you prefer different value?

mn-mikke: Example: https://github.com/apache/spark/blob/8643e5d9c50294f59b01988d99d447a38776178e/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/collectionOperations.scala#L46

jakubhava: Ah, we have 4 spaces! Will fix

jakubhava: Is it this rule? https://scalameta.org/scalafmt/docs/configuration.html#continuationindentdefnsite

mn-mikke: 4 spaces are used only for class and method parameters.

jakubhava: I'm lost a little bit. So what should I do? :D You prefer 2 spaces everywhere?

mn-mikke: Example:
https://github.com/apache/spark/blob/6704103499d2003b1879ff0b4b8e29141e401b9f/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/stringExpressions.scala#L2271
but it's up to us. We can select the way which is more convenient for us.

jakubhava: Yup, switched to 2, it makes it easier to read. Please have additional look :)

PR #2003, opened 2020-03-29
[SW-2092] Switch H2OFrameSupport methods to use rest api

jakubhava:
JIRA for removal in 3.32 is here https://0xdata.atlassian.net/browse/SW-2134

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
@@ -78,7 +78,7 @@ class H2OFrame private (
   }
 
   def split(splitRatios: Double*): Array[H2OFrame] = {
-    if (splitRatios.sum > 1.0) {
+    if (splitRatios.sum >= 1.0) {
       throw new IllegalArgumentException("Split ratios must be lower than 1.0")
     }
     val endpoint = getClusterEndpoint(conf)
--- a/core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
@@ -51,33 +51,48 @@ class H2OFrame private (
   }
 
   def convertColumnsToCategorical(columns: Array[String]): H2OFrame = {
-    val endpoint = getClusterEndpoint(conf)
     val indices = columnNames.zipWithIndex.toMap
     val selectedIndices = columns.map { name =>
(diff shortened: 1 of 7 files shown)


core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
```diff
@@ -47,37 +47,51 @@ class H2OFrame private(val frameId: String, val columns: Array[H2OColumn], priva
...
       )
       val rapidsFrameV3 = update[RapidsFrameV3](endpoint, "99/Rapids", conf, params)
       H2OFrame(rapidsFrameV3.key.name)
     }
   }
 
-  def splitToTrainAndValidationFrames(splitRatio: Double): Array[H2OFrame] = {
-    if (splitRatio >= 1.0) {
-      throw new IllegalArgumentException("Split ratio must be lower than 1.0")
+  def split(splitRatios: Array[Double]): Array[H2OFrame] = {
```
jakubhava: I think it is better to keep the name just to `split`. We can do this change without any deprecation as this API has not been announced so far

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OClientUtils.scala
```diff
@@ -140,4 +141,23 @@ object H2OClientUtils extends SharedBackendUtils {
...
+  /**
+   * This method should be used whenever the Frame needs to be updated. This method ensures to use proper
+   * locking mechanism.
+   *
+   * @param fr frame to update
+   * @param f  function to run on the frame
+   * @tparam T H2O Frame Type
+   * @return returns the updated frame
+   */
+  def withLockAndUpdate[T <: Frame](fr: T)(f: T => Any): T = {
```
jakubhava: Keep the few specific client related methods in this object so they are not scattered around the code

core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
```diff
@@ -47,37 +47,52 @@ class H2OFrame private(val frameId: String, val columns: Array[H2OColumn], priva
...
```
mn-mikke: nit: Isn't better have just one method with multiple number of parameter?

jakubhava: yup, will do

PR #2004, opened 2020-03-30
[SW-2087] Deprecate MetricsSupport and make it possible to obtain training metrics on H2OMojoModel

jakubhava:
JIRA for removal: https://0xdata.atlassian.net/browse/SW-2088


py/src/ai/h2o/sparkling/ml/params/H2OTypeConverters.py
```diff
@@ -300,3 +301,17 @@ def scalaMapToDictStringToArrayString(value):
...
+            raise TypeError("None is not allowed.")
+        elif isinstance(value, JavaObject):
+            it = value.toIterator()
+            map = {}
+            while it.hasNext():
+                pair = it.next()
+                map[pair._1()] = pair._2()
+            return map
+        else:
+            raise TypeError("Invalid type.")
```
mn-mikke: nit: missing new line

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -105,25 +137,90 @@ trait H2OMOJOModelUtils {
...
+      json.remove("domains")
+      json.remove("help")
       val gson = new GsonBuilder().setPrettyPrinting().create
-      val prettyJson = gson.toJson(modelOutputJson)
+      val prettyJson = gson.toJson(json)
       prettyJson
     }
   }
 
+  protected def extractMetrics(json: JsonObject, metricType: String): Map[String, Double] = {
```
mn-mikke: Has anything changed in the methods that were moved from `H2OModel`?

jakubhava: Nope, complete copy

jakubhava:
> Shouldn't the changes be also reflected in the documentation?

Good catch, will update the Mojo doc

py/src/ai/h2o/sparkling/ml/params/H2OTypeConverters.py
jakubhava: 👍

PR #2005, opened 2020-03-30
[SW-2135] Deprecate JoinSupport in water.support package and make it part of ai.h2o.sparkling.H2OFrame

jakubhava:
JIRA for removal: https://0xdata.atlassian.net/browse/SW-2136

The goal of this change is just to move the current API so it uses Rest API. This api could use way more work such as exposing column names for joining etc, but I would like to focus on the porting part only in this change


core/src/bench/scala/ai/h2o/sparkling/bench/DataFrameConverterBenchSuite.scala
```diff
@@ -140,4 +141,10 @@ class DataFrameConverterBenchSuite extends BenchSuite with SharedH2OTestContext
     val hf = hc.asH2OFrame(df)
     hf.remove()
   }
+
+  private def sparseVector(len: Int, elements: Int, rng: Random = Random): org.apache.spark.ml.linalg.SparseVector = {
+    assert(elements < len)
+    val data = (1 to elements).map(_ => rng.nextInt(len)).sortBy(identity).distinct.map(it => (it, rng.nextDouble()))
+    Vectors.sparse(len, data).toSparse
```
mn-mikke: nit: Isn't `.toSparse` extra here?

core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
```diff
@@ -106,6 +108,85 @@ class H2OFrame private (
...
+    join(another, allFromCurrent = false, allFromAnother = false, "radix")
+
+  /**
+    * Outer join this frame with another frame
+    *
+    * @param another right frame
+    * @return new frame
+    */
+  def outerJoin(another: H2OFrame): H2OFrame = {
+    // Outer join is broken in H2O, simulate H2O's join via Spark for now
```
mn-mikke: Is there any Jira on H2O-3 side for this?

doc/src/site/sphinx/migration_guide.rst
```diff
@@ -47,6 +47,19 @@ From 3.30 to 3.32
...
+  are removed together with their encapsulating class. The enum ``water.support.munging.JoinMethod`` is also removed.
+  In order to perform joins, please use the following methods:
+
+  - Inner join: ``ai.h2o.sparkling.H2OFrame(idOfLeftFrame).innerJoin(rightFrame, method)``
+  - Outer join: ``ai.h2o.sparkling.H2OFrame(idOfLeftFrame).outerJoin(rightFrame, method)``
+  - Left join: ``ai.h2o.sparkling.H2OFrame(idOfLeftFrame).leftJoin(rightFrame, method)``
+  - Right join: ``ai.h2o.sparkling.H2OFrame(idOfLeftFrame).rightJoin(rightFrame, method)``
+
+
+The ``JoinMethod`` enum is no longer needed, but string is expected instead in ``method`` argument.
```
mn-mikke: nit: `method` argument -> the `method` argument

core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
jakubhava: I want to first discuss with someone from H2O before creating one.

mn-mikke: ok

jakubhava:
> Are these methods document anywhere? Are these methods also exposed for Python and R?
 answer to all is no

Would be good to expose & document later, but at this stage I just want to switch current functionality. WDYT?

core/src/bench/scala/ai/h2o/sparkling/bench/DataFrameConverterBenchSuite.scala
jakubhava: Just changed, interestingly not :( The method returns SparseVector, but in the signature it specifies just `Vector`

core/src/main/scala/ai/h2o/sparkling/H2OFrame.scala
jakubhava: Discussing with @wendycwong

doc/src/site/sphinx/migration_guide.rst
jakubhava: 👍

PR #2006, opened 2020-03-30
[SW-1779][POC] Use Kryo serialization in CategoricalPreviewWriter.

mn-mikke:


mn-mikke:
cc: @Pscheidl just if you're curious

PR #2010, opened 2020-03-31
[SW-2143] bin/build-kubernetes-images.sh should get spark version from $SPARK_HOME directory

mn-mikke:
Since the current logic gets the version from gradle.properties. The Spark version doesn’t have to match the current environment.


bin/build-kubernetes-images.sh
```diff
@@ -11,6 +11,7 @@ fi
 TOPDIR=$(cd "$(dirname "$0")/.." || exit; pwd)
 
 source "$TOPDIR/bin/sparkling-env.sh"
+export SPARK_VERSION=$(grep Spark  "$SPARK_HOME/RELEASE" | cut -d' ' -f2)
```
jakubhava, line 14: We already have SPARK_VERSION exposed in https://github.com/h2oai/sparkling-water/blob/b67d34a2ba5391cf7eb25397fb84c15f2239a809/bin/sparkling-env.sh#L146 but uses value from gradle.properties.

I guess the problem was with minor version?

We have this line to export the installed SPARK_VERSION https://github.com/h2oai/sparkling-water/blob/b67d34a2ba5391cf7eb25397fb84c15f2239a809/bin/sparkling-env.sh#L126

Maybe we can expose it in sparkling-env as `INSTALLED_SPARK_VERSION` ? WDYT?

bin/build-kubernetes-images.sh
```diff
@@ -24,27 +25,27 @@ if [ "$1" = "scala" ]; then
...
-  docker build -t "sparkling-water-r:$VERSION" -f "$WORKDIR/Dockerfile-R" "$WORKDIR"
+  docker build --build-arg spark_version=$SPARK_VERSION -t "sparkling-water-r:$VERSION" -f "$WORKDIR/Dockerfile-R" "$WORKDIR"
   echo "Done!"
 fi
 
 echo "Cleaning up temporary directories"
 rm -rf "$WORKDIR"
 
-echo "All done! You can find your images by running: docker images"
+echo "All done! You can find your images by running: docker images"
```
jakubhava, line 51: nit: missing nl

kubernetes/build.gradle
```diff
@@ -38,9 +40,13 @@ task createRDockerfile(type: Dockerfile) {
 
 task createPythonDockerfile(type: Dockerfile) {
   destFile = outputFilePython
-  from "spark-py:${sparkVersion}"
+  arg "spark_version"
+  from 'spark-py:$spark_version'
+  arg "spark_version"
   copyFile("h2o_pysparkling_${sparkMajorVersion}-${version}.zip", "/opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip")
-  runCommand "pip install /opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip"
+  runCommand 'echo "pyspark==$spark_version" > constraints'
```
jakubhava, line 47: We should run all the last 3 `runCommand` inside one `runCommand`. This way docker greates 3 layers, one with contastrains file, one with the packages and one with the removed constrains file -> bigger size.

If we do all the changes in one scope, docker won't save the intermediate files

mn-mikke: ok, thanks!

bin/build-kubernetes-images.sh
mn-mikke: ok, `installed_spark_version` sounds good to me. Will try to use.

yes, the problem i see is the minor version. IMHO, a user should be able create image of any spark version.

jakubhava: Yup, absolutely agree

kubernetes/build.gradle
```diff
@@ -38,9 +40,13 @@ task createRDockerfile(type: Dockerfile) {
...
+  runCommand "pip install /opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip -c constraints"
```
mn-mikke, line 48: this line instlalls all depencies including pyspark. Shouldn't pyspark be omitted if it's already in the image?

jakubhava: Yup, great catch! 

I would just verify after this change that the image works as expected -> starting pysparkling works

bin/build-kubernetes-images.sh
mn-mikke: The installed version doesn't contain spark fix version part, just `major.minor`. Should we expect that spark-submit is on class_path or take it from `$SPARK_HOME`

jakubhava: The sparkling-env script has method `checkSparkHome` which finds SPARK_HOME, we should follow that logic.

We can create new properties such as INSTALLED_SPARK_MAJOR_MINOR and full version as well

mn-mikke: ok

PR #2011, opened 2020-04-01
[SW-2145] Fix initialization of Sparkling Water JAR in case we call import pysparkling and sc is not yet running

jakubhava:


py/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -79,11 +84,11 @@ def __add_sparkling_jar_to_spark(sc):
         sc._jsc.addJar(sw_jar_file)
 
     @staticmethod
-    def __extracted_jar_path(sc):
+    def __extracted_jar_path():
 
         if Initializer.__extracted_jar_dir is None:
             zip_file = Initializer.__get_pysparkling_zip_path()
-            Initializer.__extracted_jar_dir = sc._temp_dir
+            Initializer.__extracted_jar_dir = tempfile.mkdtemp()
```
mn-mikke, line 91: Is this temp dir deleted somewhere?

jakubhava: Good catch, nope. But I can actually revert back and use the spark temp dir

py/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -43,7 +44,11 @@ class Initializer(object):
     @staticmethod
     def load_sparkling_jar():
         sc = SparkContext._active_spark_context
-        if Initializer.__sparklingWaterJarLoaded is not True and sc is not None:
+        if Initializer.__sparklingWaterJarLoaded is False and sc is None:
```
mn-mikke, line 47: nit: Maybe split the conditions into two `if` statements:
```
if Initializer.__sparklingWaterJarLoaded is False:
  if sc is None:
    ...
  else:
    ...
```

py/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -43,7 +44,11 @@ class Initializer(object):
...
+            # Ensure that when we do import pysparkling, spark will put later the JAR file
+            # to the driver. This option has effect only when SparkContext has not been started before.
+            os.environ["PYSPARK_SUBMIT_ARGS"] = "--jars " + Initializer.__get_sw_jar() + " pyspark-shell"
```
mn-mikke, line 50: qq: Couldn't the original `os.environ["PYSPARK_SUBMIT_ARGS"]` contain other args `pyspark-shell`. Couldn't we override them with this statement?

jakubhava: It might -> I can try to make it more robust and check for ``--jars`` presence

py/src/ai/h2o/sparkling/Initializer.py
jakubhava: 👍

py/src/ai/h2o/sparkling/Initializer.py
jakubhava: I will use the spark temp dir in the original use-case but will create new temp dir only in this specific use-case and mark it for deletion when the app ends. Just to make sure we do not affect existing users

jakubhava:
Should be more robust now

jakubhava:
@mn-mikke can you please double check?

py/examples/tests/pySparklingImportBeforeSparkSessionCreated.py
```diff
@@ -0,0 +1,13 @@
+from pysparkling import *
+from pyspark.sql import SparkSession
+
+# Initiate SparkSession
+spark = SparkSession.builder.appName("App name").getOrCreate()
+
+# Check if Sparkling Water classes are available
+
```
mn-mikke, line 8: nit: extra lines

py/tests/unit/simple/test_pyspark_submit_args.py
```diff
@@ -0,0 +1,44 @@
...
+    os.environ["PYSPARK_SUBMIT_ARGS"] = "--conf spark.app.name=test"
+    jar = Initializer._Initializer__get_sw_jar(None)
+    Initializer._Initializer__setUpPySparkSubmitArgs()
+    propEdited = os.environ["PYSPARK_SUBMIT_ARGS"]
+    del os.environ["PYSPARK_SUBMIT_ARGS"]
+    assert propEdited == "--jars {} --conf spark.app.name=test".format(jar)
+
+
+def testSparkSubmitOptsWithJars():
+    os.environ["PYSPARK_SUBMIT_ARGS"] = "--conf spark.app.name=test --jars     dummy.jar"
```
mn-mikke, line 39: nit: too many spaces `--jars     dummy.jar`?

py/tests/unit/simple/test_pyspark_submit_args.py
```diff
@@ -0,0 +1,44 @@
...
+    jar = Initializer._Initializer__get_sw_jar(None)
+    Initializer._Initializer__setUpPySparkSubmitArgs()
+    propEdited = os.environ["PYSPARK_SUBMIT_ARGS"]
+    del os.environ["PYSPARK_SUBMIT_ARGS"]
+    assert propEdited == "--conf spark.app.name=test --jars     {},dummy.jar".format(jar)
```
mn-mikke, line 44: nit: too many spaces `--jars     {},dummy.jar`?

py/tests/unit/simple/test_pyspark_submit_args.py
jakubhava: Yup, this was on purpose. I wanted to test also the case that the method does not depend on single space between jars and its argument. I can make it 2 spaces instead, that would test the same case

jakubhava: WDYT?

mn-mikke: ok, taking it back, thanks!

py/examples/tests/pySparklingImportBeforeSparkSessionCreated.py
jakubhava: thanks!

py/tests/unit/simple/test_pyspark_submit_args.py
jakubhava: Cool!

PR #2018, opened 2020-04-03
[SW-2154] Gradle Should Zip Symlinks and Don't Duplicate Content

mn-mikke:


mn-mikke:
@jakubhava It seems that I will have to close this PR since it seems that Steam is not able to read symlinks.

jakubhava:
Oki, thank you.

I was actually thinking about other solution instead of doing hacky simlinks on sparkling water side. 

Would it be possible to provide a small utility which could old steam somehow use so people are allowed to use newer sparkling water versions ? What do you think @bilcus @mn-mikke

PR #2019, opened 2020-04-03
[SW-2120][FOLLOWUP] Refresh Tokens When Using External Backend

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
```diff
@@ -131,6 +131,7 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Shell
       .addIf("-hivePrincipal", conf.hivePrincipal, conf.isHiveSupportEnabled)
       .addIf("-hiveJdbcUrlPattern", conf.hiveJdbcUrlPattern, conf.isHiveSupportEnabled)
       .addIf("-hiveToken", conf.hiveToken, conf.isHiveSupportEnabled)
+      .add("-refreshTokens", conf.isHiveSupportEnabled)
```
jakubhava, line 134: Should this be added also in internal mode?

honzasterba: in internal mode SW calls token refresh directly, so the answer is no

jakubhava: @honzasterba is it done inside of this method? https://github.com/h2oai/sparkling-water/blob/c8df8de31d3d41db4c8a2ca697a6653704fd768f/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala#L185

PR #2025, opened 2020-04-07
[SW-1971] Remove deprecated method H2OContext.getOrCreate with sparkSession or SparkContext argument

jakubhava:


py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -16,19 +16,19 @@
...
+from ai.h2o.sparkling.Initializer import Initializer
 from h2o.frame import H2OFrame
 from h2o.utils.typechecks import assert_is_type, Enum
 from pyspark.ml.util import _jvm
 from pyspark.rdd import RDD
+from pyspark.sql import SparkSession
 from pyspark.sql.dataframe import DataFrame
 from pyspark.sql.types import *
-from pyspark.sql import SparkSession
 from pyspark.sql.types import StringType, BooleanType, IntegerType, LongType, FloatType
```
mn-mikke, line 30: This line seems to be duplicated to https://github.com/h2oai/sparkling-water/pull/2025/files#diff-9143aa3a0fb6ae5a60db472c5edacf9dR29

jakubhava: Good catch!

PR #2026, opened 2020-04-07
[SW-2074] Remove Implicit Switch to External Backend when H2OConf.setH2OCluster is called

jakubhava:


jakubhava:
Python & R call Scala methods, so this was implemented only here

mn-mikke:
What about deprecation warnings in Python and R?

jakubhava:
Ah, I see, you've added warnings to python and r as well, will remove, thx!

PR #2036, opened 2020-04-10
[SW-2164] Fix class not found org.spark_project.jetty.util.thread.ThreadPool error

jakubhava:


core/src/main/scala/ai/h2o/sparkling/backend/utils/ProxyStarter.scala
```diff
@@ -23,11 +23,11 @@ import ai.h2o.sparkling.utils.SparkSessionUtils
 import org.apache.spark.SparkEnv
 import org.apache.spark.expose.Logging
 import org.apache.spark.h2o.H2OConf
-import org.spark_project.jetty.client.HttpClient
-import org.spark_project.jetty.proxy.ProxyServlet.Transparent
-import org.spark_project.jetty.server.{HttpConnectionFactory, Server, ServerConnector}
-import org.spark_project.jetty.servlet.{ServletContextHandler, ServletHandler}
-import org.spark_project.jetty.util.thread.{QueuedThreadPool, ScheduledExecutorScheduler, Scheduler}
+import org.eclipse.jetty.client.HttpClient
```
jakubhava, line 26: use "our" jetty which we shadow in the resulting assembly jar

PR #2040, opened 2020-04-16
[SW-2172] Shadow scala-compat introduced transitively from xgboost 1.0

jakubhava:


mn-mikke:
@jakubhava `SparkContext` got created with the current version

jakubhava:
Thank you!

PR #2045, opened 2020-04-17
[SW-2175] getFeaturesCols in python returns scala obj

jakubhava:


jakubhava:
Haven't checked, just fixed this issue. I will check when have a sec

jakubhava:
I think that is should be good now

PR #2046, opened 2020-04-20
[SW-2177] Context Path is Erased From Rest Calls

mn-mikke:
Calls like:
```
scala> val uri = new URI("http://127.0.0.1/mypath/")
scala> val url = uri.resolve("/3/CloudLock").toURL
url: java.net.URL = http://127.0.0.1/3/CloudLock
```
or
```
scala> val uri = new URI("http://127.0.0.1/mypath")
scala> val url = uri.resolve("3/CloudLock").toURL
url: java.net.URL = http://127.0.0.1/3/CloudLock
```
cause unwanted behavior.

The proposed solution tries to enforce the below case under all circumstances:
```
scala> val uri = new URI("http://127.0.0.1/mypath/")
scala> val url = uri.resolve("3/CloudLock").toURL
url: java.net.URL = http://127.0.0.1/mypath/3/CloudLock
```


mn-mikke:
cc @bilcus

jakubhava:
Great catch! One of the errors we saw on Steam?

bilcus:
> Great catch! One of the errors we saw on Steam?

Yup, hit that when I was trying external backend on 3.30.

PR #2047, opened 2020-04-20
[SW-2176] Expose spark.ext.h2o.allow_insecure_xgboost parameter

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/backend/utils/SharedBackendUtils.scala
```diff
@@ -153,6 +153,7 @@ trait SharedBackendUtils extends Logging with Serializable {
       .add("-user_name", conf.userName)
       .add("-login_conf", getDistributedFilePath(conf.loginConf))
       .add("-internal_security_conf", getDistributedFilePath(conf.sslConf))
+      .add("-allow_insecure_xgboost", conf.isInsecureXGBoostAllowed)
```
honzasterba, line 156: this is a boolean flag it does not have an argument, its just there or it is not

jakubhava: @honzasterba this is ok. The method `add` adds `-allow_insecure_xgboost` only if `conf.isInsecureXGBoostAllowed`. The second argument is condition when or when not to add the first arg

jakubhava:
```
  def add(arg: String, condition: Boolean): ArgumentBuilder = {
    if (condition) {
      add(arg)
    }
    this
  }
```

honzasterba: 👍

PR #2048, opened 2020-04-21
[SW-2174] Don't do version check in case user is using databricks-connect

jakubhava:
In this case the pyspark version is actually version of databricks runtime.

According to DBC docs pyspark needs to be uninstalled and databricks-connect library installed ( which installs pyspark with the incompatible version) -> so we should not see any collisions for existing users

### DIFF
--- a/py/src/ai/h2o/sparkling/Initializer.py
+++ b/py/src/ai/h2o/sparkling/Initializer.py
@@ -193,6 +193,16 @@ class Initializer(object):
             loader = loader.getParent()
 
     @staticmethod
+    def isRunningViaDBCConnect():
+        import pkg_resources as pkg
+        try:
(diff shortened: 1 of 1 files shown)


py/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -192,6 +192,16 @@ def __add_url_to_classloader(gateway, url):
...
+    @staticmethod
+    def isRunningViaDBCConnect():
+        import pkg_resources as pkg
+        try:
+            pkg.get_distribution('databricks-connect')
+            return True
+        except:
+            return False
+
+
```
mn-mikke, line 204: nit: extra line

jakubhava: Thanks!, fixed

PR #2051, opened 2020-04-21
[SW-2090] Set log level via rest API

jakubhava:
Created this change to to also reflect the latest changes in log class on master


extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/LogLevelHandler.scala
```diff
@@ -0,0 +1,51 @@
...
+/**
+  * This servlet class handles GET and PUT requests for the path /3/LogLevel
+  */
+final class LogLevelHandler extends Handler {
+
+  def setLogLevel(version: Int, request: LogLevelV3): LogLevelV3 = {
+    new MRTask() {
+      override def setupLocal() {
+        if (!Log.LVLS.contains(request.log_level)) {
+          Log.warn(s"[$request.log_level] is not a supported log level.")
```
mn-mikke, line 35: nit: Would't be possible to mention in the message that log level is remaining unchanged and mention the original log level?

py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -148,6 +148,9 @@ def show(self):
     def getConf(self):
         return self._conf
 
+    def setH2OLogLevel(self, level):
```
mn-mikke, line 151: Shouldn't we also expose a getter?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/LogLevelHandler.scala
jakubhava: 👍

py/src/ai/h2o/sparkling/H2OContext.py
jakubhava: Yup, good idea

PR #2052, opened 2020-04-22
[SW-2181] Context Path Must be Also Considered on H2O Worker Nodes

mn-mikke:
The last thing that is needed for SW to run on Steam. Manually tested and it works.


mn-mikke:
cc @bilcus

PR #2054, opened 2020-04-23
[SW-2179] Deprecate setH2OClientLogLevel and setH2ONodeLogLevel methods

jakubhava:


core/src/main/scala/org/apache/spark/h2o/H2OContext.scala
```diff
@@ -260,15 +261,11 @@ class H2OContext private (private val conf: H2OConf) extends H2OContextExtension
   /** Returns port where H2O REST API is exposed */
   def h2oLocalClientPort = leaderNode.port
 
-  /** Set log level for the client running in driver */
-  def setH2OClientLogLevel(level: String): Unit = {
-    if (H2OClientUtils.isH2OClientBased(conf)) {
-      water.util.Log.setLogLevel(level)
-    }
-  }
+  @DeprecatedMethod("setH2OLogLevel")
```
mn-mikke, line 264: Shouldn't we also specify the major version, in which the method will be removed?

PR #2058, opened 2020-04-23
[SW-2168] The AWS java sdk s3 in SW throws the exception: java.lang.IllegalStateException: Socket not created by this factory.

jakubhava:
We should shadow the AWS libraries used in SW


mn-mikke:
Why do we aws libraries in our environment? Can't we put them explicitly on class path when we're running sparkling water?

jakubhava:
Because of H2O persist S3 package, it requires specific verison of AWS libraries and these might colliide

PR #2059, opened 2020-04-24
[SW-XXX] [2090][Followup] Log structure has changed after changes on H2O side

jakubhava:


h2o-ops:
Hi @jakubhava, your PR title "[2090][Followup] Log structure has changed after changes on H2O side" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #2060, opened 2020-04-24
[SW-2091] Import Hive table over rest

jakubhava:


jakubhava:
I might be a bit lost, I thought I have updated the hive docs in this PR, am I missing something?

mn-mikke:
I'm just asking whether the section for kerberberized cluster shoul'd use SW API instead of H2O API.

jakubhava:
Ah, yes that is a good point. The links were pointing to the methods in Py/Scala so I was confused what doc did you mean. I will update the kerberos part

jakubhava:
@mn-mikke should be better now, please have a look

PR #2064, opened 2020-04-27
[SW-2160][FollowUp] More efficient stashing

jakubhava:


jakubhava:
This is just WIP for me to help me understand the issue with no space on jenkins master 
The reason why jenkins master is going so fast out of disk space are the stashes, but we actually don't need them.

PR #2099, opened 2020-05-13
[SW-2223] Fix Flow access when context path is used

jakubhava:


core/src/main/scala/ai/h2o/sparkling/backend/utils/ProxyStarter.scala
```diff
@@ -92,8 +92,8 @@ object ProxyStarter extends Logging {
     val holder = handler.addServletWithMapping(classOf[H2OFlowProxyServlet], "/*")
 
     val ipPort = RestApiUtils.getLeaderNode(conf).ipPort()
-    holder.setInitParameter("proxyTo", s"${conf.getScheme()}://$ipPort${conf.contextPath.getOrElse("")}")
-    holder.setInitParameter("prefix", conf.contextPath.getOrElse("/"))
+    holder.setInitParameter("proxyTo", s"${conf.getScheme()}://$ipPort")
```
jakubhava, line 95: this can be simplified to the variant on the right. Basically we proxy everything after the port

PR #2101, opened 2020-05-14
[SW-1569] Sparkling water fails to detect newer version of colorama

jakubhava:
We need to respect Spark python environment even for the python libraries checks

### DIFF
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -59,7 +59,11 @@ function getJarsArg() {
 
 
 function checkPythonPackages() {
-    packages=$(pip list --format=freeze)
+    PYTHON_BINARY=python
+    if [ -n "$PYSPARK_PYTHON" ]; then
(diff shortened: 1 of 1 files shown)


bin/sparkling-env.sh
```diff
@@ -59,7 +59,11 @@ function getJarsArg() {
 
 
 function checkPythonPackages() {
-    packages=$(pip list --format=freeze)
+    PYTHON_BINARY=python
+    if [ -n "$PYSPARK_PYTHON" ]; then
+      PYTHON_BINARY=$PYSPARK_PYTHON
+    fi
+    packages=$($PYTHON_BINARY -m pip list --format=freeze)
```
mn-mikke, line 66: Is this this going to work with Python 3 (commands: `python3` and `pip3`)?

jakubhava: It definitely does not break what is there right now. Default is still `python` and if PYSPARK_PYTHON points to python3 then all good as well

PR #2106, opened 2020-05-14
[SW-2229] Deprecate setClientIcedDir & setNodeIcedDir in favor of setIcedDir and clientIcedDir & nodeIcedDir in favor of icedDir

jakubhava:
In release 3.32, I would like to replace methods on H2OConf `setClientXYZ` and `setNodeXYZ` by single option `setXYZ` as another preparation to client removal.

The icedDir option wasn't also specified in case of external backend & this PR is fixing that as well.


core/src/main/scala/ai/h2o/sparkling/backend/utils/ArgumentBuilder.scala
```diff
@@ -28,7 +28,9 @@ class ArgumentBuilder() {
   }
 
   def add(args: Seq[String]): ArgumentBuilder = {
-    arguments = arguments ++ args
+    if (args.nonEmpty) {
```
mn-mikke, line 31: nit: Does this condition have any effect?

jakubhava: Yup -> https://github.com/h2oai/sparkling-water/blob/c5b2e075ef28f86e990fd24db0de45df81b75a8d/core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala#L110

jakubhava: Ah, I see. Good catch, actually no :D

arguments ++ Seq() is basically just arguments

jakubhava: will remove

PR #2110, opened 2020-05-15
[SW-2234] Remove System.exit hack used in tests 

jakubhava:
As our tests our running in client-less mode in Scala as well, we no longer need this hack as H2O never calls System.exit on Spark driver.

As the side-effect the time of our unit tests after removing this is lowered by around 20 minutes 🎉

### DIFF
--- a/core/src/test/scala/ai/h2o/sparkling/ConfigurationPropertiesTestSuite.scala
+++ b/core/src/test/scala/ai/h2o/sparkling/ConfigurationPropertiesTestSuite.scala
@@ -19,7 +19,6 @@ package ai.h2o.sparkling
 
 import java.net.{HttpURLConnection, URL}
 import java.nio.file.{Files, Path}
-import java.security.Permission
 
 import org.apache.spark.sql.SparkSession
 import org.junit.runner.RunWith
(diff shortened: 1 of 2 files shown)


core/src/test/scala/ai/h2o/sparkling/ConfigurationPropertiesTestSuite.scala
```diff
@@ -35,42 +34,6 @@ abstract class ConfigurationPropertiesTestSuite
...
-    // The method H2O.exit calls System.exit which confuses Gradle and marks the build
-    // as successful even though some tests failed.
-    // We can solve this by using security manager which forbids System.exit call.
-    // It is safe to use as all the methods closing H2O cloud and stopping operations have been
-    // already called and we just need to ensure that JVM with the client/driver doesn't call the System.exit method
-    try {
-      val securityManager = new NoExitCheckSecurityManager
-      System.setSecurityManager(securityManager)
-      if (hc != null) {
-        hc.stop()
```
mn-mikke, line 49: Don't we still need to call this line?

jakubhava: Nope :) It is called via normal execution path -> `stop` method during spark shutdown hook

PR #2112, opened 2020-05-15
[SW-2193] on Spark 2.4.5, build python docker image fails as apk is not found

jakubhava:
As we discussed, what do you think of supporting just latest Spark in our images for Kubernetes? 

Enterprise customers will probably have their own images

### DIFF
--- a/kubernetes/build.gradle
+++ b/kubernetes/build.gradle
@@ -45,9 +45,7 @@ task createPythonDockerfile(type: Dockerfile) {
   copyFile("h2o_pysparkling_${sparkMajorVersion}-${version}.zip", "/opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip")
   runCommand """\\
                 pip install /opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip && \\
-                apk add build-base python-dev && \\
-                pip install numpy && \\
-                apk del build-base python-dev
+                pip install numpy
                 """
 }
 


jakubhava:
@mn-mikke sorry for that many PRs today :) Can you please have a look on this last one? Thanks!

mn-mikke:
@jakubhava I will, I just want to verify the behavior of different Spark image versions locally.

kubernetes/build.gradle
```diff
@@ -45,9 +45,7 @@ task createPythonDockerfile(type: Dockerfile) {
   copyFile("h2o_pysparkling_${sparkMajorVersion}-${version}.zip", "/opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip")
   runCommand """\\
                 pip install /opt/spark/pyspark/python/lib/h2o_pysparkling_${sparkMajorVersion}-${version}.zip && \\
-                apk add build-base python-dev && \\
```
mn-mikke, line 48: Don't you want to add a condition that `apk add` && `apk del` will be called up to the version 2.4.4?

mn-mikke:
btw, docker R and Python docker files dosn't work for Spark 2.1-2.3 since Spark (`docker-image-tool.sh`) provides only one docker image called `spark`.

kubernetes/build.gradle
jakubhava: I was thinking about it and I think it makes sense to support images just for latest spark

jakubhava:
^ Good to know, but I think it's not that bug of a problem. We can improve in the future

jakubhava:
This will be fixed as part of the https://0xdata.atlassian.net/browse/SW-2235

PR #2115, opened 2020-05-15
[SW-2239] Use parallel build for H2O in the test pipeline

jakubhava:
As long as the pipeline is green, I think we are good to go


jakubhava:
We get speedup of around 3 minutes on building H2O

PR #2120, opened 2020-05-21
[SW-2250] Generate AutoML Scala Parameters from H2O-3 code

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
```diff
@@ -62,12 +57,12 @@ trait H2OCommonParams extends H2OMOJOAlgoSharedParams {
...
   def getSplitRatio(): Double = $(splitRatio)
 
-  def getSeed(): Long = $(seed)
-
-  def getNfolds(): Int = $(nfolds)
-
   def getColumnsToCategorical(): Array[String] = $(columnsToCategorical)
 
+  // To be overridden
+  def getSeed(): Long
```
jakubhava, line 63: qq: It seems like `getSeed` is already generated ( Removed from H2OAlgoCommonParams etc). Should this be removed?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -46,25 +46,18 @@ class H2OAutoML(override val uid: String)
   with H2OAutoMLParams
   with RestCommunication {
 
-  // Override default values
-  setDefault(nfolds, 5)
```
jakubhava, line 50: 🎉

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -75,26 +68,15 @@ class H2OAutoML(override val uid: String)
       Map()
     }
     val extra = if (algoParameters.nonEmpty) Map("algo_parameters" -> algoParameters) else Map()
-    Map("include_algos" -> determineIncludedAlgos(), "exclude_algos" -> null) ++ extra
+
+    val essentialParameters = getH2OAutoMLBuildModelsParams() - ("include_algos", "exclude_algos")
```
jakubhava, line 72: NIT: Maybe a small comment here for future readers?

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/Parameter.scala
```diff
@@ -17,6 +17,6 @@
 
 package ai.h2o.sparkling.api.generation.common
 
-case class Parameter(swName: String, h2oName: String, defaultValue: AnyRef, dataType: DataType, comment: String)
+case class Parameter(swName: String, h2oName: String, defaultValue: AnyRef, dataType: Class[_], comment: String)
 
 case class DataType(name: String, fullName: String, isEnum: Boolean)
```
jakubhava, line 22: Seems like `DataType` class is unused now

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/scala/ParametersTemplate.scala
```diff
@@ -59,7 +58,10 @@ object ParametersTemplate extends ScalaEntityTemplate with ParameterResolver {
          |${generateSetters(parameters)}
          |
          |  override private[sparkling] def getH2OAlgorithmParams(): Map[String, Any] = {
-         |    super.getH2OAlgorithmParams() ++
+         |    super.getH2OAlgorithmParams() ++ get${parameterSubstitutionContext.entityName}()
+         |  }
```
jakubhava, line 62: Sorry, don't see why this code change :(

jakubhava:
btw: great work!

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
mn-mikke: Some tests expect this method on parent classes (e.g. https://github.com/h2oai/sparkling-water/blob/master/ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OSupervisedMOJOModelTestSuite.scala#L129). So i defined just prototype of getters and setters in the parent and an implementation is in a generated class.

Another solution could be to duplicate the test for each specific algorithm tested  like this.

I'm not sure which approach is better. I'm ok with both.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/Parameter.scala
mn-mikke: Spot on, thanks!

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
jakubhava: I see!

What about different approach and make sure tests depends on generated sources via Gradle?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
mn-mikke: np, will add

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/scala/ParametersTemplate.scala
mn-mikke: This, is just becuase of Automl. I need to call each parameter group separately in automl (e.g. https://github.com/h2oai/sparkling-water/pull/2120/files#diff-edb3333884f3e1b4c38327d84eab0583R79) and modify it a bit. The original solution could give me all parameters without information to which group the parameter belongs to.

jakubhava: Ah, I see! I was having hard time reading it on Github. Makes sense :)

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
mn-mikke: >What about different approach and make sure tests depends on generated sources via Gradle?

Can share more about this approach? Tests can utilize generated classes already, but are shared by multiple algorithms.

jakubhava: Sorry, my bad. Now I see. What about using reflections in tests?

mn-mikke: yes, that's possible. Will change it.

mn-mikke:
@jakubhava I've tried to address your comments, can you have a second look?

PR #2122, opened 2020-05-22
[SW-2255] Deprecate H2OConf and H2OContext in package org.apache.spark

jakubhava:
The migration guide already contains the information, I just forgot to the the deprecation as part of that change


jakubhava:
@mn-mikke failed on unrelated R intermittent tests ( I wish we could fix that, still no idea), otherwise green

core/src/main/scala/org/apache/spark/h2o/H2OConf.scala
```diff
@@ -27,12 +28,17 @@ import org.apache.spark.internal.Logging
   */
 class H2OConf(sparkConf: SparkConf) extends ai.h2o.sparkling.H2OConf(sparkConf) {
   this.set("spark.ext.h2o.rest.api.based.client", value = false)
+  logWarning(
+    "The class org.apache.spark.h2o.H2OConf is deprecated and will be removed in release 3.34." +
```
mn-mikke, line 32: nit: `the release 3.34.0.1-1` or `the version 3.34`

PR #2123, opened 2020-05-22
[SW-2256] Put generated code into gitignore

jakubhava:


jakubhava:
@mn-mikke failed on unrelated R intermittent tests ( I wish we could fix that, still no idea), otherwise green

mn-mikke:
On the other hand, what you think about ignoring all src-gen folder in the main `.gitignore`?

jakubhava:
You mean globally? In case other submodule has src-gen folder so it is ignored as well?

I would to manually per submodule

mn-mikke:
ok, no problem.

PR #2125, opened 2020-05-22
[SW-2257] Expose XGBoost Parameter colsample_bynode in Sparkling Water

mn-mikke:


mn-mikke:
you are right, thanks!

PR #2128, opened 2020-05-25
[SW-2254] Smoke tests on Kubernetes, Internal backend in k8s cluster mode

jakubhava:
As part of this PR, I'm fixing a bug as that is required for R in interactive mode to pass -> we weren't using proxy when connecting from R/Py client which was a bug. We always need to use a proxy to hide internals. For example on Kubernetes we don't even have access to ports outside the driver.

Next effort will go into tests of external backend of Kubernetes in cluster mode. After that I will start looking into support & test client mode


doc/src/site/sphinx/deployment/kubernetes.rst
```diff
@@ -1,82 +1,48 @@
 Running Sparkling Water in Kubernetes
 -------------------------------------
 
-Sparkling Water can be executed inside of the Kubernetes cluster. Please note that we currently
-support only internal backend of sparkling water with cluster deployment mode, which means
-that both executors and driver are running inside Kubernetes. Sparkling Water supports
-Kubernetes since Spark version 2.3.
+Sparkling Water can be executed inside the Kubernetes cluster. Only cluster deployment mode is supported at this
```
jakubhava, line 4: I will be updating docs as I continue writing tests

kubernetes/src/main/scala/ai/h2o/sparkling/InitTest.scala
```diff
@@ -0,0 +1,49 @@
...
+    val frame = H2OFrame(new URI("file://" + SparkFiles.get("prostate.csv")))
+    val sparkDF = hc.asSparkFrame(frame).withColumn("CAPSULE", $"CAPSULE" cast "string")
+    val Array(trainingDF, testingDF) = sparkDF.randomSplit(Array(0.8, 0.2))
+
+    import ai.h2o.sparkling.ml.algos.H2OXGBoost
+    val estimator = new H2OXGBoost().setLabelCol("CAPSULE")
+    val model = estimator.fit(trainingDF)
+
+    model.transform(testingDF).show(false)
+    spark.stop()
```
mn-mikke, line 47: Shouldn't `hc` be stopped here as well?

kubernetes/src/main/scala/ai/h2o/sparkling/InitTest.scala
```diff
@@ -0,0 +1,49 @@
...
+      "https://raw.githubusercontent.com/h2oai/sparkling-water/master/examples/smalldata/prostate/prostate.csv")
+    val frame = H2OFrame(new URI("file://" + SparkFiles.get("prostate.csv")))
+    val sparkDF = hc.asSparkFrame(frame).withColumn("CAPSULE", $"CAPSULE" cast "string")
+    val Array(trainingDF, testingDF) = sparkDF.randomSplit(Array(0.8, 0.2))
+
+    import ai.h2o.sparkling.ml.algos.H2OXGBoost
+    val estimator = new H2OXGBoost().setLabelCol("CAPSULE")
+    val model = estimator.fit(trainingDF)
+
+    model.transform(testingDF).show(false)
```
mn-mikke, line 46: Shouldn't we assert the result?

kubernetes/src/r/initTest.R
```diff
@@ -0,0 +1,35 @@
...
+config <- spark_config_kubernetes(master = master,
+                                 image = paste0(registryId, ".dkr.ecr.us-east-2.amazonaws.com/sw_kubernetes_repo/sparkling-water:r-", version),
+                                 account = "default",
+                                 driver ="driver-r",
+                                 executors = 3,
+                                 ports = c(8880, 8881, 4040, 54321))
+config["spark.home"] <-  sparkHome
+sc <- spark_connect(config = config, spark_home = sparkHome)
+hc <- H2OContext.getOrCreate()
+expect_equal(length(invoke(hc$jhc, "getH2ONodes")), 3)
```
mn-mikke, line 34: can we try to run `asH2OFrame`, `asSparkFrame` transformations here?

kubernetes/src/main/scala/ai/h2o/sparkling/InitTest.scala
jakubhava: No need, H2O is stopped automatically because it recognizes Spark shutdown hook

kubernetes/src/main/scala/ai/h2o/sparkling/InitTest.scala
jakubhava: The main test here was to test that the algo can finish (communication), I don't think it is crucial to test the that the algorithm provides correct result as we have unit tests for that. WDYT?

kubernetes/src/r/initTest.R
jakubhava: Yup, good idea

doc/src/site/sphinx/deployment/kubernetes.rst
```diff
@@ -1,117 +1,111 @@
...
-
-        .. tab-container:: Scala
-            :title: Scala
+2. Ensure that you have working Kubernetes Cluster and ``kubectl`` installed
 
-            .. code:: bash
+3. Ensure you have ``SPARK_HOME`` set up to home of your Spark distribution of version SUBST_SPARK_VERSION
 
-                ./bin/build-kubernetes-images.sh scala
+4. Run ``kubectl cluster-info`` to obtain Kubernetes master URL.
```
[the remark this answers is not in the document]
jakubhava: kubectl is standard kuberentes client, so I would say yes

core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
```diff
@@ -96,9 +96,19 @@ class H2OContext private[sparkling] (private val conf: H2OConf) extends H2OConte
...
+
   private val (flowIp, flowPort) = {
     val uri = ProxyStarter.startFlowProxy(conf)
-    (uri.getHost, uri.getPort)
+    // SparklyR implementation of Kubernetes connection works in a way that it does port-forwarding
+    // from the driver node to the node where interactive session is running. We also need to make
+    // sure that we provide valid ip. If this wouldn't be done, we would return dns record internal to
+    // kubernetes
+    if (conf.get("spark.ext.h2o.client.language") == "r"
+        && sparkContext.master.startsWith("k8s")) {
```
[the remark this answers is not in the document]
jakubhava: Yup, but it would break spotless formatting. I would keep it as it is

core/src/main/scala/ai/h2o/sparkling/H2OConf.scala
```diff
@@ -34,6 +34,9 @@ class H2OConf(val sparkConf: SparkConf)
   with ExternalBackendConf
   with Serializable {
 
+  if (!sparkConf.contains("spark.ext.h2o.client.language")) {
+    sparkConf.set("spark.ext.h2o.client.language", "scala")
```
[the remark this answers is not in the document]
jakubhava: Yup, new internal property to be able to distinguish on the backend who is the client.

Good point about the default value, will improve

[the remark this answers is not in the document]
jakubhava: Should be fixed

kubernetes/src/r/initTest.R
jakubhava: Done :)

core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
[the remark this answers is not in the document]
jakubhava: Created internal getter for the option and moved to one line

kubernetes/src/main/scala/ai/h2o/sparkling/InitTest.scala
mn-mikke: ok, we change `show(false)` to `foreach` to iterate over the whole dataset (to get over all partitions).

jakubhava: Good point!, I think collect will do, it's dataset with just 381 rows :)

jakubhava:
@mn-mikke both Kubernetes tests and regular PR tests are green. Can you please double check if all looks good or if you think any additional changes are needed?

Thanks a lot!

PR #2129, opened 2020-05-25
[SW-2258] Generate GridSearch Scala Parameters from H2O-3 code

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/ParameterNameConverter.scala
```diff
@@ -19,6 +19,8 @@ package ai.h2o.sparkling.api.generation.common
 
 object ParameterNameConverter {
   val h2oToSWExceptions: Map[String, String] = Map(
+    "response_column" -> "labelCol",
+    "weights_column" -> "weightCol",
```
jakubhava, line 23: In the conversion rule bellow for Columns, is is able to capture `fold_column` ? The rule has capital `C` in `Column`

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OGridSearchParams.scala
```diff
@@ -20,31 +20,23 @@ import java.util
 
 import ai.h2o.sparkling.ml.algos.{H2OGridSearch, H2OSupervisedAlgorithm}
 import ai.h2o.sparkling.ml.internals.H2OMetric
-import hex.grid.HyperSpaceSearchCriteria
-import hex.{Model, ScoreKeeper}
+import hex.Model
```
jakubhava, line 23: NIT: extra import?

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
```diff
@@ -21,10 +21,8 @@ import org.apache.spark.ml.param._
 /**
   * This trait contains parameters that are shared across all algorithms.
   */
-trait H2OCommonParams extends H2OMOJOAlgoSharedParams {
+trait H2OCommonParams extends H2OMOJOAlgoSharedParams with H2OAlgoParamsBase {
```
jakubhava, line 24: nit: I was just thinking about renaming `H2OMOJOAlgoSharedParams` to `H2OBaseMOJOParams`. Then it would be consistent with `H2OSupervisedMOJOParams`, `H2OTreeBasedSupervisedMOJOParams`. 

If we decide to do it, I would also rename `H2OTargetEncoderParams` to `H2OTargetEncoderMOJOParams` so it is all consistent.

All these classes are in scoring package in `ai.h2o.sparkling.params`

py/src/ai/h2o/sparkling/ml/params/H2OCommonSupervisedParams.py
```diff
@@ -23,17 +23,59 @@
 
 
 class H2OCommonSupervisedParams(H2OCommonParams, H2OSupervisedMOJOParams):
+    foldCol = Param(
```
jakubhava, line 26: The duplication of params with H2OCommonUnsupervisedParams.py will be removed after we start generating python if I understand correctly? At this stage, we just prepared H2OCommonParams.py so it will remain as it is.

mn-mikke: Yes, exactly. I just wanted to do a minimal change to make python tests to pass. When python code gets generated, there won't be nieghter `H2OCommonUnsupervisedParams` or `H2OCommonSupervisedParams`.

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OGridSearchParams.scala
mn-mikke: thanks!

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OCommonParams.scala
mn-mikke: NP, will do.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/ParameterNameConverter.scala
mn-mikke: The conversion rule is applied on a capitalized sparkling water name. Here is the conversion logic:
```
  def convertFromH2OToSW(parameterName: String): String = {
    val parts = parameterName.split("_")
    val capitalizedParts = parts.head +: parts.tail.map(_.capitalize)
    val regularValue = conversionRules.foldLeft(capitalizedParts.mkString) {
      case (input, (from, to)) => input.replace(from, to)
    }
    h2oToSWExceptions.getOrElse(parameterName, regularValue)
  }
```

py/src/ai/h2o/sparkling/ml/params/H2OCommonSupervisedParams.py
jakubhava: Thanks for explanation!

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/ParameterNameConverter.scala
jakubhava: Thanks!

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OGridSearchParams.scala
mn-mikke: In fact, the import is required by [the line 65](https://github.com/h2oai/sparkling-water/pull/2129/files/175f47c5ca617d0b60e9f29a0333a1667d3cc2a0#diff-7419246e33dc368bbd698a3f802b79c2R65)

jakubhava: You're right, sorry

jakubhava:
I think this is fine to merge, it failed on unrelated R tests. btw: I have made another attempt today to fix the intermittent R issue ( this PR is not on that change yet)

mn-mikke:
@jakubhava I've removed dependency on `H2OCommonParams`. Can you take a look once again before I merge it.

PR #2131, opened 2020-05-26
[SW-1608] GridSearch Should Reference Hyper Parameters with SW Names

jakubhava:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -202,16 +205,18 @@ class H2OGridSearch(override val uid: String)
   def getGridModelsParams(): DataFrame = {
     ensureGridSearchIsFitted()
     val hyperParamNames = getHyperParameters().keySet().asScala.toSeq
+    val h2oToSwParamMap = getAlgo().getSWtoH2OParamNameMap().map(_.swap)
     val rowValues = gridModels.zip(gridModels.map(_.uid)).map {
       case (model, id) =>
-        val outputParams = model.getTrainingParams().filter { case (key, _) => hyperParamNames.contains("_" + key) }
+        val outputParams =
+          model.getTrainingParams().filter { case (key, _) => hyperParamNames.contains(h2oToSwParamMap(key)) }
```
mn-mikke, line 212: Isn't this a breaking change? it seems that `hyperParamNames` contains SW names only.

jakubhava: Do you mean that in case of `outputParams` we always output SW names? That is a good catch, I will try to make sure that the param names remain consistent in the fix release

mn-mikke: In this statement `hyperParamNames.contains(h2oToSwParamMap(key))`, the `key` is always converted to sw name. What if `hyperParameNames` contains old h2o names? This comment is als related to [the line 218](hyperParamNames.contains(h2oToSwParamMap(key))).

jakubhava: Thanks, will fix

jakubhava: I decided to deprecated the behaviour in 3.30 and in 3.32 support just the new version. I think this solution will be cleaner than supporting two paths in the code

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -205,16 +192,28 @@ class H2OGridSearch(override val uid: String)
...
     val hyperParamNames = getHyperParameters().keySet().asScala.toSeq
+    val h2oToSwParamMap =
+      getAlgo().getSWtoH2OParamNameMap().map(_.swap)
     val rowValues = gridModels.zip(gridModels.map(_.uid)).map {
       case (model, id) =>
-        val outputParams = model.getTrainingParams().filter { case (key, _) => hyperParamNames.contains("_" + key) }
+        val outputParams =
+          model.getTrainingParams().filter {
+            case (key, _) =>
+              !IgnoredParameters.all.contains(key) && hyperParamNames.contains(h2oToSwParamMap(key))
```
jakubhava, line 202: Note: currently, we ignore parameters also in `IgnoredParamters.implementedInParent` such as `distribution` etc.

Parameters in this array are not gridable anyways, so not actual issue, but after merging https://github.com/h2oai/sparkling-water/pull/2129 only parameters in this array left are training and validation frame, so all good

ml/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
   api("ai.h2o:h2o-ext-target-encoder:${h2oVersion}")
   api("ai.h2o:mojo2-runtime-api:${mojoPipelineVersion}")
   api("ai.h2o:mojo2-runtime-impl:${mojoPipelineVersion}")
+  api(project(':sparkling-water-api-generation'))
```
jakubhava, line 29: @mn-mikke Since I'm using `IgnoredParameters` class in the code, I need to include the generation code into the assembly.

To be honest, I do not like this that much. The other approach I was thinking about was to copy the IgnoredParameters class to ml src-gen directory via gradle copy task so we do not need to have such dependency. WDYT?

mn-mikke: I don't see a big problem with that. The dependency was there already, but as compile-only [line 31](https://github.com/h2oai/sparkling-water/pull/2131/files#diff-0e661b343c1a53714aa5ee80b5db48a8L31). 

Btw, can you use `implementation` instead of `api`?

jakubhava: oki oki, thanks.

Will change

jakubhava: Done :)

jakubhava:
@mn-mikke I will let the test pass, in the meantime, can you please have a second look?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -205,16 +192,17 @@ class H2OGridSearch(override val uid: String)
   def getGridModelsParams(): DataFrame = {
     ensureGridSearchIsFitted()
     val hyperParamNames = getHyperParameters().keySet().asScala.toSeq
+    val h2oToSwParamMap =
```
mn-mikke, line 195: nit: Can be  on one line?

jakubhava: yup!

PR #2138, opened 2020-05-27
[SW-2271] Missing LogUtil class on external h2o backend

jakubhava:
User on gitter discovered this issue.

This PR goes directly into 3.30 as this error does not happen on 3.32 because there we set log level via rest API.

I have created JIRA to create test on master https://0xdata.atlassian.net/browse/SW-2272 for this as well. But since on master we have prepared methods to query log level via rest api & we don't use client the test can be much more simple there


extensions/src/main/scala/ai/h2o/sparkling/extensions/internals/LogUtil.scala
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package ai.h2o.sparkling.backend.utils
+package ai.h2o.sparkling.extensions.internals
```
jakubhava, line 18: This class needs to be available on external H2O since the MRTask bellow

PR #2139, opened 2020-05-27
[SW-2272] Add test for setH2ONodeLogLevel

jakubhava:


core/src/test/scala/ai/h2o/sparkling/H2OContextTestSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+@RunWith(classOf[JUnitRunner])
+class H2OContextTestSuite extends FunSuite with SharedH2OTestContext {
+
+  override def createSparkSession(): SparkSession = sparkSession("local[*]")
+
+  test("setH2OLogLevel") {
+    hc.setH2OLogLevel("DEBUG")
+    assert(hc.getH2OLogLevel() == "DEBUG")
+  }
+
```
mn-mikke, line 36: nit: extra line

mn-mikke:
Was there any troubles with setting log level?

jakubhava:
Yup, not on master but https://github.com/h2oai/sparkling-water/pull/2138. User on gitter was calling `setH2OLogLevel` in external backend but the LogUtil class was missing in the extensions. This is no longer case on master as we user rest api.

jakubhava:
But when fixing in on rel branch, I created a test on master as well to have this feature verified

core/src/test/scala/ai/h2o/sparkling/H2OContextTestSuite.scala
jakubhava: thank you!

PR #2142, opened 2020-05-28
[SW-2275] Ensure pytest cache is generated into build directory

jakubhava:
Doing some small improvements whilst testing kubernetes


jakubhava:
Just that the file gets generated to build directory where I think all temporary files should be generated. Otherwise no real impact

PR #2150, opened 2020-06-02
[SW-2265] Support Kmeans in Grid search

jakubhava:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -155,6 +155,7 @@ class H2OGridSearch(override val uid: String)
         case H2OModelCategory.Regression => H2OMetric.RMSE
         case H2OModelCategory.Binomial => H2OMetric.AUC
         case H2OModelCategory.Multinomial => H2OMetric.Logloss
+        case H2OModelCategory.Clustering => return gridModels // no sorting for unsupervised clustering
```
jakubhava, line 158: I'm not sure if we can sort the models in case of unsupervised algos, for now I'm not sorting them

jakubhava: Discussed with @MartinBarus  as well. I would keep it like this as we can't automatically sort the models

mn-mikke: @maurever Is there any performance metric for clustering algorithms in H2O-3?

maurever: There are several cluster metrics, I did not find h2o documentation for them, but some other [source](https://sites.google.com/site/dataminingatuoc/k-means) says:

**betweenss:** is the between clusters sum of squares. In fact it is the mean of distances between cluster centers. One expects, this ratio, to be as higher as possible, since we would like to have heterogenous clusters.

**withinss:** is the within cluster sum of squares. So it results in a vector with a number for each cluster. One expects, this ratio, to be as lower as possible for each cluster, since we would like to have homogeneity within the clusters.

**tot_withinss** = sum ( withinss )
**totss** =  tot_withinss + betweenss

jakubhava: Thank you @maurever

I see that performance method in H2O on kmeans returns:

```
Total Within Cluster Sum of Square Error: 12.3864300025
Total Sum of Square Error to Grand Mean: 546.909233123
Between Cluster Sum of Square Error: 534.522803121
```

maurever: Yes, that are the metrics I mentioned above. 

Total Within Cluster Sum of Square Error = tot_withinss 
Total Sum of Square Error to Grand Mean = totss
Between Cluster Sum of Square Error = betweenss

jakubhava: I will try to incorporate this to the code on SW side

maurever: :+1:

scoring/src/main/scala/ai/h2o/sparkling/ml/internals/H2OMetric.java
```diff
@@ -39,7 +39,11 @@
   MaxPerClassError(false),
   Accuracy(true),
   MSE(false),
-  RMSE(false);
+  RMSE(false),
+  Withinss(false),
+  Betweenss(true),
+  TotWithinss(false),
+  Totss(false);
```
jakubhava, line 46: The boolean here means `higherTheBetter`. @maurever can you please double check this make sense?

If I undertand correctly, `totss` is a bit more tricky and could have value `true` and `false` so I just picked one

maurever: I double-check and it looks correct. For `Totss` I found the less the better.

jakubhava: Thank you!

PR #2152, opened 2020-06-03
[SW-2284] Ensure that all requests to backend cluster go via leader node

jakubhava:


core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -144,37 +144,35 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
...
-    }
+  private def verifyWebOpen(conf: H2OConf): Unit = {
+    val endpoint = RestApiUtils.getClusterEndpoint(conf)
+    val verifyWebOpenV3 = RestApiUtils.query[VerifyWebOpenV3](endpoint, "/3/verifyWebOpen", conf)
+    val nodesWithoutWeb = verifyWebOpenV3.nodes_web_disabled
     if (nodesWithoutWeb.nonEmpty) {
-      throw new H2OClusterNotReachableException(
-        s"""
+      throw new H2OClusterNotReachableException(s"""
     The following worker nodes are not reachable, but belong to the cluster:
```
mn-mikke, line 153: Would't be possible to add information to the message that web endpoints of those nodes are disabled?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/VerificationHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+      nodesWithDisabledWeb.appendAll(mrt.nodesWithDisabledWeb)
+    }
+  }
+
+  class VerifyVersionMRTask(referencedVersion: String) extends MRTask[VerifyVersionMRTask] {
+    var nodesWithWrongVersion: util.HashMap[String, String] = new util.HashMap[String, String]()
+
+    override def setupLocal(): Unit = {
+      val currentVersion = H2O.ABV.projectVersion
+      if (referencedVersion != currentVersion && !referencedVersion.endsWith("-SNAPSHOT")) {
```
mn-mikke, line 63: it seems if `referencedVersion.endsWith("-SNAPSHOT")` holds true, we don't have execute MRTask and make REST call.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/schema/VerifyVersionV3.java
```diff
@@ -0,0 +1,34 @@
...
+import water.api.API;
+import water.api.schemas3.RequestSchemaV3;
+
+public class VerifyVersionV3 extends RequestSchemaV3<Iced, VerifyVersionV3> {
+
+  @API(help = "Reference version to validate against", direction = API.Direction.INPUT)
+  public String referenced_version;
+
+  @API(help = "Nodes with wrong versions", direction = API.Direction.OUTPUT)
+  public String[] nodes_wrong_version;
```
mn-mikke, line 30: is there any chance to serialize and deserialize `nodes_wrong_version` and `versions` as a map?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/VerificationHandler.scala
jakubhava: Good point!

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/schema/VerifyVersionV3.java
jakubhava: Originally I thought that H2O won't be able to serialize Map. I just verified in a local build that H2O can't serialize Map in V3 object. But to avoid passing the information separately I created a new V3 object storing the IP and version together.

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -113,10 +116,9 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
   protected def getAndVerifyWorkerNodes(conf: H2OConf): Array[NodeDesc] = {
     try {
       lockCloud(conf)
-      val nodes = RestApiUtils.getNodes(conf)
-      verifyWebOpen(nodes, conf)
+      verifyWebOpen(conf)
```
mn-mikke, line 119: Isn't this check necessary only with manually deployed external backend?

jakubhava: Usually it is. But I can image really rare case where user would pass `disable_web` to `extraOptions` which then gets propagated to h2o start command

mn-mikke: ok

PR #2159, opened 2020-06-03
[SW-2287] Generate Python Parameter Classes

mn-mikke:


py/src/ai/h2o/sparkling/ml/params/H2OAutoMLParams.py
```diff
@@ -17,253 +17,20 @@
...
-    def setExploitationRatio(self, value):
-        return self._set(exploitationRatio=value)
+class H2OAutoMLParams(
+    H2OCommonParams,
+    H2OAutoMLBuildControlParams,
+    H2OAutoMLBuildModelsParams,
+    H2OAutoMLInputParams,
+    H2OAutoMLStoppingCriteriaParams,
+    HasMonotoneConstraints
+):
```
jakubhava, line 35: nit: `):` could be on previous line

py/src/ai/h2o/sparkling/ml/params/H2OCommonParams.py
```diff
@@ -61,42 +42,18 @@ class H2OCommonParams(H2OBaseMOJOParams):
...
-    def setFoldCol(self, value):
-        return self._set(foldCol=value)
-
-    def setWeightCol(self, value):
-        return self._set(weightCol=value)
-
     def setSplitRatio(self, value):
         return self._set(splitRatio=value)
 
-    def setSeed(self, value):
```
jakubhava, line 94: Method `setWithContributionsCol` is missing across the code, but I guess that is because this branch is not on top of master?

py/src/ai/h2o/sparkling/ml/params/H2OAutoMLParams.py
mn-mikke: python seems to be ok with the proposal, but Idea starts complaining: `PEP8: continuation line with same indent as next logical line` If this warning is unimportant, I can do it. WDYT?

jakubhava: Ah, cool. Didn't know about PEP8. Let's keep it as it is :)

py/src/ai/h2o/sparkling/ml/params/H2OCommonParams.py
mn-mikke: I've rebased to master.

mn-mikke:
@jakubhava Yes, I will generate algorithm classes in another PR.

py/src/ai/h2o/sparkling/ml/params/H2OCommonParams.py
mn-mikke: (see [line 92](https://github.com/h2oai/sparkling-water/pull/2159/files#diff-0c9f45aaa9f977bf9890ad82d269e9f8R92)

PR #2161, opened 2020-06-04
[SW-2252] Publish kubernetes docker images to docker hub

jakubhava:
@mn-mikke Finally ready to review.

The images are published as part of nightlies and as part of release. I have tested the upload to my personal docker account in both cases ( nightly & release)


ci/sparklingWaterPipeline.groovy
```diff
@@ -76,7 +76,9 @@ def withSharedSetup(sparkMajorVersion, config, code) {
                 checkout scm
                 config.commons = load 'ci/commons.groovy'
                 config.put("sparkVersion", getSparkVersion(config))
-
+                def kubernetesBoundaryVersionLine = readFile("gradle.properties").split("\n").find() { line -> line.startsWith('kubernetesSupportSinceSpark') }
```
mn-mikke, line 79: Haven't you already implemented reading such properties via a conversion to a map?

ci/sparklingWaterPipeline.groovy
```diff
@@ -359,6 +362,9 @@ def publishNightly() {
                         def path = getS3Path(config)
                         sh """
                             sed -i 's/^version=.*\$/version=${version}/' gradle.properties
+                            sed -i 's/^h2oMajorName=.*\$/h2oMajorName=${getH2OBranchMajorName()}/' gradle.properties
+                            sed -i 's/^h2oMajorVersion=.*\$/h2oMajorName=${getH2OBranchMajorVersion()}/' gradle.properties
```
mn-mikke, line 366: ditto

ci/sparklingWaterPipeline.groovy
```diff
@@ -382,4 +388,55 @@ def publishNightly() {
...
+def publishSparklingWaterDockerImage(String type, version, sparkMajorVersion) {
+    sh """
+        H2O_HOME=${env.WORKSPACE}/h2o-3 ./bin/build-kubernetes-images.sh ${type}
+        docker tag sparkling-water-${type}:${version} h2oai/sparkling-water-${type}:latest-nightly-${sparkMajorVersion}
+        docker push h2oai/sparkling-water-${type}:latest-nightly-${sparkMajorVersion}
+        docker rmi h2oai/sparkling-water-${type}:latest-nightly-${sparkMajorVersion}
+    """
+}
+
+def publishNightlyDockerImages() {
```
mn-mikke, line 400: This method seems to be very similar to https://github.com/h2oai/sparkling-water/pull/2161/files#diff-19c268ee1cd25cf26d1e8c79d4053ea0R388. Would be possible to share some logic?

ci/sparklingWaterPipeline.groovy
jakubhava: I have implemented it when I wrote ci/Jenkinsfile-kubernetes from scratch. There is still JIRA for doing that for release in tests pipeline. I prefer to do that separately

ci/sparklingWaterPipeline.groovy
jakubhava: Good catch!

ci/sparklingWaterPipeline.groovy
jakubhava: Good catch, done! 

I have however kept the method `publishDockerImage` in each file without sharing the logic as when I wrote it, there were too many conditions and switches between release & nightly. In that case, I think it is clearer to keep it separate

PR #2162, opened 2020-06-04
[SW-2290] Enable to Run Benchmarks from Local PC by Passing VPC and Subnet

mn-mikke:


benchmarks/run_benchmarks.sh
```diff
@@ -40,14 +53,13 @@ if [ -z "$run_local_internal" ]; then
     run_local_internal="true"
 fi
 
+
+
```
jakubhava, line 57: Nit: extra lines

PR #2164, opened 2020-06-05
[SW-2291] Generate Python Algorithm Classes

mn-mikke:
Generation of class for GridSearch and AutoML will come in a different PR.

When a new algorithm is added `__init__.py` files still have to be updated manually.


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/Runner.scala
```diff
@@ -37,13 +37,26 @@ import water.automl.api.schemas3.AutoMLBuildSpecV99._
...
-    val monotonicity = ExplicitField("monotone_constraints", "HasMonotoneConstraints")
-    val userPoints = ExplicitField("user_points", "HasUserPoints")
+    val monotonicity =
+      ExplicitField("monotone_constraints", "HasMonotoneConstraints", new util.HashMap[String, Double]())
+    val userPoints = ExplicitField("user_points", "HasUserPoints", null)
     type DeepLearningParametersV3 = DeepLearningV3.DeepLearningParametersV3
 
     val explicitDefaultValues =
-      Map("max_w2" -> "java.lang.Float.MAX_VALUE", "response_column" -> """"label"""", "model_id" -> "null")
+      Map[String, Any]("max_w2" -> 3.402823e38f, "response_column" -> "label", "model_id" -> "null")
```
jakubhava, line 59: just qq: why we changed `java.lang.Float.MAX_VALUE` to `3.402823e38f`?

mn-mikke: I can't use tha value of `java.lang.Float.MAX_VALUE`  directly because of this bug in scala compiler https://github.com/scala/bug/issues/10148. For generating of Python code, I also need some  description of float value, which is language indepenendent so that's the reason why `"java.lang.Float.MAX_VALUE"` was replaced with a bit smaller value directly.

jakubhava: I see, thanks for the explanation!

PR #2165, opened 2020-06-08
[SW-2062] Verify that training frame has all columns constants using Spark instead of H2O error

jakubhava:
@mn-mikke promised this change a while ago. As you suggested, we can check the validity even before we start the training

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
@@ -55,8 +55,8 @@ trait H2OAlgoCommonUtils extends EstimatorCommonUtils {
     val featureColumns = getFeaturesCols().map(sanitize).map(col)
 
     if (dataset.select(featureColumns: _*).distinct().count() == 1) {
-      throw new IllegalArgumentException(s"H2O could not use any of the specified feature" +
-        s" columns: '${getFeaturesCols().mkString(", ")}' because they are all constants. H2O ignores constant columns.")
+      throw new IllegalArgumentException(s"H2O could not use any of the specified features" +
+        s" columns: '${getFeaturesCols().mkString(", ")}' because they are all constants. H2O requires at least one non-constant column.")
     }
     val excludedColumns = excludedCols.map(sanitize).map(col)
     val columns = featureColumns ++ excludedColumns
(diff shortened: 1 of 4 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
```diff
@@ -52,6 +53,11 @@ trait H2OAlgoCommonUtils extends EstimatorCommonUtils {
     }
 
     val featureColumns = getFeaturesCols().map(sanitize).map(col)
+
+    if (dataset.select(featureColumns: _*).distinct().count() == 1) {
```
mn-mikke, line 57: Isn't [this H2O option](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/algo-params/ignore_const_cols.html#ignore-const-cols) important for make the condition?

jakubhava: Good find! But I don't think so; we just need to check whether all columns are constants. If this option is applied, H2O will have no columns to use as features and we throw the same exception which I thing is correct. WDYT?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
```diff
@@ -52,6 +53,11 @@ trait H2OAlgoCommonUtils extends EstimatorCommonUtils {
...
+      throw new IllegalArgumentException(s"H2O could not use any of the specified feature" +
```
mn-mikke, line 58: nit `feature` ->  `features`

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
```diff
@@ -52,6 +53,11 @@ trait H2OAlgoCommonUtils extends EstimatorCommonUtils {
...
+        s" columns: '${getFeaturesCols().mkString(", ")}' because they are all constants. H2O ignores constant columns.")
```
mn-mikke, line 59: `H2O ignores constant columns.` -> `H2O requires at least one non-constant column.`

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
mn-mikke: It seems that you right. The comment inside H2O code says:
`Check that at least some columns are not-constant and not-all-NAs`

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAlgoCommonUtils.scala
jakubhava: 👍

PR #2168, opened 2020-06-08
[SW-2297] Exclude Content of site/.doctrees from SW Distribution Archive

mn-mikke:
The options `useDoctreeCache` and `doctreeCacheDirectory ` of  [`sphinx-gradle-plugin`](https://trustin.github.io/sphinx-gradle-plugin/configuration.html) seem to not have any effect.

### DIFF
--- a/doc/build.gradle
+++ b/doc/build.gradle
@@ -47,6 +47,10 @@ task substitute() {
   }
 }
 
+task deleteDocTrees(type: Delete) {
+  delete "${buildDir}/site/.doctrees"
+}
+
 sphinx {
   environments = ["PYTHONWARNINGS" : ""]
   if (VersionNumber.parse(sparkVersion) >= VersionNumber.parse(kubernetesSupportSinceSpark)) {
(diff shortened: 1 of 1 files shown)


jakubhava:
@mn-mikke Was there any problem with documentation?

doc/build.gradle
```diff
@@ -95,7 +99,8 @@ createVersionFile.dependsOn cleanDoc
 sphinx.dependsOn createVersionFile
 clean.dependsOn cleanDoc
 substitute.dependsOn sphinx
-sourcesDocJar.dependsOn substitute
+deleteDocTrees.dependsOn substitute
+sourcesDocJar.dependsOn deleteDocTrees
```
jakubhava, line 103: `sourcesDocJar` does not need to depend on anything as it creates the jar from the src dir where no files are generated

mn-mikke:
> @mn-mikke Was is the motivation for this change please?

The motivation is to reduce the size of SW distribution archive.

jakubhava:
Thanks! I also found the Slack discussion, good idea!

PR #2172, opened 2020-06-11
[SW-2302] Fix org.apache.hadoop.fs.FsUrlConnection cannot be cast to java.net.HttpURLConnection

jakubhava:
Gave @pastx custom build with this change and waiting for confirmation whether it helped or not


jakubhava:
@pastx can you please share your output? I think you have mentioned that the fix worked.
If yes, I think we can proceed with the change

pastx:
Yes, it works, sorry for the late reaction.

jakubhava:
np, thanks for confirmation!

PR #2173, opened 2020-06-11
[SW-2303] Create Sparkling Py4j Getaway for Steam

jakubhava:
Steam is trying to get rid of Livy. 

To achieve this, we can use the Py4j Gateway directly and allow the Spark python client to connect to remote Spark cluster without the Livy complexities.


core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,118 @@
...
+      logError("GatewayServer failed to bind; exiting")
+      System.exit(1)
+    } else {
+      logInfo(s"""Running Py4j Gateway on port $boundPort""")
+    }
+    // Exit on EOF or broken pipe to ensure that this process dies when the Python driver dies:
+    while (!(spark.sparkContext.isStopped || hc.isStopped())) {
+      Thread.sleep(1000)
+    }
+    logDebug("Exiting due to broken pipe from Python driver")
```
bilcus, line 45: This debug message is not accurate

jakubhava: Thanks!

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,117 @@
...
+import scala.io.Source
+import scala.util.Random
+
+object SparklingGateway extends Logging {
+
+  def main(args: Array[String]) {
+    val spark = SparkSession.builder.getOrCreate()
+    val conf = spark.sparkContext.getConf
+    val secret = readSecret(conf)
+    val hc = H2OContext.getOrCreate()
```
mn-mikke, line 25: is there any need to pass Spark/SW options?

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,117 @@
...
+    val address = InetAddress.getByName("0.0.0.0")
+    val builder = new GatewayServer.GatewayServerBuilder()
+      .javaPort(gatewayPort(conf))
```
mn-mikke, line 28: may it happen that the port is already occupied?

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,117 @@
...
+    if (option.isEmpty) {
+      throw new RuntimeException(s"Missing $PROP_GATEWAY_PORT")
+    } else {
+      option.get.toInt
+    }
+  }
+  private def gatewaySecretFileName(conf: SparkConf) = {
+    val option = conf.getOption(PROP_GATEWAY_SECRET_FILE_NAME._1)
+    if (option.isEmpty) {
+      throw new RuntimeException(s"Missing $PROP_GATEWAY_SECRET_FILE_NAME")
```
mn-mikke, line 74: shouldn't be there `s"Missing ${PROP_GATEWAY_SECRET_FILE_NAME._1}"`?

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,117 @@
...
+    } else {
+      SparkFiles.get(option.get)
+    }
+  }
+
+  private def pkcs12KeyStoreFileName(conf: SparkConf) = {
+    val option = conf.getOption(PROP_GATEWAY_KEYSTORE_FILE_NAME._1)
+    if (option.isEmpty) {
+      throw new RuntimeException(s"Missing $PROP_GATEWAY_KEYSTORE_FILE_NAME")
```
mn-mikke, line 83: ditto

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
```diff
@@ -0,0 +1,117 @@
...
+  private val PROP_GATEWAY_SECRET_FILE_NAME = ("spark.ext.h2o.py4j.gateway.secret.file.name", None)
+  // Sparkling Water expect keystore in a format of PKCS12 containing private key encoded as PKCS8 and certificate
+  // Also we expect that the PKCS12 keystore is not protected by a password
+  private val PROP_GATEWAY_KEYSTORE_FILE_NAME = ("spark.ext.h2o.py4j.gateway.keystore.file.name", None)
+  private val DUMMY_JKS_PASSWORD = Random.alphanumeric.take(20).mkString("").toCharArray
+
+  private def gatewayPort(conf: SparkConf) = {
+    val option = conf.getOption(PROP_GATEWAY_PORT._1)
+    if (option.isEmpty) {
+      throw new RuntimeException(s"Missing $PROP_GATEWAY_PORT")
```
mn-mikke, line 66: shouldn't be there s"Missing ${PROP_GATEWAY_PORT._1}"?

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
jakubhava: Yup, that can happen. The steam admin needs to pick a free port.

We could make this more robust by letting the code automatically select the port, but in that case we need to implement some sort of callback, to let Steam know about the selected port. I'm fine with that, but I would say this is mostly about what Steam prefers, CC @bilcus 

We already discussed this with @bilcus

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
mn-mikke: ok

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
bilcus: I almost forgot about the port.. I would prefer for SW to automatically select the port (pass 0 or -1 and it should pick a free port). However, callback might not be possible due to networking. Since you have the H2O proxy, could you add h2o-3 REST endpoint with this information?

jakubhava: Good idea, I can expose an endpoint Flow Proxy on the driver side which steam can query, does it make sense @bilcus ?

bilcus: Yes, please!

jakubhava:
@bilcus I have implemented the automatic port selection. Steam can then obtain the port by reading the following endpoint:

`FLOW_PROXY_URL/3/option?name=spark.ext.h2o.py4j.gateway.port`

I wanted to keep it generic so the rest of our code doesn't contain references to the gateway code

core/src/main/scala/ai/h2o/sparkling/backend/SparklingGateway.scala
jakubhava: Should be done

bilcus:
@jakubhava 
Nice, I like it, thanks!

PR #2177, opened 2020-06-15
[SW-2305] Ensure that call /3/Shutdown handles H2OContext stop in case of Sparkling Water ( via Flow Proxy)

jakubhava:
This is to allow Steam to properly stop Sparkling Water when calling shutdown endpoint

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
@@ -39,9 +39,7 @@ private[api] class ShutdownServlet extends ServletBase {
 
 object ShutdownServlet extends ServletRegister {
 
-  override protected def getRequestPaths(): Array[String] = {
-    Array("/3/shutdown")
-  }
+  override protected def getRequestPaths(): Array[String] = Array("/3/shutdown")
 
   override protected def getServlet(conf: H2OConf): Servlet = new ShutdownServlet
 }
--- a/core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
@@ -39,7 +39,7 @@ private[api] class ShutdownServlet extends ServletBase {
 
 object ShutdownServlet extends ServletRegister {
 
-  override protected def getEndpoints(): Array[String] = {
+  override protected def getRequestPaths(): Array[String] = {
     Array("/3/shutdown")
   }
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
```diff
@@ -0,0 +1,47 @@
...
+      case invalid => throw new H2ONotFoundArgumentException(s"Invalid endpoint $invalid")
+    }
+    sendResult(obj, resp)
+  }
+
+}
+
+object ShutdownServlet extends ServletRegister {
+
+  override protected def getEndpoints(): Array[String] = {
```
mn-mikke, line 42: Shouldn't `getEndpoints()`  return ip addresses?

jakubhava: Nope, we return the request path. The `getEndpoints` method should be renamed

jakubhava:
The motivation is mentioned in the description above. Basically, we want to allow steam to stop H2O & Spark stack with the resulting YARN app having status as SUCCESS. Currently, Steam is killing the app and the app has the status KILLED

core/src/main/scala/ai/h2o/sparkling/backend/api/rdds/RDDsServlet.scala
```diff
@@ -105,7 +105,7 @@ private[api] class RDDsServlet extends ServletBase {
 
 object RDDsServlet extends ServletRegister {
 
-  override protected def getEndpoints(): Array[String] = {
+  override protected def getRequestPaths(): Array[String] = {
```
mn-mikke, line 108: nit: could be in one line

core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
```diff
@@ -0,0 +1,47 @@
...
+      case invalid => throw new H2ONotFoundArgumentException(s"Invalid endpoint $invalid")
+    }
+    sendResult(obj, resp)
+  }
+
+}
+
+object ShutdownServlet extends ServletRegister {
+
+  override protected def getRequestPaths(): Array[String] = {
```
mn-mikke, line 42: nit: Could be in one line?

core/src/main/scala/ai/h2o/sparkling/backend/api/ShutdownServlet.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  override def doPost(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+    val obj = req.getPathInfo match {
+      case null =>
+        H2OContext.get().foreach(_.stop(stopSparkContext = true))
+      case invalid => throw new H2ONotFoundArgumentException(s"Invalid endpoint $invalid")
+    }
+    sendResult(obj, resp)
+  }
+
```
mn-mikke, line 37: nit: extra line

core/src/main/scala/ai/h2o/sparkling/backend/api/rdds/RDDsServlet.scala
jakubhava: 👍

PR #2178, opened 2020-06-15
[SW-2307] H2O Flow Proxy is not stopped as part of hc.stop() call

jakubhava:


core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
```diff
@@ -244,6 +244,7 @@ class H2OContext private[sparkling] (private val conf: H2OConf) extends H2OConte
           RestApiUtils.shutdownCluster(conf)
         }
       }
+      ProxyStarter.stopFlowProxy()
```
mn-mikke, line 247: What about the cases if a user calls `sc.stop()`? Is proxy going to be shutted down through spark hook?

jakubhava: Yup. The stop method is invoked during Spark shutdown hook

PR #2180, opened 2020-06-15
[SW-2309] Upgrade Gradle to 6.5

jakubhava:
Since we need to build a new docker image because of https://github.com/h2oai/sparkling-water/pull/2179, now it's a good time to also upgrade Gradle to the latest version


mn-mikke:
LGTM

PR #2181, opened 2020-06-15
[SW-2311] Expose InitialWeights and InitialBiases on H2ODeepLearning

mn-mikke:
There was a intermittent error:

020-06-16T19:13:40.196Z] ai.h2o.sparkling.backend.api.scalainterpreter.ScalaInterpreterServletTestSuite > classMethod FAILED

[2020-06-16T19:13:40.196Z]     ai.h2o.sparkling.backend.exceptions.RestApiCommunicationException at ScalaInterpreterServletTestSuite.scala:28

But that is not related. I have created a JIRA for this issue.

Also can you please rebase on top of current master? I have fixed the intermittent R errors which occured also during this build


py/tests/unit/with_runtime_sparkling/test_deeplearning.py
```diff
@@ -45,3 +46,28 @@ def testLoadAndTrainMojo(prostateDataset):
...
+            hidden=[3, ])
+
+    referenceDeepLearning = createInitialDeepLearningDefinition()
+    referenceModel = referenceDeepLearning.fit(traningDataset)
+    referenceResult = referenceModel.transform(testingDataset)
+
+    deepLearning = createInitialDeepLearningDefinition()
+    matrix0 = [[.1, .2, .3, .4], [.4, .5, .6, .7], [.7, .8, .9, .6]]
+    matrix1 = [[.2, .3, .4], ]
+    deepLearning.setInitialWeights([matrix0, matrix1])
```
mn-mikke, line 68: In the current proposal, `initialWeights` takes an array of matrices (`Array[Array[Array[Double]]]]`) and `initialBiases` an array of vectors (`Array[Array[Double]]]`) I'm wondering whether it won't be confusing for users.

E.g.: 
`matrix0 = [[.1, .2, .3, .4], [.4, .5, .6, .7], [.7, .8, .9, .6]]` -> `[.1, .2, .3, .4]` is a row
`deepLearning.setInitialBiases([[.1, .2, .3], [.1]])` -> `[.1, .2, .3]` is a vector (column)

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OAlgoParamsBase.scala
```diff
@@ -87,4 +89,44 @@ trait H2OAlgoParamsBase extends Params {
...
+      import spark.implicits._
+      val hc = H2OContext.ensure(
+        s"H2OContext needs to be created in order to train the ${this.getClass.getSimpleName} model. " +
+          "Please create one as H2OContext.getOrCreate().")
+      input.map(i => hc.asH2OFrame(spark.sparkContext.parallelize(i).toDF()).frameId)
+    }
+  }
+
+  protected def convert3dArrayToH2OFrameKeyArray(input: Array[Array[Array[Double]]]): Array[String] = {
+    if (input == null) {
```
jakubhava, line 121: The implementation of this method seems to be identical to the method above. What do you think about sharing some logic?

py/tests/unit/with_runtime_sparkling/test_deeplearning.py
jakubhava: Just an open idea. Would it make sense for this API to also expect Spark data frame with the data formatted as the matrices?

michalkurka:
@mn-mikke what I am thinking here is that the conversion to H2OFrame is cumbersome, the H2O API accepts frames because it is relatively straightforward to convert matrix-like objects in R&Py to H2OFrames and send them down to the backend.

However, on the Java-level we could skip the Frames and pass a matrix represented in a simpler way.

This will work as well. One question though: how are the created Frames disposed of? The model itself doesn't do anything to delete them. It is expected they will be garbage collected in R/Py or explicitly disposed of by the caller.

py/tests/unit/with_runtime_sparkling/test_deeplearning.py
michalkurka: > I'm wondering whether it won't be confusing for users

@mn-mikke what you could perhaps do is it use DenseMatrix from mllib, would that be an option? And DenseVector for biases.

mn-mikke: H2O type for `initial_weights` and `initial_biases` is array of H2OFrames. Each frame represents matrix or vector. Are you proposing to expose it as a list of dataframes?

I'm not sure what it's more user-friendly. To create a dataframe in Spark/Pyspark from scratch requires to write some boiler plate code and sometimes to specify the schema.

mn-mikke:
> how are the created Frames disposed of?

@michalkurka That's a good point, they will stay there forever and user doesn't have a chance to delete them. But IMHO, SW could delete the created frames after the model is trained. There is already another parameter on KMeans (`userPoints`) that will have a similar problem.

py/tests/unit/with_runtime_sparkling/test_deeplearning.py
mn-mikke: I like the idea of using `DenseMatrix` and `DenseVector`. I hope there won't be any restriction for older versions of spark. @kuba Or is this, what you meant? WDYT?

jakubhava: Yup, that would be perfect!

Just small note: I would try to use `ml` package as much as possible as `mllib` is considered  the "older one"

wendycwong: @mn-mikke The different sizes between the weight and bias should not be confusing to users.  They are pretty standard.

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OAlgoParamsBase.scala
mn-mikke: done ;-)

mn-mikke:
@jakubhava Do you have any further comments to this PR.

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OAlgoParamsBase.scala
jakubhava: Perfect!

jakubhava:
Nope, feel free to merge :)

PR #2182, opened 2020-06-15
[SW-2312] JsonSyntaxException when using setNfolds() on algorithm

jakubhava:
I was able to reproduce the issue just on grid search, but added tests for Automl and one algo based on H2OAlgorithm just to be sure

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
@@ -19,7 +19,6 @@ package ai.h2o.sparkling.ml.algos
 import java.util
 
 import ai.h2o.sparkling.api.generation.common.IgnoredParameters
-import ai.h2o.sparkling.{H2OContext, H2OFrame}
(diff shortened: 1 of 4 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -105,7 +106,10 @@ class H2OGridSearch(override val uid: String)
       internalFeatureCols: Array[String]): Array[H2OMOJOModel] = {
     val conf = H2OContext.ensure().getConf
     val endpoint = RestApiUtils.getClusterEndpoint(conf)
-    val skippedFields = Seq((classOf[GridSchemaV99], "summary_table"), (classOf[GridSchemaV99], "scoring_history"))
+    val skippedFields = Seq(
+      (classOf[GridSchemaV99], "cross_validation_metrics_summary"),
```
mn-mikke, line 110: Is this line fixing the bug?

jakubhava: Yup -> H2O is sending invalid JSON ( the same issue you bumped into and which led us to introducing of `skippedFields` field)

PR #2185, opened 2020-06-17
[SW-2318] Expose rand_link and rand_family on GLM

mn-mikke:
The output parameter are `randLink` and `randFamily`. I'm just wondering whether we shouldn't rename them to `randomLink` and `randomFamily`


jakubhava:
I'm in favor of renaming. At least it would be consistent with randomColumns. WDYT?

mn-mikke:
ok, wil rename

PR #2186, opened 2020-06-17
[SW-2317] Use leader node from the beginning of Rest API communication

jakubhava:
This change ensures that the REST API communicates with the leader node from the beginning.

In Internal backend, we can first report IP & port of some node which might not be a leader and we reassign it later. However, calls for the cloud lock and verification calls such for version and web open are done before this and therefore communicate with the non-leader node.

This is relevant only to the internal backend as in the external cluster, in the auto cluster start we provide automatically leader IP and port and in manual cluster start the users need to provide leader node IP.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
@@ -21,9 +21,10 @@ import java.net.InetAddress
 
 import ai.h2o.sparkling.H2OConf
 import ai.h2o.sparkling.backend.NodeDesc
+import org.apache.spark.SparkEnv
 import org.apache.spark.rpc.{RpcCallContext, RpcEnv, ThreadSafeRpcEndpoint}
 import water.util.Log
-import water.{H2O, H2ONode}
+import water.{H2O, H2ONode, Paxos}
 
 import scala.collection.JavaConverters._
 
@@ -48,12 +49,19 @@ class H2ORpcEndpoint(override val rpcEnv: RpcEnv) extends ThreadSafeRpcEndpoint 
 
     case StopEndpointMsg =>
       this.stop()
+    case LockClusterMsg() =>
+      Paxos.lockCloud("Locking the cloud from Sparkling Water as we have reached the expected cluster size.")
   }
 
   override def receiveAndReply(context: RpcCallContext): PartialFunction[Any, Unit] = {
     case StartH2OWorkersMsg(conf) =>
       val nodeDesc = InternalH2OBackend.startH2OWorker(conf)
       context.reply(nodeDesc)
+    case CheckClusterSizeMsg() =>
+      context.reply(H2O.CLOUD.size())
+    case IsLeaderNodeMsg() =>
+      val currentNode = NodeDesc(SparkEnv.get.executorId, H2O.SELF_ADDRESS.getHostAddress, H2O.API_PORT)
+      context.reply((currentNode, H2O.SELF.isLeaderNode))
   }
 }
 
@@ -62,3 +70,9 @@ case class StopEndpointMsg()
 case class FlatFileMsg(nodes: Array[NodeDesc], portOffset: Int)
 
 case class StartH2OWorkersMsg(conf: H2OConf)
+
+case class CheckClusterSizeMsg()
+
+case class LockClusterMsg()
+
+case class IsLeaderNodeMsg()
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
@@ -59,7 +59,7 @@ class H2ORpcEndpoint(override val rpcEnv: RpcEnv) extends ThreadSafeRpcEndpoint 
       context.reply(nodeDesc)
     case CheckClusterSizeMsg =>
       context.reply(H2O.CLOUD.size())
-    case IsLeaderNodeMsg =>
+    case GetLeaderNodeMsg =>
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -62,3 +70,9 @@ case class StopEndpointMsg()
 case class FlatFileMsg(nodes: Array[NodeDesc], portOffset: Int)
 
 case class StartH2OWorkersMsg(conf: H2OConf)
+
+case class CheckClusterSizeMsg()
```
mn-mikke, line 74: nit Couldn't we use case objects instead of case classes here?

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -48,12 +49,19 @@ class H2ORpcEndpoint(override val rpcEnv: RpcEnv) extends ThreadSafeRpcEndpoint
...
   }
 
   override def receiveAndReply(context: RpcCallContext): PartialFunction[Any, Unit] = {
     case StartH2OWorkersMsg(conf) =>
       val nodeDesc = InternalH2OBackend.startH2OWorker(conf)
       context.reply(nodeDesc)
+    case CheckClusterSizeMsg() =>
+      context.reply(H2O.CLOUD.size())
+    case IsLeaderNodeMsg() =>
+      val currentNode = NodeDesc(SparkEnv.get.executorId, H2O.SELF_ADDRESS.getHostAddress, H2O.API_PORT)
```
mn-mikke, line 63: is `H2O.SELF_ADDRESS.getHostAddress` going to be an IP address in 100% cases?

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
```diff
@@ -76,6 +89,23 @@ object InternalH2OBackend extends InternalBackendUtils {
     }
   }
 
+  private def lockCloud(endpoints: Array[RpcEndpointRef]): Unit = {
+    endpoints.foreach {
```
mn-mikke, line 93: Do we have send this message to all of them. it would cause any harm, but would the cloud be locked multiple times?

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -48,12 +49,19 @@ class H2ORpcEndpoint(override val rpcEnv: RpcEnv) extends ThreadSafeRpcEndpoint
...
+      context.reply((currentNode, H2O.SELF.isLeaderNode))
```
mn-mikke, line 64: Do you think is possible to add a flag of being a leader node to directly to `NodeDesc`?

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
jakubhava: Yup, from doc
>      * Returns the IP address string in the textual presentation.

But in the PR  I'm working on (with the fix for HTTPS) I will change this anyways to hostname so both IP and hostname is just fine here. I just did not want to this as part of this PR so the following PR contains the whole change.

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
jakubhava: Yup, will do

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
jakubhava: But just to make it clearer for the future, I have changed the method to send only NodeDesc of the leader node so worker nodes are not exposed at all

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
jakubhava: I have changed this method to co return Option[NodeDesc] we so don't have to send two items. I would prefer to keep the NodeDesc as it is as the information about leader is used only here but NodeDesc is used at a lot of places around the codebase

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalH2OBackend.scala
jakubhava: It is safe to call cloud lock multiple times, but not neccessary. I will invoke it just on the head of the list

core/src/main/scala/org/apache/spark/h2o/backends/internal/H2ORpcEndpoint.scala
```diff
@@ -48,17 +49,34 @@ class H2ORpcEndpoint(override val rpcEnv: RpcEnv) extends ThreadSafeRpcEndpoint
...
 
 case class FlatFileMsg(nodes: Array[NodeDesc], portOffset: Int)
 
 case class StartH2OWorkersMsg(conf: H2OConf)
+
+case object CheckClusterSizeMsg
+
+case object LockClusterMsg
+
+case object IsLeaderNodeMsg
```
mn-mikke, line 82: nit: if `Option[NodeDesc]` is returned shouldn't it be renamed to `getLeaderNodeMsg` or similar?

jakubhava: Good point, agree

PR #2189, opened 2020-06-17
[SW-2322] Expose plug_values on GLM

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPlugValues.scala
```diff
@@ -0,0 +1,68 @@
...
+trait HasPlugValues extends H2OAlgoParamsBase {
+  private val plugValues = new DictionaryParam(
+    this,
+    "plugValues",
+    "A map containing values that will be used to impute missing values of the training/validation frame, " +
+      "use with conjunction missing_values_handling = PlugValues)")
+
+  setDefault(plugValues -> null)
+
+  def getPlugValues(): Map[String, Double] ={
```
mn-mikke, line 37: @wendycwong Can plug values contain anything else then `Double`? Can be plug values used for imputing values of categorical columns, or they could be applied only on numeric columns.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
```diff
@@ -30,16 +30,14 @@ object IgnoredParameters {
...
-    "plug_values", // GLM
     "interaction_pairs", // GLM
     "beta_constraints", // GLM
     "rand_link", // GLM
     "rand_family", // GLM
     "random_columns", // GLM
     "initial_biases", // DeepLearning
     "initial_weights", // DeepLearning
-    "pretrained_autoencoder", // DeepLearning
-    "build_control") //AutoML
```
jakubhava, line 42: Is build control removed on purpose here?

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPlugValues.scala
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.spark.sql.types._
+
+import scala.collection.JavaConverters._
+
+trait HasPlugValues extends H2OAlgoParamsBase {
+  private val plugValues = new DictionaryParam(
+    this,
+    "plugValues",
+    "A map containing values that will be used to impute missing values of the training/validation frame, " +
+      "use with conjunction missing_values_handling = PlugValues)")
```
jakubhava, line 33: Should the doc be updated missing_values_handling -> `missingValuesHandling` ?

py/src/ai/h2o/sparkling/ml/params/HasPlugValues.py
```diff
@@ -0,0 +1,35 @@
...
+
+from ai.h2o.sparkling.ml.params.H2OTypeConverters import H2OTypeConverters
+
+
+class HasPlugValues(Params):
+    plugValues = Param(
+        Params._dummy(),
+        "plugValues",
+        "A dictionary containing values that will be used to impute missing values of the training/validation frame, "
+        "use with conjunction missing_values_handling = PlugValues)",
```
jakubhava, line 28: ditto about the doc

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
mn-mikke: Yes, AutoML is handled in `AutoMLIgnoredParameters`

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPlugValues.scala
mn-mikke: Yes, thanks!

py/src/ai/h2o/sparkling/ml/params/HasPlugValues.py
mn-mikke: Yes, thanks!

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
jakubhava: Thanks! I checked https://github.com/h2oai/sparkling-water/blob/master/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AutoMLIgnoredParameters.scala and build_control is missing in the list, should we add it there?

mn-mikke: IMO, it won't have any effect since we generate each parameter section separately. https://github.com/h2oai/sparkling-water/blob/master/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/Runner.scala#L112

`"build_control"` was added when i didn't understand how Automl parameters are structured, but forgot to remove it.

jakubhava: Perfect, thanks for great explanation!

michalkurka:
@mn-mikke looks good to me, there are some edge cases to be careful about

full test suite on the H2O side is in JUnits: https://github.com/h2oai/h2o-3/blob/master/h2o-algos/src/test/java/hex/glm/GLMPlugValuesTest.java

My questions:
1) Are categorical values handled?
2) What about interactions - can the dots (eg. X.y) cause any issue for JSON representation?

mn-mikke:
@michalkurka Thanks for the link!!!

1. categoricals -> That's the think I was not sure whether should be supported or not. I will change the type of dictionary/map to be able to hold anything as value.

2. I guess there shouldn't be problem in JSON serialization if columns has names with dots. The key type is String. But if I think about it there might be same escaping needed for conversion from a map to Spark DataFrame. In any case, I'll add some test to cover that.

Thanks!

py/tests/unit/with_runtime_sparkling/test_glm.py
```diff
@@ -60,3 +61,54 @@ def testPropagationOfPredictionCol(prostateDataset):
     model = algo.fit(prostateDataset)
     columns = model.transform(prostateDataset).columns
     assert True == (predictionCol in columns)
+
+
+def testPlugValuesAffectResult(spark, carsDatasetPath):
+    carsDataset=spark.read.csv(carsDatasetPath, header=True, inferSchema=True)
+    carsDataset=carsDataset.withColumn("economy_20mpg", carsDataset.economy_20mpg.cast("string"))
+    [traningDataset, testingDataset] = carsDataset.randomSplit([0.9, 0.1], 1)
+    testingDataset=testingDataset.na.fill({"economy_20mpg": "0"})
```
mn-mikke, line 70: @michalkurka `GLMMojoModel`/`EasyPredictModelWrapper` currently doesn't impute unknown categorical values. Is that correct?

michalkurka: not by default, if you enable `setConvertUnknownCategoricalLevelsToNa` then it will impute

mn-mikke: Even if I set `convertUnknownCategoricalLevelsToNa` to `true`, I get the following exception:
```
 java.lang.IllegalArgumentException: categorical value out of range
                at hex.genmodel.algos.glm.GlmMojoModel.glmScore0(GlmMojoModel.java:41)
                at hex.genmodel.algos.glm.GlmMojoModelBase.score0(GlmMojoModelBase.java:32)
                at hex.genmodel.easy.EasyPredictModelWrapper.predict(EasyPredictModelWrapper.java:889)
                at hex.genmodel.easy.EasyPredictModelWrapper.preamble(EasyPredictModelWrapper.java:867)
                at hex.genmodel.easy.EasyPredictModelWrapper.predictRegression(EasyPredictModelWrapper.java:783)
                at ai.h2o.sparkling.ml.models.H2OMOJOPredictionRegression$$anonfun$getRegressionPredictionUDF$2.apply(H2OMOJOPredictionRegression.scala:40)
                at ai.h2o.sparkling.ml.models.H2OMOJOPredictionRegression$$anonfun$getRegressionPredictionUDF$2.apply(H2OMOJOPredictionRegression.scala:37)
```

mn-mikke: @michalkurka Doesn't imputation work only if `missing_values_handling` is set to `"mean-impute"`?

https://github.com/h2oai/h2o-3/blob/43f8ab952a69a8bc9484bd0ffac909b6e3e820ca/h2o-genmodel/src/main/java/hex/genmodel/algos/glm/GlmMojoModelBase.java#L29

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPlugValues.scala
wendycwong: @mn-mikke:  plug values just need to be the same as the column type.  If they are double, plug value should be double.  For enum, they will take integer values as plug values.  However, you can store the plug values to be double and just convert it to integer when you actually do the plugging in.

py/tests/unit/with_runtime_sparkling/test_glm.py
wendycwong: @mn-mikke: GLM only do mean or mode imputation for missing values right now.

wendycwong: @mn-mikke:  Let me quickly take a look at the GLMMojo with convertUnknownCategoricalLevelsToNa to true.  Will fix if necessary.  Here is he JIRA: https://0xdata.atlassian.net/browse/PUBDEV-7628

michalkurka: @mn-mikke missing values should be imputed unless missing values handling == skip

If you select mean(mode for categoricals) or plug value imputation it should impute the unknown value.

I check the stacktrace and it leads to:
```
        int ival = (int) data[i];
        if (ival != data[i]) throw new IllegalArgumentException("categorical value out of range");
```
is there a chance a non-integer value is somehow getting in?

michalkurka: also I think your plug value "economy_20mpg": "0" is not actually in the training dataset, imputing a value that is not already in the dataset is not supported IMHO

michalkurka: @wendycwong please wait

wendycwong: @michalk:  Will wait.

mn-mikke: @michalkurka 
The column "economy_20mpg" contains only values 0, 1 + some values are missing:
```
"name","economy","cylinders","displacement","power","weight","acceleration","year","economy_20mpg"
"AMC Ambassador Brougham",13.0,8,360.0,175,3821,11.0,73,0
"AMC Ambassador DPL",15.0,8,390.0,190,3850,8.5,70,0
"AMC Ambassador SST",17.0,8,304.0,150,3672,11.5,72,0
"AMC Concord DL 6",20.200000000000003,6,232.0,90,3265,18.2,79,1
...
```
Spark parse it as integer and i convert it to string and SW to categorical:
```
>>> frame=hc.asH2OFrame(traningDataset)
>>> frame
name                       economy    cylinders    displacement    power    weight    acceleration    year    economy_20mpg
-----------------------  ---------  -----------  --------------  -------  --------  --------------  ------  ---------------
AMC Ambassador Brougham       13              8             360      175      3821            11        73                0
AMC Ambassador DPL            15              8             390      190      3850             8.5      70                0
AMC Ambassador SST            17              8             304      150      3672            11.5      72                0
AMC Concord                   19.4            6             232       90      3210            17.2      78                0
AMC Concord DL                18.1            6             258      120      3410            15.1      78                0
AMC Concord DL                23              4             151      nan      3035            20.5      82                1
AMC Concord DL 6              20.2            6             232       90      3265            18.2      79                1
AMC Gremlin                   18              6             232      100      2789            15        73                0
AMC Gremlin                   19              6             232      100      2634            13        71                0
AMC Gremlin                   20              6             232      100      2914            16        75                0

[362 rows x 9 columns]

>>> frame.types
{'name': 'enum', 'economy': 'real', 'cylinders': 'int', 'displacement': 'real', 'power': 'int', 'weight': 'int', 'acceleration': 'real', 'year': 'int', 'economy_20mpg': 'enum'}

```

Testing Frame then contains following values:
```
>>> testingDataset.groupBy("economy_20mpg").count().show()
+-------------+-----+
|economy_20mpg|count|
+-------------+-----+
|            0|   12|
|         null|    1|
|            1|   31|
+-------------+-----+

>>> testingDataset.printSchema
<bound method DataFrame.printSchema of DataFrame[name: string, economy: double, cylinders: int, displacement: double, power: int, weight: int, acceleration: double, year: int, economy_20mpg: string]>
```

mn-mikke: Details for plug values Frame:
```
label | type | Missing | Zeros | +Inf | -Inf | min | max | mean | sigma | cardinality | Actions
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
economy | real | 0 | 0 | 0 | 0 | 1.1000 | 1.1000 | 1.1000 | 0 | · | ·
weight | real | 0 | 0 | 0 | 0 | 4.4000 | 4.4000 | 4.4000 | 0 | · | ·
acceleration | real | 0 | 0 | 0 | 0 | 5.5000 | 5.5000 | 5.5000 | 0 | · | ·
economy_20mpg | enum | 0 | 1 | 0 | 0 | 0 | 0 | 0 | 0 | 1 | Convert to numeric
year | int | 0 | 0 | 0 | 0 | 2000.0 | 2000.0 | 2000.0 | 0 | · | Convert to enum
power | real | 0 | 0 | 0 | 0 | 3.3000 | 3.3000 | 3.3000 | 0 | · | ·
displacement | real | 0 | 0 | 0 | 0 | 2.2000 | 2.2000 | 2.2000 | 0 | · | ·
```

mn-mikke: @michalkurka I wouldn't bet on Sparkling Water that there is no bug, but shouldn't `GlmMojoModel` or `GlmMojoModelBase` contain some flag/variable identifing plug value imputation?

michalkurka: seems all good to me... I will have to debug this

michalkurka: @mn-mikke it does have that, see GlmMojoModelBase

```
  int _cats;
  int[] _catModes;
  int[] _catOffsets;

  int _nums;
  double[] _numMeans;
  boolean _meanImputation;
```

michalkurka: the plug values are stored there, it is reusing the same structures - I will write a test on H2O-3 side to see if we can reproduce just there

ml/src/main/scala/ai/h2o/sparkling/ml/params/DictionaryParam.scala
```diff
@@ -19,6 +19,8 @@ package ai.h2o.sparkling.ml.params
 
 import org.apache.spark.ml.param.{Param, Params}
 import org.json4s.DefaultFormats
+import org.json4s.JsonAST.JNull
```
jakubhava, line 22: NIT: Are these imports required?

mn-mikke: No, will remove them, thanks!

py/tests/unit/with_runtime_sparkling/test_glm.py
mn-mikke: Thanks @michalkurka for fixing the problem!

PR #2190, opened 2020-06-18
[SW-2323] Uprade shadowjar plugin to 6.0.0 ( fixes deprecation warnings)

jakubhava:


assembly-extensions/build.gradle
```diff
@@ -25,6 +25,7 @@ artifacts {
 }
 
 task embeddedAssemblyJar(type: Copy, dependsOn: shadowJar) {
+  duplicatesStrategy = 'include'
```
mn-mikke, line 28: Is there any problem with duplicates?

jakubhava: Not a problem -> Gradle just complained about the deprecation warning. But after your note, I had a second look and actually we were really copying the file again in case it was build before (we haven't cleaned the build dir)

PR #2192, opened 2020-06-18
[SW-2326] Add support for spark 3.0

jakubhava:
The tests will fail as the docker image does not contain Spark 3.0

I need to:
1) Merge this PR
2) Regenerate the docker images

Then tests will start also for spark 3.0


jakubhava:
Our code compiles fine against 3.0, I would prefer supporting it also from fix release as there will be OSS users wanting to use it

PR #2198, opened 2020-06-22
[SW-2335] Expose random_columns on GLM

mn-mikke:


mn-mikke:
@michalkurka @wendycwong @Pscheidl As I understand `random_columns` can't be present in the list of feature columns, but must be present in the training dataset as categorical columns, is that correct?

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasIgnoredCols.scala
```diff
@@ -0,0 +1,50 @@
...
+  def setIgnoredCols(value: Array[String]): this.type = set(ignoredCols, value)
+
+  override private[sparkling] def getH2OAlgorithmParams(trainingFrame: H2OFrame): Map[String, Any] = {
+    val newIgnoredCols = getIgnoredCols()
+    val existingMap = super.getH2OAlgorithmParams(trainingFrame)
+    val oldIgnoredCols = existingMap.getOrElse("ignored_columns", null).asInstanceOf[Array[String]]
+
+    if (oldIgnoredCols == null) {
+      existingMap + ("ignored_columns" -> newIgnoredCols)
+    } else if (newIgnoredCols == null) {
```
jakubhava, line 40: Just to understand this, this bit is here because we alter ignored columns as part of HasRandomCols, is that correct?

mn-mikke: Yes, `randomCols` must be part of `ignoredCols`, otherwise `fit` method will throws a strange exception about a missing values in the weight column which is not set. H2O Python client does the same thing (`randomCols` -> `ignoredCols`)

jakubhava: Thanks for the explanation!

wendycwong:
@mn-mikke Yes, you are correct.  Random columns are categorical only

mn-mikke:
Thanks a lot @wendycwong!

PR #2205, opened 2020-06-29
[SW-2343] Fix MOJO Model Predictions on Dataframes with ArrayType or Vector

mn-mikke:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOFlattenedInput.scala
```diff
@@ -43,11 +48,33 @@ trait H2OMOJOFlattenedInput {
     }
   }
 
+  protected def getRelevantColumnNames(flatDataFrame: DataFrame, inputs: Array[String]): Array[String] = {
+    val result = ArrayBuffer[String]()
+
+    def addIfPrefixExists(prefix: String): Unit = {
+      if (inputs.exists(i => i.startsWith(prefix + ".") && Try(i.substring(prefix.length + 1).toInt).isSuccess)) {
```
jakubhava, line 55: Nit: better name for the variable `i` ?

jakubhava, line 55: Nit: Just thinking out loud -> should we also test edge cases such as array of array and vector of vectors?

mn-mikke: > vector of vectors
This is not possible. Vector is an array of doubles.

> array of array
There is already a logic in codebase that could do that, but as far i remember we didn't use it because of making predictions on streamed dataframes. To be able to deeply nested structures, we would need to materialize the whole dataframe to get array sizes.

mn-mikke: ok, will change it. Thanks!

jakubhava: I see, all good then!

mn-mikke:
@jakubhava This PR will cause a breaking change for MOJO models trained on vectors.Do you think, it's ok?

jakubhava:
@mn-mikke I think that this breaking change is ok as it is fixing the behavior

jakubhava:
I was going over the PRS and saw that this PR failed on the issue already fixed on master. Rebase should help :)

jakubhava:
Failed on unrelated tests, I think we are good to merge :)

mn-mikke:
@jakubhava Such error happens very often on Spark 3.0 tests

jakubhava:
Yup, I will have a look tomorrow. It's failing on the newly added Py4J gateway test

PR #2206, opened 2020-06-30
[SW-2344] Fix mojo test -> H2O added a new parameter and therefore number of parameters does not match now

jakubhava:
H2O exposed a new parameter h2o-core/src/main/java/hex/Model.java which leads to failure of this test


jakubhava:
@mn-mikke I will create a new PR for the fix release where I expose the parameter. On the master branch, we have it auto-generated

PR #2209, opened 2020-07-01
[SW-352] Flow UI Scala Repl: use paste mode to interpret commands passed through Flow UI

jakubhava:
Oldtimer JIRA. Implemented while waiting for the K8s tests to finish.

See https://0xdata.atlassian.net/browse/SW-352 for more details

### DIFF
--- a/repl/src/main/scala/ai/h2o/sparkling/repl/BaseH2OInterpreter.scala
+++ b/repl/src/main/scala/ai/h2o/sparkling/repl/BaseH2OInterpreter.scala
@@ -96,7 +96,10 @@ private[repl] abstract class BaseH2OInterpreter(val sparkContext: SparkContext, 
     initBeforeRunningCode(code)
     // Redirect output from console to our own stream
     scala.Console.withOut(consoleStream) {
-      try loop()
+      try {
+        val text = Iterator.continually(in.readLine("")).takeWhile(x => x != null).mkString("\n")
+        command(text)
+      }
       catch AbstractOrMissingHandler()
     }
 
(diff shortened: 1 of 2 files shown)


repl/src/main/scala/ai/h2o/sparkling/repl/BaseH2OInterpreter.scala
```diff
@@ -96,8 +96,10 @@ private[repl] abstract class BaseH2OInterpreter(val sparkContext: SparkContext,
     initBeforeRunningCode(code)
     // Redirect output from console to our own stream
     scala.Console.withOut(consoleStream) {
-      try loop()
-      catch AbstractOrMissingHandler()
+      try {
+        val text = Iterator.continually(in.readLine("")).takeWhile(x => x != null).mkString("\n")
```
mn-mikke, line 100: Why  `""` is passed to `readline`?

jakubhava: Based on the method `readLine` it seems like it's a prompt the method expects, but in this case we don't have any, so that is why ""

 ```  def readLine(prompt: String): String =
    // hack necessary for OSX jvm suspension because read calls are not restarted after SIGTSTP
    if (isMac) restartSysCalls(readOneLine(prompt), reset())
    else readOneLine(prompt)
```

mn-mikke: oh ok, thanks!

PR #2210, opened 2020-07-01
[SW-1510] Fix misleading error message from incompatible Java version

jakubhava:
See https://0xdata.atlassian.net/browse/SW-1510

### DIFF
--- a/bin/sparkling-env.sh
+++ b/bin/sparkling-env.sh
@@ -124,14 +124,13 @@ if [ -z $(which java) ]; then
     exit -1
 fi
 
+version=$(java -version 2>&1 | awk -F '"' '/version/ {print $2}')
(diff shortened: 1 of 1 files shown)


bin/sparkling-env.sh
```diff
@@ -118,11 +118,19 @@ if [ -z "$TOPDIR" ]; then
...
 fi
 
-function checkJava(){
+function checkJava() {
 if [ -z $(which java) ]; then
     echo "Java is not installed. Please install Java first before continuing with Sparkling Water."
     exit -1
 fi
+
+if [ "$SPARK_MAJOR_VERSION" != "2.1" ]; then
```
mn-mikke, line 127: Should there be a check for Spark 2.1 -> java 7+

jakubhava: Good point, will do

PR #2216, opened 2020-07-08
[SW-2285] Convert MapTypes under detailed_prediction to StructTypes

mn-mikke:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPredictionAnomaly.scala
```diff
@@ -47,16 +48,15 @@ trait H2OMOJOPredictionAnomaly {
...
     val scoreField = StructField("score", predictionColType, nullable = false)
     val fields = if (getWithDetailedPredictionCol()) {
       val normalizedScoreField = StructField("normalizedScore", predictionColType, nullable = false)
       scoreField :: normalizedScoreField :: Nil
     } else {
       scoreField :: Nil
     }
-    Seq(StructField(getDetailedPredictionCol(), StructType(fields), nullable = true))
-
+    StructType(fields)
```
jakubhava, line 59: `object H2OMOJOPredictionAnomaly` with the defined case classes can be removed

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPrediction.scala
```diff
@@ -83,18 +85,21 @@ trait H2OMOJOPrediction
   }
 
   override def getDetailedPredictionColSchema(): Seq[StructField] = {
+    Seq(StructField(getDetailedPredictionCol(), gePredictionSchema(), nullable = true))
```
jakubhava, line 88: typo: `gePredictionSchema` -> `getPredictionSchema`

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPrediction.scala
```diff
@@ -51,6 +52,7 @@ trait H2OMOJOPrediction
   }
 
   def getPredictionUDF(): UserDefinedFunction = {
+    SQLConf.get.setConfString("spark.sql.legacy.allowUntypedScalaUDF", "true")
```
jakubhava, line 55: I'm fine with this for now. Just qq: Is there a change that future versions will remove support for this, what the alternative would be then?

mn-mikke: In future versions, we could use `def udf(f: UDF2[_, _, _], returnType: DataType): UserDefinedFunction` but is available since Spark 2.3 or create `SparkUserDefinedFunction` function directly (in spark 2.1 is named UserDefinedFunction).

I will implement some workaround for Spark 3.0 so that the option is not needed

jakubhava: Thanks for the info!

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPredictionAnomaly.scala
mn-mikke: Good catch, thanks!

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOPredictionRegression.scala
```diff
@@ -95,27 +87,4 @@ trait H2OMOJOPredictionRegression extends PredictionWithContributions {
     getWithLeafNodeAssignments() && H2OMOJOCache.canGenerateLeafNodeAssignments(
```
mn-mikke, line 87: @jakubhava why the condition for leaf node assignmets in regression problems is different from conditions in Binomial or Multinomial classification problems?

jakubhava: Good question! That might be probably a bug.

Originally, the code had the condition as in binomial & multinomial. But our regression tests started to fail on GLM as it can't produce leaf node assignments (only tree-based models can) & I think I forgot to apply the change to other parts as well.

Let me fixed that in a followup PR

mn-mikke: AFIAK, contributions are not supported by GLM eighter. If the property is disabled by default, a user won't hit the problem anyway. I'm just wondering what it's better behavior, throw an exception or silently ignore the configuration of the property. An exception might not be a bad solution. WDYT?

jakubhava: Created https://github.com/h2oai/sparkling-water/pull/2218, let's continue there :)

Regarding the exception -> I think that a user warning would do in this case, WDYT?

mn-mikke: Yes, warning is fine.

PR #2214, opened 2020-07-07
[SW-2349] Enable leaf node assignment for H2OMOJOModel

jakubhava:

PR #2218, opened 2020-07-09
[SW-2349][FollowUp] Enable leaf node assignment for H2OMOJOModel (#2214)

jakubhava:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -314,10 +314,10 @@ object H2OMOJOCache extends H2OMOJOBaseCache[EasyPredictModelWrapper, H2OMOJOMod
     config.setConvertUnknownCategoricalLevelsToNa(model.getConvertUnknownCategoricalLevelsToNa())
     config.setConvertInvalidNumbersToNa(model.getConvertInvalidNumbersToNa())
     if (canGenerateContributions(config.getModel)) {
-      config.setEnableContributions(model.getWithDetailedPredictionCol())
+      config.setEnableContributions(model.getWithContributions() && model.getWithDetailedPredictionCol())
```
jakubhava, line 317: I think this was another bug which was there -> we should enable contributions only if both contributions and detailed prediction col are true

scoring/src/main/scala/ai/h2o/sparkling/ml/models/PredictionWithContributions.scala
```diff
@@ -33,4 +34,8 @@ trait PredictionWithContributions {
       contributionNames.zip(contributions).toMap
     }
   }
+
+  def supportsContributions(): Boolean = {
+    getWithContributions() && H2OMOJOCache.canGenerateContributions(H2OMOJOCache.getMojoBackend(uid, getMojo, this).m)
```
mn-mikke, line 39: Can we generate warning here? Or better throw an exception that contributions are not supported in the model? I'm afraid that warning will get easily lost in tons of spark logging messages.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/PredictionWithLeafNodeAssignments.scala
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package ai.h2o.sparkling.ml.models
+
+trait PredictionWithLeafNodeAssignments {
+  self: H2OMOJOModel =>
+  def supportsLeafNodeAssignments(): Boolean = {
+    getWithLeafNodeAssignments() && H2OMOJOCache.canGenerateLeafNodeAssignments(
+      H2OMOJOCache.getMojoBackend(uid, getMojo, this).m)
```
mn-mikke, line 24: ditto

scoring/src/main/scala/ai/h2o/sparkling/ml/models/PredictionWithContributions.scala
jakubhava: If we decide to throw an exception then it makes more sense to throw it right away in the `canGenerateContributions` method as that check is done in `loadMojoBackend` before any computation starts.

mn-mikke: ok, makes sense

jakubhava:
@mn-mikke Please have another look. We fail right away if we don't support the given option.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -292,19 +292,30 @@ object H2OMOJOCache extends H2OMOJOBaseCache[EasyPredictModelWrapper, H2OMOJOMod
...
-      case _: PredictContributionsFactory =>
+      case m: PredictContributionsFactory =>
         val modelCategory = model.getModelCategory
-        modelCategory == ModelCategory.Regression || modelCategory == ModelCategory.Binomial
-      case _ => false
+        if (modelCategory != ModelCategory.Regression && modelCategory != ModelCategory.Binomial) {
+          throw new IllegalArgumentException(s"""
+              | Computing contributions on MOJO of type '${m.getModelCategory}' is only supported for regression
+              | and binomial model categories!
+              | The column 'contributions' will not be generated."
```
mn-mikke, line 301: I would delete this sentence. In case of the exceptaion, nothing will be generated.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -292,19 +292,30 @@ object H2OMOJOCache extends H2OMOJOBaseCache[EasyPredictModelWrapper, H2OMOJOMod
...
+              |""".stripMargin)
+        }
+        true
+      case unsupported =>
+        throw new IllegalArgumentException(s"""
+            | Computing contributions is not allowed on MOJO of type '${unsupported.getClass}'!
+            | The column 'contributions' will not be generated.")""".stripMargin)
```
mn-mikke, line 308: ditto

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -292,19 +292,30 @@ object H2OMOJOCache extends H2OMOJOBaseCache[EasyPredictModelWrapper, H2OMOJOMod
...
-  private[sparkling] def canGenerateLeafNodeAssignments(model: GenModel): Boolean = {
+  private def canGenerateLeafNodeAssignments(model: GenModel): Boolean = {
     model match {
       case _: TreeBackedMojoModel => true
       case _ =>
-        logWarning("Computing Leaf Node Assignments is only available on Tree based models!")
-        false
+        throw new IllegalArgumentException("""
+            | Computing leaf node assignments is only available on tree based models!
+            | The column 'leafNodeAssignments' will not be generated""".stripMargin)
```
mn-mikke, line 318: ditto

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
jakubhava: Good point, will do

PR #2223, opened 2020-07-13
[SW-2166] Automatic smoke tests on DBC

jakubhava:
Also added a new job named `TESTS_DATABRICKS` which runs every day


ci/Jenkinsfile-databricks
```diff
@@ -0,0 +1,246 @@
...
+            '
+        """, returnStdout: true).trim()
+    def parsedJson = new JsonSlurperClassic().parseText(rawJson)
+    def runId = parsedJson.run_id
+    waitForJobToFinish(runId, language)
+}
+
+def waitForJobToFinish(jobId, language) {
+    def done = false
+    while(!done) {
```
mn-mikke, line 135: Is a progress reported anywhere?

ci/Jenkinsfile-databricks
```diff
@@ -0,0 +1,246 @@
...
+""", returnStdout: true).trim()
+    def jsonSlurper = new JsonSlurperClassic()
+    def cfg = jsonSlurper.parseText(jsonOutput)
+    return cfg["cluster_id"]
+}
+
+def waitForCluster(clusterId) {
+    def ready = false
+    while(!ready) {
+        def json = sh(script: "databricks clusters get --cluster-id ${clusterId}", returnStdout: true).trim()
```
mn-mikke, line 66: What happens if the creation of the cluster fails?

ci/Jenkinsfile-databricks
```diff
@@ -0,0 +1,246 @@
...
+    def sparkVersions = props["supportedSparkVersions"].split(" ").toList()
+    def boundaryVersion = props["databricksTestSinceSpark"]
+    def list = new ArrayList<String>()
+    list.addAll(sparkVersions.subList(sparkVersions.indexOf(boundaryVersion), sparkVersions.size()))
+    return list
+}
+
+def waitForArtifactToAttach(clusterId, artifact, type) {
+    def ready = false
+    while(!ready) {
```
mn-mikke, line 88: ditto: what happens if the operation fails?

gradle.properties
```diff
@@ -30,5 +30,6 @@ spark=3.0
 version=3.32.0.1-1-SNAPSHOT
 # Spark version from which is Kubernetes Supported
 kubernetesSupportSinceSpark=2.4
+databricksTestSinceSpark=2.4
```
mn-mikke, line 33: Why older versions are no supported?

ci/Jenkinsfile-databricks
jakubhava: You can see progress in Jenkins console

ci/Jenkinsfile-databricks
jakubhava: Tests fail

ci/Jenkinsfile-databricks
jakubhava: Tests fail, but I can do nicer cleanup

gradle.properties
jakubhava: DBC supports only Spark 2.4 +

ci/Jenkinsfile-databricks
mn-mikke: So each loop iteration prints out something to stdout?

jakubhava:
yes, on AWS. But I think it is just fine to test it only on AWS. What we want to test is the DBC, not the underlying hw

ci/Jenkinsfile-databricks
jakubhava: yes

ci/Jenkinsfile-databricks
mn-mikke: So no change that the test will get stuck in the loop?

ci/Jenkinsfile-databricks
mn-mikke: ok, thanks!

ci/Jenkinsfile-databricks
jakubhava: Good point about the job being stuck -> I will add a timeout to the job specification.

I will also do a proper cleanup if something bad happens

gradle.properties
mn-mikke: I see, DBC versions with older Spark versions are unsupported, thanks!

ci/Jenkinsfile-databricks
jakubhava: I have added more explicit handling in case something wrong happens during start of the cluster

ci/Jenkinsfile-databricks
jakubhava: Done :)

PR #2224, opened 2020-07-13
[SW-2029] Add H2O GLRM to Algo API

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -177,6 +179,31 @@ class H2OGridSearch(override val uid: String)
...
+          JField("name", JString(result)) <- obj
+        } yield result
+        result.head
+      }
+
+      val metricValueOption = for {
+        JObject(obj) <- ast
+        JField("model_summary", JObject(modelSummary)) <- obj
+        JField("columns", JArray(columns)) <- modelSummary
+        ("final_objective_value", index) <- columns.arr.map(getColumnName).zipWithIndex
```
mn-mikke, line 198: @wendycwong I'm extracting the metric from model_summary on MOJO model. The name of the value is `final_objective_value`. Is this the value that we try to minimize during the model training?

py/tests/unit/with_runtime_sparkling/test_glrm.py
```diff
@@ -0,0 +1,125 @@
...
+    unit_test_utils.assert_data_frames_have_different_values(reference, result)
+
+def testLossByColHasEffectOnTrainedModel(arrestsDataset):
+    referenceAlgo = getPreconfiguredAlgorithm()
+    referenceModel = referenceAlgo.fit(arrestsDataset)
+    reference = referenceModel.transform(arrestsDataset)
+
+    algo = getPreconfiguredAlgorithm()
+    algo.setLossByCol(["absolute", "huber"])
+    algo.setLossByColNames(["Murder", "Rape"])
```
mn-mikke, line 121: I've replace `loss_by_col_idx` with `lossByColNames` which reference column by their names. Column indices wouldn't work well in SW. I also considered to merge `lossByCol` and `lossByColNames` into one map/dictionary parameter, but I didn't that since `lossByCol` can be used in gridSearch according to the documentation.

py/tests/unit/with_runtime_sparkling/test_glrm.py
```diff
@@ -0,0 +1,125 @@
...
+def testUserYHasEffectOnTrainedModel(spark, arrestsDataset):
+    referenceAlgo = getPreconfiguredAlgorithm()
+    referenceModel = referenceAlgo.fit(arrestsDataset)
+    reference = referenceModel.transform(arrestsDataset)
+
+    userYData = [(5.412,  65.24,  -7.54, -0.032),
+                 (2.212,  92.24, -17.54, 23.268),
+                 (0.312, 123.24,  14.46,  9.768),
+                 (1.012,  19.24, -15.54, -1.732)]
+    userYDataFrame = spark.createDataFrame(userYData, ['a', 'b', 'c', 'd'])
```
mn-mikke, line 105: This PR proposes to exposes userX and userY as a dataframe. An alternative could be a DenseMatrix from mllib, but I assume that both matrices could be quite large. WDYT?

doc/src/site/sphinx/ml/sw_glrm.rst
```diff
@@ -0,0 +1,105 @@
...
+
+        .. code:: python
+
+            model.getModelDetails()
+
+        Transform your dataset, the output will be under the ``prediction`` column.
+
+        .. code:: python
+
+            model.transform(sparkDF).show(truncate = False)
```
jakubhava, line 105: It would be good to mention how to obtain the reduced frame.

We expose parameter `representation_name` which is the key of the frame where the reduced frame gets saved as H2OFrame.

WDYT?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -146,26 +149,25 @@ class H2OGridSearch(override val uid: String)
...
-  private def sortGridModels(gridModels: Array[H2OMOJOModel]): Array[H2OMOJOModel] = {
+  private def sortGridModels(gridModels: Array[H2OMOJOModel], algoName: String): Array[H2OMOJOModel] = {
     val metric = if (getSelectBestModelBy() == H2OMetric.AUTO.name()) {
       val category = H2OModelCategory.fromString(gridModels(0).getModelCategory())
       category match {
         case H2OModelCategory.Regression => H2OMetric.RMSE
         case H2OModelCategory.Binomial => H2OMetric.AUC
         case H2OModelCategory.Multinomial => H2OMetric.Logloss
         case H2OModelCategory.Clustering => H2OMetric.TotWithinss
+        case H2OModelCategory.DimReduction if algoName == "glrm" => H2OMetric.GLRMMetric
```
jakubhava, line 164: NIT: `if algoName == "glrm"` is required?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -177,6 +179,31 @@ class H2OGridSearch(override val uid: String)
       "The fit method of the grid search must be called first to be able to obtain a list of models.")
   }
 
+  private def getMetricValue(model: H2OMOJOModel, metric: H2OMetric): Double = metric match {
+    case H2OMetric.GLRMMetric =>
+      val ast = parse(model.getModelDetails())
+
+      def getColumnName(value: JValue): String = {
```
jakubhava, line 186: WDYT of extracting the part for obtaininig GLRM metric to a different method so the case is smaller?

ml/src/test/scala/ai/h2o/sparkling/ml/algos/H2OGLRMTestSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+    pipeline.write.overwrite().save("ml/build/glrm_pipeline")
+    val loadedPipeline = Pipeline.load("ml/build/glrm_pipeline")
+    val model = loadedPipeline.fit(trainingDataset)
+    val expected = roundResult(model.transform(testingDataset))
+
+    model.write.overwrite().save("ml/build/glrm_pipeline_model")
+    val loadedModel = PipelineModel.load("ml/build/glrm_pipeline_model")
+    val result = roundResult(loadedModel.transform(testingDataset))
+
+    TestUtils.assertDataFramesAreIdentical(expected, result)
```
jakubhava, line 73: What do you think about testing also the produced frame in `responseName` ?

py/tests/unit/with_runtime_sparkling/test_glrm.py
jakubhava: I like the data frame approach!

py/tests/unit/with_runtime_sparkling/test_glrm.py
jakubhava: Good catch!

jakubhava:
Also just realized that it might be hard to obtain the reduced frame if we do not explicitly set the responseName because H2O creates the name as `"GLRMLoading_" + Key.rand()` if not specified.

What do you think of generating some default name on SW side and mentioning that in docs?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
mn-mikke: For now, I can remove it. But in future with more algorithms, there will have to be such a condition since the metric is specific only to GLRM.

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
mn-mikke: np, will do

ml/src/test/scala/ai/h2o/sparkling/ml/algos/H2OGLRMTestSuite.scala
mn-mikke: ok, np. will thanks!

doc/src/site/sphinx/ml/sw_glrm.rst
mn-mikke: will do thanks!

mn-mikke:
@jakubhava
>Also just realized that it might be hard to obtain the reduced frame if we do not explicitly set the responseName because H2O creates the name as "GLRMLoading_" + Key.rand() if not specified.

Is a user able to get `representation_name` through h2O-3 if it the value wasn't explicitly specified?

> What do you think of generating some default name on SW side and mentioning that in docs?

We could generate some default value, but would it cause collisions in DKV if train algorithm multiple times?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
jakubhava: It is fine, we can keep this there, was just asking :)

jakubhava:
Good points on the last comments:

> Is a user able to get representation_name through h2O-3 if it the value wasn't explicitly specified?
Don't know, maybe @wendycwong knows?

> We could generate some default value, but would it cause collisions in DKV if the train algorithm multiple times?
What I meant generating a unique DKV name and setting it to the default value of the parameter( if that's possible). So a user could get the name via the algorithm parameter. But it's true that in some real edge cases even this key could become taken before we run the algo. But just thinking out loud.

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
wendycwong: @mn-mikke: that is correct.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
```diff
@@ -22,7 +22,8 @@ object IgnoredParameters {
     "r2_stopping", // All
     "max_confusion_matrix_size", // Deep Learning
     "col_major", // Deep Learning
-    "max_hit_ratio_k") // GBM, DRF, Deep Learning
+    "max_hit_ratio_k", // GBM, DRF, Deep Learning
+    "loading_name") // GLRM
```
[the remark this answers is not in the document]
mn-mikke: Thanks @wendycwong for the comment! I think this is already addressed. We generate parameters from h2o V3 classes. [This list](https://github.com/h2oai/sparkling-water/pull/2224/files#diff-47e4179c219099038269faa68f455740R21) indicates parameters that are deprecated in H2O-3 and won't be generated to SW API. Since `representation_name` is defined on V3 class and it's not on a blacklist anywhere, the relevant parameter will be generated. [Here](https://github.com/h2oai/sparkling-water/pull/2224/files#diff-51830a9432a74304b2536d848462099dR132) is a test using `representation_name`.

PR #2226, opened 2020-07-15
[SW-2356] Fix timeout on long running Rest API commands through Proxy

jakubhava:
Increase the default value to 5 minutes and make it configurable as well

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
@@ -171,7 +171,7 @@ trait SharedBackendConf {
 
   def icedDir: Option[String] = sparkConf.getOption(PROP_ICED_DIR._1)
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
```diff
@@ -171,6 +171,8 @@ trait SharedBackendConf {
 
   def icedDir: Option[String] = sparkConf.getOption(PROP_ICED_DIR._1)
 
+  def sessionTimeout: Int = sparkConf.getInt(PROP_SESSION_TIMEOUT._1, PROP_SESSION_TIMEOUT._2)
```
mn-mikke, line 174: I'm wondering whether it isn't better to choose a different name for the property. IMHO, people will associate this name with `SparkSession` or `H2OContext`. WDYT?

jakubhava: Good point, I didn't think about that. Maybe "RestApiTimeout" ?

PR #2227, opened 2020-07-15
[SW-2357] Some tests in ml package are not being run

jakubhava:


ml/src/test/scala/ai/h2o/sparkling/ml/algos/TransformSchemaTestSuite.scala
```diff
@@ -32,14 +32,17 @@ trait TransformSchemaTestSuite extends FunSuite with Matchers {
 
   protected def expectedPredictionCol: StructField
 
+  protected def getWithLeafNodeAssignments: Boolean = false
```
mn-mikke, line 35: why this method is needed?

jakubhava: I had to introduce the method because, for example both AnomalyPredictionTestSuite and WordEmbeddingTestSuite uses the test structure defined in `TransformSchemaTestSuite`. I want to test the leaf node assignments in `AnomalyPredictionTestSuite` but leaf node assignments are not available on `WordEmbeddings`, that is why I created this method to be able to tell in test whether we should generate the algo with or without leaf node assignments

mn-mikke: oh ok, thanks!

PR #2236, opened 2020-07-20
[SW-2371] Missing getWithLeafNodeAssignments() Method on MOJOModelBase In Python & R

mn-mikke:


jakubhava:
@mn-mikke can you please rebase on top of the master? I have upgraded docker image as R tests required to have newer version of `devtools` lib

PR #2238, opened 2020-07-21
[SW-2372] Expose option used for waiting before the clouding starts in internal backend

jakubhava:
CC: @FengAtH2O

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
@@ -51,6 +51,8 @@ trait InternalBackendConf extends SharedBackendConf {
 
   def hdfsConf: Option[String] = sparkConf.getOption(PROP_HDFS_CONF._1)
 
+  def waitBeforeCloudingMillis:  Int = sparkConf.getInt(PROP_WAIT_BEFORE_CLOUDING._1, PROP_WAIT_BEFORE_CLOUDING._2)
+
   /** Setters */
   def setNumH2OWorkers(numWorkers: Int): H2OConf = set(PROP_CLUSTER_SIZE._1, numWorkers.toString)
 
@@ -77,6 +79,8 @@ trait InternalBackendConf extends SharedBackendConf {
     }
     set(PROP_HDFS_CONF._1, hdfsConfigTempFile.getAbsolutePath)
   }
+
+  def setWaitBeforeCloudingMillis(timeout: Int): H2OConf = set(PROP_WAIT_BEFORE_CLOUDING._1, timeout.toString)
 
   def internalConfString: String =
     s"""Sparkling Water configuration:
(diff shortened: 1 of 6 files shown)


doc/src/site/sphinx/configuration/configuration_properties.rst
```diff
@@ -264,6 +264,10 @@ Internal backend configuration properties
 |                                                    |                |                                                 | settings and other HDFS-related        |
 |                                                    |                |                                                 | configurations.                        |
 +----------------------------------------------------+----------------+-------------------------------------------------+----------------------------------------+
+| ``spark.ext.h2o.wait.before.clouding``             | ``0``          | ``setWaitBeforeCloudingMillis(Int)``            | ow long we should wait before we start |
```
mn-mikke, line 267: typo: `ow long`

mn-mikke:
Any reason for executing the `SpreadRDD`  after expiration of the timeout? Couldn't the SpreadRDD be executed in the loop during the period specified by the timeout value?

jakubhava:
Yup, the reason is that the spread rdd is influenced by options such as num retries, subseq tries etc. During the loops, we might end earlier if these counters hit zero

doc/src/site/sphinx/configuration/configuration_properties.rst
jakubhava: Thanks!

jakubhava:
@mn-mikke What do you think about this change please?

jakubhava:
@mn-mikke thanks for your input! I'll give it another thought

jakubhava:
@mn-mikke I gave it a thought and try to incorporate the timeout into the main build loop. Can you please have a look on the PR one more time?

doc/src/site/sphinx/configuration/configuration_properties.rst
```diff
@@ -264,6 +264,16 @@ Internal backend configuration properties
 |                                                    |                |                                                 | settings and other HDFS-related        |
 |                                                    |                |                                                 | configurations.                        |
 +----------------------------------------------------+----------------+-------------------------------------------------+----------------------------------------+
+| ``spark.ext.h2o.internal.clouding.timeout``        | ``0``          | ``setInternalBackendCloudingTimeout(Int)``      | Specifies how long the clouding should |
```
mn-mikke, line 267: `Specifies how long the clouding should last.` -> `Specifies how long the discovering of Spark executors should last.` Clouding up of H2O cluster will happen afterwards, won't it?

core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
```diff
@@ -115,4 +119,7 @@ object InternalBackendConf {
 
   /** Path to whole Hadoop configuration serialized into XML readable by org.hadoop.Configuration class */
   val PROP_HDFS_CONF: (String, None.type) = ("spark.ext.h2o.hdfs_conf", None)
+
+  /** Timeout for the clouding, unit is milliseconds */
+  val PROP_CLOUDING_TIMEOUT: (String, Int) = ("spark.ext.h2o.internal.clouding.timeout", 0)
```
mn-mikke, line 124: WDYT about `spark.ext.h2o.spreadrdd.retries.timeout`?

doc/src/site/sphinx/configuration/configuration_properties.rst
jakubhava: Yup, good catch!

core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
jakubhava: 👍

PR #2246, opened 2020-07-23
[SW-2377] Double Usage of Parenthesis in H2OMOJOBase.py

mn-mikke:


jakubhava:
I hope it's ok if I merge this one so I can start the release

PR #2248, opened 2020-07-24
[SW-2381] Deprecate removed XGBoost options

jakubhava:
H2O has removed ``minSumHessianInLeaf`` and ``minDataInLeaf`` and our build on rel-3.30.1 started to fail as the we still refer to them

### DIFF
--- a/py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
+++ b/py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
@@ -15,14 +15,13 @@
 # limitations under the License.
 #
 
-from pyspark.ml.param import *
-
 from ai.h2o.sparkling.ml.params.H2OAlgoSupervisedParams import H2OAlgoSupervisedParams
 from ai.h2o.sparkling.ml.params.H2OTreeBasedSupervisedMOJOParams import H2OTreeBasedSupervisedMOJOParams
 from ai.h2o.sparkling.ml.params.H2OTypeConverters import H2OTypeConverters
 from ai.h2o.sparkling.ml.params.HasMonotoneConstraints import HasMonotoneConstraints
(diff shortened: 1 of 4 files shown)


py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
```diff
@@ -309,10 +296,12 @@ def getMaxLeaves(self):
         return self.getOrDefault(self.maxLeaves)
 
     def getMinSumHessianInLeaf(self):
-        return self.getOrDefault(self.minSumHessianInLeaf)
+        warnings.warn("The method 'getMinSumHessianInLeaf' is deprecated without replacement and will be removed in 3.32.")
+        return 100
```
mn-mikke, line 300: nit: The method used to return a float value (`100.0`).

py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
```diff
@@ -309,10 +296,12 @@ def getMaxLeaves(self):
...
 
     def getMinDataInLeaf(self):
-        return self.getOrDefault(self.minDataInLeaf)
+        warnings.warn("The method 'getMinDataInLeaf' is deprecated without replacement and will be removed in 3.32.")
+        return 0
```
mn-mikke, line 304: nit: The method used to return a float value (`0.0`).

py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
jakubhava: Thanks!

py/src/ai/h2o/sparkling/ml/params/H2OXGBoostParams.py
jakubhava: Thanks!

PR #2251, opened 2020-07-24
[SW-1847] Automatically Document Sparkling Water Algo API

mn-mikke:


doc/src/site/sphinx/ml/sw_xgboost.rst
```diff
@@ -1,8 +1,8 @@
 Train XGBoost Model in Sparkling Water
 --------------------------------------
 
-Sparkling Water provides API for H2O XGBoost in Scala and Python.
-The following sections describe how to train the XGBoost model in Sparkling Water in both languages.
+Sparkling Water provides API for H2O XGBoost in Scala and Python. The following sections describe how to train
+theXGBoost model in Sparkling Water in both languages. See also :ref:`parameters_H2OXGBoost`.
```
jakubhava, line 5: nit: missing space after `the`

doc/src/site/sphinx/ml/sw_target_encoder.rst
```diff
@@ -32,55 +32,6 @@ An example of converting a categorical feature to continues with Target Encoder
 Target Encoding can help to improve accuracy of machine learning algorithms when columns with high
 cardinality are used as features during a training phase.
 
-Parameters
```
jakubhava, line 35: What  do you think of adding link to generated parameters doc?

doc/src/site/sphinx/ml/sw_kmeans.rst
```diff
@@ -102,6 +102,7 @@ The following sections describe how to train the KMeans model in Sparkling Water
 
 H2O KMeans Parameters
 ~~~~~~~~~~~~~~~~~~~~~
+See also :ref:`parameters_H2OKmeans`.
 
 - **maxIterations**
```
jakubhava, line 107: Should we keep the parameters listed bellow?

doc/src/site/sphinx/ml/sw_xgboost.rst
mn-mikke: will fix, thanks!

doc/src/site/sphinx/ml/sw_target_encoder.rst
mn-mikke: I wanted to do it, but forgot about that. Good catch, thanks!

doc/src/site/sphinx/ml/sw_kmeans.rst
mn-mikke: I was also thinking about that but I'm not sure whether it's better remove it or not. The parameter descriptions comes from H2O-3 V3 classes and sometimes give very little pieces of information. On the other hand, the link contains a complete list of parameters.

I will do whatever you prefer...

jakubhava: Let's keep it. Small duplication is better then missing documentation. WDYT?

PR #2261, opened 2020-07-27
[SW-2393]Doc: Multinode Xgboost is no longer experimental in AutoML

jakubhava:


sebhrusen:
@mn-mikke confirmed!
Code being source of truth:
```
      // on single node, XGBoost is enabled by default if the extension is enabled.
      // on multinode, the same condition applies, but only on Linux by default: needs to be activated explicitly for other platforms.
      boolean enabledOnMultinode = Boolean.parseBoolean(System.getProperty(DISTRIBUTED_XGBOOST_ENABLED, isLinux() ? "true" : "false"));
```
Which means that for Spark, it's enabled by default.

mn-mikke:
Thanks @sebhrusen!!!

PR #2273, opened 2020-08-05
[SW-1366] Be able to annotate jira with flag indicating the change should not be mentioned in release notes

jakubhava:
Verified and the generated release notes have the same format as before. 

We are now able to mark a JIRA by a special tag which ensures that it will not be included in the release notes. This may be helpful for some technical tasks which do not have any effect on users.

Screenshot 2020-08-05 at 12 34 15

By not setting the field, the JIRA will be included in the release notes, so the default behaviour is not changing

### DIFF
--- a/ci/Jenkinsfile-release
+++ b/ci/Jenkinsfile-release
@@ -1,5 +1,8 @@
 #!/usr/bin/groovy
 @Library('test-shared-library') _
+import groovy.json.JsonSlurperClassic
+import java.util.Date
+import java.text.SimpleDateFormat
 
 // Job parameters
 properties(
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile-release
```diff
@@ -157,10 +179,10 @@ def prepareReleaseNotes(sparkMajorVersions, commons) {
             git fetch --no-tags
             git checkout master
             git pull
-            git checkout -b "master-changelog-${versionNoSparkSuffix}"
+            git checkout -b "master-changelog-${versionNoSparkSuffix}-\${LAST_COMMIT}"
```
jakubhava, line 182: To be able to generate the release notes on the same machine, we need to have unique branch name

PR #2276, opened 2020-08-05
[SW-2403] Replace deprecated org.scalatest.junit.JUnitRunner with org.scalatestplus.junit.JUnitRunner

mn-mikke:


mn-mikke:
You're right, it works only on 2.12

PR #2278, opened 2020-08-06
[SW-2406] Enable 'detailed_prediction' Column for MOJO Predictions by Default

mn-mikke:
Users are very often confused how to enable content of the detailed prediction column and get access to probabilities. This property proposed to deprecate the `withDetailedPredictionCol` on `H2OMOJOSettings` always produce the detailed prediction column for MOJO predictions. 

This doesn't have any performance impact since content of the detailed predictions column is calculated even if the column is disabled.

Users also would have to set only one option to enable contributions, leaf node assignment, stage probabilities, etc.

If users are not interested in the detailed prediction column, they can simply call drop on Spark DataFrame.


booklet/src/sections/productionazing.tex
```diff
@@ -181,10 +180,9 @@ \subsection{Customizing the MOJO Settings}
 \begin{itemize}
     \item \texttt{predictionCol} - Specifies the name of the generated prediction column. Default value is \texttt{prediction}.
     \item \texttt{detailedPredictionCol} - Specifies the name of the generated detailed prediction column. The detailed prediction column, if enabled, contains additional details, such as probabilities, shapley values etc. The default value is \texttt{detailed\_prediction}.
-    \item \texttt{withDetailedPredictionCol} - Enables or disables the generation of the detailed prediction column. By default, it is disabled.
```
mn-mikke, line 184: @jakubhava Are you talking about this file? I thought that I remove all occurences of `withDetailedPredictionCol`, but maybe I missed something...

jakubhava: Ah, sorry, my bad. I didn't mention on the first look that you already updated this file. All good then!

py/tests/unit/with_runtime_sparkling/test_drf.py
```diff
@@ -60,7 +60,8 @@ def testCalibrationDataFrameCauseGenerationOfCalibratedProbabilities(prostateDat
                   seed=1,
                   splitRatio=0.8,
                   calibrateModel=True,
-                  calibrationDataFrame=calibrationDataset)
+                  calibrationDataFrame=calibrationDataset,
+                  withDetailedPredictionCol=True)
```
jakubhava, line 64: qq: should we completely remove this parameter since it is not used?

mn-mikke: I've added this parameter back to test it doesn't cause any braking change. By manual testing, I've noticed that init method throws exception if `withDetailedPredictionCol=True` is there. I will remove the parameter completely in 3.34.

PR #2279, opened 2020-08-06
[SW-2030] Add H2O PCA to Algo API

jakubhava:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/ProblemType.scala
```diff
@@ -27,7 +27,7 @@ object ProblemType extends Enumeration {
   def familyToProblemType(family: String): ProblemType = {
     val enumValue = Family.valueOf(family)
     enumValue match {
-      case Family.AUTO => Both
+      //case Family.AUTO => Both
```
mn-mikke: The options must be uncommented

py/tests/unit/with_runtime_sparkling/test_pca.py
```diff
@@ -0,0 +1,58 @@
...
+                  pcaMethod="GramSVD",
+                  seed=42,
+                  transform="standardize")
+
+def testPipelineSerialization(arrestsDataset):
+    [traningDataset, testingDataset] = arrestsDataset.randomSplit([0.9, 0.1], 42)
+    algo = getPreconfiguredAlgorithm()
+
+    pipeline = Pipeline(stages=[algo])
+    pipeline.write().overwrite().save("file://" + os.path.abspath("build/glrm_pipeline"))
```
mn-mikke, line 49: Can you rename "build/glrm_pipeline" so it doesn't conflict with test for GLRM?

py/tests/unit/with_runtime_sparkling/test_pca.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def testParamsPassedByConstructor():
+    assertParamsViaConstructor("H2OPCA")
+
+
+def testParamsPassedBySetters():
+    assertParamsViaSetters("H2OPCA")
+
+
```
mn-mikke, line 34: nit: too many empty lines

py/tests/unit/with_runtime_sparkling/test_pca.py
```diff
@@ -0,0 +1,58 @@
...
+def getPreconfiguredAlgorithm():
+    return H2OPCA(k=4,
+                  maxIterations=500,
+                  pcaMethod="GramSVD",
+                  seed=42,
+                  transform="standardize")
+
+def testPipelineSerialization(arrestsDataset):
```
mn-mikke, line 44: Can you add some test making assertion on a real values. (For instance to check that predictions are not the same for all rows in the dataset)?

py/tests/unit/with_runtime_sparkling/test_pca.py
jakubhava: 👍

py/tests/unit/with_runtime_sparkling/test_pca.py
jakubhava: 👍

py/tests/unit/with_runtime_sparkling/test_pca.py
jakubhava: Will do, thanks!

jakubhava:
@mn-mikke please have a look. @wendycwong suggested that we can use sum of variances as a metric. There has been discussion that it is not ideal, but I would like to support PCA in grid at least like this, we can always improve in the future.

doc/src/site/sphinx/ml/sw_pca.rst
```diff
@@ -0,0 +1,108 @@
...
+
+            import ai.h2o.sparkling._
+            import java.net.URI
+            val hc = H2OContext.getOrCreate()
+
+        Parse the data using H2O and convert them to Spark Frame
+
+        .. code:: scala
+
+	        import org.apache.spark.SparkFiles
```
mn-mikke, line 33: nit: Is the indentation ok, or github displays that incorrectly?

PR #2284, opened 2020-08-10
[SW-2408] Add H2O Isolation Forest to Algo API

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/algos/ProblemType.scala
```diff
@@ -27,7 +27,7 @@ object ProblemType extends Enumeration {
   def familyToProblemType(family: String): ProblemType = {
     val enumValue = Family.valueOf(family)
     enumValue match {
-      case Family.AUTO => Both
+//      case Family.AUTO => Both
```
jakubhava, line 30: Should be uncommented

r/src/R/ai/h2o/sparkling/ml/models/H2OMOJOModel.R
```diff
@@ -79,3 +79,10 @@ H2OTreeBasedSupervisedMOJOModel <- setRefClass("H2OTreeBasedSupervisedMOJOModel"
 #' @export H2OUnsupervisedMOJOModel
 H2OUnsupervisedMOJOModel <- setRefClass("H2OUnupervisedMOJOModel", contains = ("H2OMOJOModel"), fields = list(jmojo = "ANY"), methods = list(
 ))
+
+#' @export H2OTreeBasedUnsupervisedMOJOModel
+H2OTreeBasedUnsupervisedMOJOModel <- setRefClass("H2OTreeBasedUnsupervisedMOJOModel", contains = ("H2OUnsupervisedMOJOModel"), methods = list(
+  getNtrees = function() {
+    invoke(.self$jmojo, "getNtrees")
+  }
+))
```
jakubhava, line 88: nit: nl

ml/src/main/scala/ai/h2o/sparkling/ml/algos/ProblemType.scala
mn-mikke: Thanks!

r/src/R/ai/h2o/sparkling/ml/models/H2OMOJOModel.R
mn-mikke: Thanks!

PR #2298, opened 2020-08-25
[SW-2413] Upgrade MOJO Runtime to 2.4.8

mn-mikke:


jakubhava (approved):
LGTM from the code point of view. Not sure if mojo hasn't changed other APIs as well

mn-mikke:
>  Not sure if mojo hasn't changed other APIs as well
@jakubhava What do you mean by other APIs?

jakubhava:
Just mojo API -> what I meant is that if it complies, we are probably fine...

mn-mikke:
I see. I want to check it with @mmalohlava or @pkozelka that the upgrade won't break anything.

pkozelka (approved):

PR #2299, opened 2020-08-25
[SW-2414] Make Documentation More Descriptive about Extraction of pipeline.mojo from mojo.zip

mn-mikke:


jakubhava (approved):
Good change!

PR #2300, opened 2020-08-26
[SW-2416] Kubernetes tests should clean up the environment in case of an error

jakubhava:
Kubernetes tests keep the resources on-line in case some tests fails


mn-mikke (approved):
LGTM

PR #2303, opened 2020-08-27
[SW-2417] Clean up w2v tokenizer and expose minTokenLength and pattern to provide same features as H2O tokenize method

jakubhava:
-> No need for the marker
-> Expose minTokenLength
-> Expose pattern


ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OWord2VecTokenizer.scala
```diff
@@ -39,28 +39,30 @@ class H2OWord2VecTokenizer(override val uid: String)
...
+      .setPattern(getPattern())
 
     val stopWordsRemover = new StopWordsRemover()
       .setInputCol(tokenizer.getOutputCol)
       .setStopWords(getStopWords())
       .setOutputCol(s"stop_words_removed_$uid")
 
+    val tokenized = tokenizer
+      .transform(dataset)
+      .withColumn(tokenizer.getOutputCol, array_union(col(tokenizer.getOutputCol), lit(Array(""))))
```
mn-mikke, line 60: AFAIK `array_union` was introduced in Spark 2.4. I'm afraid this won't work in older Spark versions.

jakubhava: Thanks!

jakubhava: Should be fixed, added custom UDF which does the same thing, tested on Spark 2.1 as well

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OWord2VecTokenizerParams.scala
```diff
@@ -16,32 +16,40 @@
...
+import org.apache.spark.ml.param.{IntParam, Param, Params, StringArrayParam}
 
 trait H2OWord2VecTokenizerParams extends Params {
 
   //
   // Param definitions
   //
   private final val stopWords = new StringArrayParam(this, "stopWords", "List of stop words.")
-  private final val inputCol = new Param[String](this, "inputCol", "input column name")
+  private final val inputCol = new NullableStringParam(this, "inputCol", "input column name")
```
mn-mikke, line 27: Does it make sense to allow `null` values on `inputCol`? if `inputCol` is null what is the subject of tokenization?

mn-mikke (approved):
LGTM

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OWord2VecTokenizerParams.scala
jakubhava: If we don't allow nulls, then we need to set input column to some default value which I wated to avoid

jakubhava: But happy to change it if you think this is not a good pattern

mn-mikke: IMHO, we should follow as it's in Spark mllib. People are used to the Spark's behavior if the input column is not set. I'm afraid `NullPointerException` or similar error might be a bit misleading.

jakubhava: done

PR #2307, opened 2020-09-03
[SW-2421] Improve Documentation for XGBoost Memory Requirements

mn-mikke:


jakubhava (approved):
Perfect change! Brings more clarity. Thank you!

doc/src/site/sphinx/ml/sw_xgboost.rst
```diff
@@ -118,18 +118,23 @@ XGBoost Memory Configuration
...
-In Spark Standalone Mode or IBM Conductor environment, please make sure to configure the following configurations:
+Note: In case of Pysparkling, the YARN container will be bigger about the memory required by the Python process.
 
 
-- ``spark.memory.offHeap.enabled=true``
-- ``spark.memory.offHeap.size=4G`` - example of setting this property to 4G of off-heap memory
\ No newline at end of file
+Memory Overhead on Spark driver
+```````````````````````````````
+If you set the SW property ``spark.ext.h2o.rest.api.based.client`` to ``false``, you should also set the following properties
```
michalkurka, line 136: Does this refer to the "H2O client"?

doc/src/site/sphinx/ml/sw_xgboost.rst
```diff
@@ -118,18 +118,23 @@ XGBoost Memory Configuration
...
+```````
+If you set ``spark.executor.memory`` to ``10g``, ``SUBST_EXECUTOR_MEMORY_OVERHEAD`` should be set to ``12288``.
+The size of the corresponding YARN or Kubernetes container will be at least 22 GiB.
 
-In Spark Standalone Mode or IBM Conductor environment, please make sure to configure the following configurations:
+Note: In case of Pysparkling, the YARN container will be bigger about the memory required by the Python process.
 
 
-- ``spark.memory.offHeap.enabled=true``
-- ``spark.memory.offHeap.size=4G`` - example of setting this property to 4G of off-heap memory
```
mn-mikke, line 135: @jakubhava IMHO, this option has a different meaning. This option limits Spark itself when doing manual memory allocation to the offheap space. E.g.: Tungsten memory representation, `df.persist(persist(StorageLevel.OFF_HEAP)`

Wouldn't ``spark.executor.memoryOverhead`` work for Spectrum Conductor as well.  I don't see it in [the list of unsupported parameters](https://www.ibm.com/support/knowledgecenter/en/SSZU2E_2.4.1/reference_s/spark_properties_unsupported.html). 

@FengAtH2O what option do you for XGBoost and specifying  the memoryOverhead (off-heap) space on Spectrum Conductor?

doc/src/site/sphinx/ml/sw_xgboost.rst
mn-mikke: yes, you're right. By default, H2O client is disabled since the version `3.30.0.1-1`.

michalkurka: ok, I think I get it, please confirm that my understanding is correct:

Setting `spark.ext.h2o.rest.api.based.client` to `false` enables the "H2O client mode" and you thus need set overhead also for the process where the client code is running, right?

jakubhava: Just clarification -> `spark.ext.h2o.rest.api.based.client` is valid only when running Sparkling Water in Scala/Java. There is no option to enable H2O client in R/Python, no matter what this option is set to. Python/R is always using Rest client

mn-mikke: @jakubhava What about SW Python API it's backed with Scala objects. the same holds for `H2OContext`.

jakubhava: Yup, it is backed by Scala objects, but no H2O client running. There is no H2O client on Py/R at all

mn-mikke: ok, i see the property `spark.ext.h2o.rest.api.based.client` manually hardcoded to "true"

mn-mikke: Now, I'm wondering if the memory overhead is needed on the driver at all. If I'm not mistaken all the Rest API requests goes to the H2O client if it's enabled. Otherwise, it goes to the leader node. 

@michalkurka As far I remember, XGboost model building was performed on the H2O client only when a specific Java method for triggering the model training was called. I remember at some point that this could be omitted by calling a different one. Now all commands go through Rest API (`/3/ModelBuilders/`). Do you think that H2O client still will be part of model training?

michalkurka: The client will definitely need some native memory if invoked from Java/Scala. If REST API is used, it will be forwarded to an H2O node instead of the client.

The client is never part of the training, but does access the native memory to retrieve the booster.

mn-mikke: I see, thank you both! I will add a note that this is relevant only the scala API.

honzasterba (commented):
lgtm

honzasterba (approved):

michalkurka (approved):

PR #2309, opened 2020-09-04
[SW-2423] Remove Irrelevant Parameters from Kmeans API

mn-mikke:
I have also a question about two parameters which are not included in the [MOJO json file](https://gist.github.com/mn-mikke/8177b9503bcffb124131c5b8e15b5b46#file-kmeans-model-json). The parameters are:
- `max_categorical_levels` - This parameter is mentioned in [H2O-3 documentation](http://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/k-means.html). So shouldn't be included in MOJO as well?
- `weights_column` - Is there any plan to implement this parameter on Kmeans. Or it doesn't make sence for KMeans at all?

### DIFF
--- a/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
+++ b/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
@@ -37,10 +37,25 @@ object IgnoredParameters {
 
   def common: Seq[String] = deprecated ++ implementedInParent ++ unimplemented
 
-  def all(algorithm: String): Seq[String] = algorithm match {
-    case "H2OKMeans" => common ++ Seq("response_column", "offset_column")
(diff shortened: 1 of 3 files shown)


maurever (approved):
Looks good. These parameters are not used in KMeans. However, I don't know from which list do you get these parameters, so I am not sure if any is missing. :)

mn-mikke:
@maurever Thanks a lot for the review! The complete list of parameters is made of fields on V3 classes. Since some base classes (`ModelParametersSchemaV3`) contains some parameter which do not make sense for particular algorithms, these parameters have to be put on the ignored list.

maurever:
Thank you, @mn-mikke, for the explanation! I also read the description of this PR. :facepalm:  

With `max_categorical_levels ` is the same case as with GBM.

The `weights_column` is not supported, and there is no planned JIRA for it. It is not a trivial task, I guess, to implemented it.

PR #2310, opened 2020-09-04
[SW-2424] Remove Irrelevant Parameters from DeepLearning API

mn-mikke:


mn-mikke:
This PR is not needed thank to https://github.com/h2oai/sparkling-water/pull/2317.

PR #2311, opened 2020-09-04
[SW-2425] Remove max_categorical_levels from GBM API

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/IgnoredParameters.scala
```diff
@@ -37,10 +37,13 @@ object IgnoredParameters {
...
   def common: Seq[String] = deprecated ++ implementedInParent ++ unimplemented
 
-  def all(algorithm: String): Seq[String] = algorithm match {
-    case "H2OKMeans" => common ++ Seq("response_column", "offset_column")
-    case "H2OGAM" => common ++ Seq("plug_values") // According to MK the parameter doesn't make much sense for GAM
-    case "H2ODeepLearning" => common ++ Seq("pretrained_autoencoder")
-    case _ => common
+  def all(algorithm: String): Seq[String] = common ++ {
+    algorithm match {
+      case "H2OGBM" => Seq("max_categorical_levels")
```
maurever, line 42: Why is it removed for GBM? I think it is a valid parameter for all tree algos. I checked the GBM documentation, and the parameter is missing there; however, it is probably a bug. Or am I missing something?

mn-mikke: I can't find this parameter in [MOJO json file](https://gist.github.com/mn-mikke/8177b9503bcffb124131c5b8e15b5b46#file-gbm-model-json) so I thought the parameter is not related to the algorithm. If think that the parameter should be there, I can close this PR, but I need help why it's not propagated to MOJO.

maurever: Maybe the reason why it is missing in your MOJO is that this parameter is relevant only if `categorical_encoding = "EnumLimited" `.

mn-mikke: Will try, thanks!

mn-mikke: @maurever I've tried to set [`categorical_encoding` to `EnumLimited`](https://gist.github.com/mn-mikke/8177b9503bcffb124131c5b8e15b5b46#file-gbm-model-enumlimited-json-L1170),  but `max_categorical_levels` is still missing.

maurever: So, it looks like a bug. Please, make a JIRA for it and asign it to me. Thank you @mn-mikke!

mn-mikke: @maurever I've created [this JIRA ticket](https://0xdata.atlassian.net/browse/PUBDEV-7768) for the problem. Thanks a lot! Closing the PR...

mn-mikke:
This problem will be handled in https://0xdata.atlassian.net/browse/PUBDEV-7768

PR #2312, opened 2020-09-04
[SW-2427] Fix parameter generation in doc

jakubhava:
H2OWord2VecTokenizer has no default value for input column and documentation generation fails.

H2OWord2VecTokenizer will be removed as part of the Word2Vec but before that, this commit ensures the CI is green


mn-mikke (approved):
LGTM, thanks!

PR #2314, opened 2020-09-05
[SW-2429] Update H2OTargetEncoder According to Changes in PUBDEV-7702

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -50,12 +50,14 @@ class H2OTargetEncoder(override val uid: String)
...
+      "data_leakage_handling" -> getHoldoutStrategy(),
       "blending" -> getBlendedAvgEnabled(),
-      "k" -> getBlendedAvgInflectionPoint(),
-      "f" -> getBlendedAvgSmoothing(),
+      "inflection_point" -> getBlendedAvgInflectionPoint(),
+      "smoothing" -> getBlendedAvgSmoothing(),
       "response_column" -> getLabelCol(),
       "fold_column" -> getFoldCol(),
       "ignored_columns" -> ignoredColumns,
+      "seed" -> getNoiseSeed(),
```
sebhrusen, line 60: doesn't SW expose the `noise` param?
If not, it means that noise is always enabled (not sure this is a good default, but this is the one that was chosen originally...), and the amount of noise is computed from the range of values in the target.
I think you should at least allow users to disable this noise: in other clients, it's done by setting `noise=0`, and they can otherwise specify that amount of noise, e.g. `noise=0.5` means that for each row the encoded value with be altered by some random noise comprised between -0.5 and +0.5.

ml/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderModel.scala
```diff
@@ -69,12 +69,11 @@ class H2OTargetEncoderModel(override val uid: String, targetEncoderModel: H2OMod
     val params = Map(
       "model" -> targetEncoderModel.modelId,
       "frame" -> input.frameId,
-      "data_leakage_handling" -> getHoldoutStrategy(),
       "noise" -> getNoise(),
```
sebhrusen, line 72: I see, `noise` is provided only at `transform` level, which is where it is needed technically.
Although providing it at training level allows to define a default noise value that will then be used for all subsequent calls to `transform`: this may be more convenient for some scenarios.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMOJOModel.scala
```diff
@@ -42,7 +42,7 @@ class H2OTargetEncoderMOJOModel(override val uid: String)
 
   @transient private lazy val orderOfInputColumns = {
     val mojoModel = Utils.getMojoModel(getMojo()).asInstanceOf[TargetEncoderMojoModel]
-    val indexes = mojoModel._teColumnNameToIdx
+    val indexes = mojoModel._columnNameToIdx
```
sebhrusen, line 45: interesting, I didn't think that users would access those Mojo properties directly.
What do you need this one for exactly? To know the order of the predictions columns?

sebhrusen (approved):
looks good, thanks for doing this quickly @mn-mikke

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OTargetEncoderMOJOModel.scala
mn-mikke: For the order of input columns. SW user provides the list of `inputColumns` in a arbitrary order. And in the same order, the result should be return. I remember TE MOJO was not returning predictions in an expected order (training order). `_teColumnNameToIdx` was a solution for ordering in [the SW case](https://github.com/h2oai/sparkling-water/pull/2314/files/2350260b812f11de283547bbd219cfff8b0c9b87#diff-cec96bfbb0969ede0661e2aaee3d9915R54).

sebhrusen: Ok, FYI the predictions from the Mojo should always follow the column order by index.
Maybe in a previous version, this was not the case, but it's been the case even before my refactoring.

PR #2317, opened 2020-09-14
[SW-2432] Limit Generated Parameters Only to Parameters in *ParameteV3.fields

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/params/HasDistribution.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ai.h2o.sparkling.ml.params
+
+trait HasDistribution {
+  def getDistribution(): String = "AUTO"
```
mn-mikke, line 21: This needs to be deprecated in a separate PR. `getDistribution` and `setDistribution` have been added to make the code compilable.

PR #2319, opened 2020-09-17
[SW-2439] Missing Import of H2OBinaryModel in Python Classification and Regression Classes

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/AlgorithmTemplateBase.scala
```diff
@@ -50,10 +50,7 @@ trait AlgorithmTemplateBase extends PythonEntityTemplate {
          |        self._setDefaultValuesFromJava()
          |        kwargs = Utils.getInputKwargs(self)
          |        self._set(**kwargs)
-         |        self._transfer_params_to_java()
-         |
-         |    def getBinaryModel(self):
```
mn-mikke, line 55: IMHO, static code shouldn't be generated if possible. Code generation ads complexity at the expense of code readability

PR #2321, opened 2020-09-17
[SW-2440] Generate Algorithm-specific Python MOJO Classes

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/MOJOModelTemplate.scala
```diff
@@ -0,0 +1,57 @@
...
+    generateEntity(entitySubstitutionContext) {
+      generateGetterMethods(parameters)
+    }
+  }
+
+  private def generateGetterMethods(parameters: Seq[Parameter]): String = {
+    parameters.map { parameter =>
+      val parameterName = parameter.swName.capitalize
+      s"""
+         |    def get$parameterName(self):
```
jakubhava, line 53: small note - get${parameterName} is easier to read at least for me

jakubhava (approved):
looks good

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/MOJOModelTemplate.scala
mn-mikke: Will update, thanks!

PR #2324, opened 2020-09-21
[SW-2443] Remove H2OWord2VecTokenizer

jakubhava:
No need to deprecate as it was introduced in the same release.


mn-mikke (approved):
LGTM!

PR #2327, opened 2020-09-23
[SW-2434] Deprecate 'weightCol' Parameter on H2OKmeans

mn-mikke:
The parameter is not supported by the implementation of KMeans in H2O-3 and won't be implemented in near future.


jakubhava (approved):
looks good, thanks

PR #2328, opened 2020-09-23
[SW-2436] Deprecate 'trainingParameters' Method on H2OMOJOModel

mn-mikke:


jakubhava (commented):
getTrainingParams is exposed also in H2OMOJOModel in RSparkling -> https://github.com/h2oai/sparkling-water/blob/57bfa19651abffd8f2a17af103aa5412c11b1820/r/src/R/ai/h2o/sparkling/ml/models/H2OMOJOModel.R#L59 We should deprecate it there as well.

small note: Did you please create a JIRA for removal in 3.34 backlog?

mn-mikke:
@jakubhava I would maybe prefer to keep this method in R API since we don't any replacement for it. I think we can remove it once we generate R API.

jakubhava:
That's a valid point, thanks 👍

jakubhava (approved):

PR #2332, opened 2020-09-25
[SW-2435] Use `family` Parameter on H2OGLM, H2OGAM for Determining a Need to Convert the Label Column to Categoricals

mn-mikke:
GLM in Sparkling Water used use the `distribution` parameter for this purpose. Unfortunately, `distribution` is not a parameter of GLM or GAM.


honzasterba (approved):

mn-mikke:
Thanks @honzasterba for your review!

PR #2341, opened 2020-10-02
[SW-2449] asH2OFrame Method Could Fail on a String Column Having More Than 10 Million Distinct Values

mn-mikke:
This PR converts all Spark string columns to H2O-3 categorical columns as a first step. The H2O backend then converts  the columns back to string if they are identified as too unique or if the limit for maximum categorical levels is exceeded.


core/src/main/scala/ai/h2o/sparkling/backend/converters/DataTypeConverter.scala
```diff
@@ -21,37 +21,69 @@ import ai.h2o.sparkling.backend.utils.SupportedTypes
...
+    val categoricalColumns = stringColumns.zip(expectedTypes).filter(_._2 == ExpectedTypes.Categorical).map(_._1)
+
+    if (categoricalColumns.nonEmpty) {
+      val (categoricalNames, categoricalIndices) = categoricalColumns.unzip
+
+      val allowedError = 0.05 // Spark default on approx_count_distinct
+
+      // Adding margin 0.01 just to be sure that the real distinct count won't be under Categorical.MAX_CATEGORICAL_COUNT
+      // and the approximate count won't be above threshold.
+      val threshold = Categorical.MAX_CATEGORICAL_COUNT * (1 + allowedError + 0.01)
```
honzasterba, line 70: since the count is approximate should't you lower the threshold to make sure the count fits into the threshold?

mn-mikke: IMHO, the threshold should be higher than `Categorical.MAX_CATEGORICAL_COUNT` if we want to get a similar bahavior as in H2O.  If the true distinct count is exactly `Categorical.MAX_CATEGORICAL_COUNT` the function can return `approx_count_distinct` can return a higher number. With a lower threshold, we would lose some categorical columns. On the other hand, categorical columns with cardinality close to  `Categorical.MAX_CATEGORICAL_COUNT` might not to make sense anaway.

Or did you mean it from the point of view of avoiding the exception? (long strings and high cardinality)?

honzasterba: my point was making 100% sure to avoid breking the limit, the limit itself is pretty much theretical and I don't think you want to come close it

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
```diff
@@ -41,28 +42,83 @@ class ImportFrameHandler extends Handler {
...
       }
     }
+  }
 
-    request
+  private def convertUniqueCategoricalColumnsToStringColumns(frame: Frame, categoricalColumnIndices: Array[Int]) = {
+    categoricalColumnIndices.foreach { idx =>
+      val vector = frame.vec(idx)
+      val ratio = vector.cardinality() / vector.length().asInstanceOf[Double]
+      if (ratio > 0.95) {
```
mn-mikke, line 96: The heuristic logic utilized by H2O parsers (`PreviewParseWriter`):
```
  @Override public void addStrCol(int colIdx, BufferedString str) {
    if(colIdx < _ncols) {
      // Check for time
      if (ParseTime.isTime(str)) {
        ++_ndates[colIdx];
        return;
      }

      //Check for UUID
      if(ParseUUID.isUUID(str)) {
        ++_nUUID[colIdx];
        return;
      }

      //Add string to domains list for later determining string, NA, or categorical
      ++_nstrings[colIdx];
      _domains[colIdx].put(str.toString(),"");

      if (_nlines < MAX_PREVIEW_LINES)
        _data[_nlines][colIdx] = str.toString();
    }
  }
```

```
      // UUID
      if (_nUUID[i] > _ndates[i]
              && _nUUID[i] > (_nnums[i] + _nzeros[i])
              && (_nUUID[i] > _nstrings[i] || _domains[i].size() <= 1)) {
        types[i] = Vec.T_UUID;
        continue;
      }

      // Strings, almost no dups
      if (_nstrings[i] > _ndates[i]
              && _nstrings[i] > _nUUID[i]
              && _nstrings[i] > (_nnums[i] + _nzeros[i])
              && _domains[i].size() >= 0.95 * _nstrings[i]) {
        types[i] = Vec.T_STR;
        continue;
      }

      // categorical or string?
      // categorical with 0s for NAs
      if(_nzeros[i] > 0
              && ((_nzeros[i] + _nstrings[i]) >= nonemptyLines) //just strings and zeros for NA (thus no empty lines)
              && (_domains[i].size() <= 0.95 * _nstrings[i]) ) { // not all unique strings
        types[i] = Vec.T_CAT;
        continue;
      }
      // categorical mixed with numbers
      if(_nstrings[i] >= (_nnums[i]+_nzeros[i]) // mostly strings
              && (_domains[i].size() <= 0.95 * _nstrings[i]) ) { // but not all unique
        types[i] = Vec.T_CAT;
        continue;
      }
```

I'm not sure whether it's necessary to cover also `Vec.T_TIME` and `Vec.T_UUID`. `Vec.T_TIME` is usually converted from Spark's `DateType` or `TimestampType`.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
```diff
@@ -41,28 +42,83 @@ class ImportFrameHandler extends Handler {
...
     updateCategoricalIndicesTask.doAll(frame)
 
-    // Convert categorical columns with too many categories to T_STR
-    categoricalColumnIndices.foreach { catColIdx =>
-      val vector = frame.vec(catColIdx)
-      if (vector.cardinality() > Categorical.MAX_CATEGORICAL_COUNT) {
-        Log.warn(
-          s"Categorical column with index '$catColIdx' exceeded maximum number of categories. " +
+    // Convert unique categorical columns to T_STR
+    convertUniqueCategoricalColumnsToStringColumns(frame, categoricalColumnIndices)
```
michalkurka, line 64: I think it might be a good idea to keep this logic on the H2O side - close to the PreviewParseWriter - then we can make sure the logic is actually consistent.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
```diff
@@ -41,28 +42,83 @@ class ImportFrameHandler extends Handler {
...
+      columnTypes.map(columnType => if (columnType == Vec.T_CAT) Vec.T_STR else columnType)
+    } else {
+      columnTypes
+    }
+    val domains = expandDomains(stringDomains, columnTypesAwareOfEmptyFrames)
+    ChunkUtils.finalizeFrame(request.key, rowsPerChunk, columnTypesAwareOfEmptyFrames, domains)
+    val frame = DKV.getGet[Frame](frameKey)
 
-    ChunkUtils.finalizeFrame(request.key, rowsPerChunk, columnTypes, domains)
+    convertColumnsWithTooManyCategoricalLevelsToStringColumns(frame, columnTypesAwareOfEmptyFrames, stringDomains)
```
michalkurka, line 55: this is a part I don't understand - the frame is already in DKV and has "too many" categoricals - should thee frame not be in DKV yet?

is there any locking in place, the ParseDataset locks the frame against reading

core/src/test/scala/ai/h2o/sparkling/backend/converters/DataFrameConverterCategoricalTestSuite.scala
```diff
@@ -0,0 +1,96 @@
...
+
+    assertH2OFrameInvariants(df, h2oFrame)
+    assert(h2oFrame.columns(0).isCategorical())
+
+    val resultDF = hc.asSparkFrame(h2oFrame)
+    TestUtils.assertDataFramesAreIdentical(df, resultDF)
+    h2oFrame.delete()
+  }
+
+  test("DataFrame[String] with more than 10M unique values to H2OFrame[T_STR] and back") {
```
michalkurka, line 76: would it make sense to also add a test that has all the categorical values in a single partition?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
mn-mikke: The frame is in DKV from the beginning of the conversion. The initialize method puts the Frame into DKV (`ChunkUtils.initFrame`) as partial frame: 
```
    public static void initFrame(String keyName, String[] names) {
        Frame fr = new water.fvec.Frame(Key.<Frame>make(keyName));
        fr.preparePartialFrame(names);
        // Save it directly to DKV
        fr.update();
    }
```

Do you think that keeping the frame as partial and not finalizing it can avoid to read frame in a provisional state?

The columns with too many levels have null domains and during finalization are converted to `T_NUM` and then those columns are converted to `T_STR` with [ConvertCategoricalToStringColumnsTask](https://github.com/h2oai/sparkling-water/pull/2341/files/8ca5d9c1060fa02b7ce6d9edb6adf4a140bcc12b#diff-27d5892be9be15668ac386bf9b874876R27). 

I needed to finalize the frame first to get Vector instances which are [passed to ConvertCategoricalToStringColumnsTask](https://github.com/h2oai/sparkling-water/pull/2341/files/8ca5d9c1060fa02b7ce6d9edb6adf4a140bcc12b#diff-dcee268e9b2cd60b0fbd5684dc28114fR83).

Or is there any dedicated mechanism for locking frames for reading?

core/src/test/scala/ai/h2o/sparkling/backend/converters/DataFrameConverterCategoricalTestSuite.scala
mn-mikke: I will try, thanks!

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
mn-mikke: I can add the cases also for T_UUID and T_TIME. But do you mean to add a piece of code to H2O-3 repository which will utilized here?

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
mn-mikke: I've tried to use `PreviewParseWriter` where Vec.T_TIME and Vec.T_UUID are unsupported.

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
mn-mikke: the decission logic is done via `PreviewParseWriter`, but without `T_UUID` and `T_TIME`

extensions/src/main/scala/water/parser/CategoricalPreviewParseWriter.java
```diff
@@ -0,0 +1,43 @@
...
+  public CategoricalPreviewParseWriter(String[] domain, int totalCount) {
+    super(1);
+    this._nstrings[0] = totalCount;
+    IcedHashMap<String, String>[] domains = new IcedHashMap[1];
+    domains[0] = new IcedHashMapWrapper(domain);
+    setPrivateDomains(domains);
+  }
+
+  private void setPrivateDomains(IcedHashMap<String, String>[] domains) {
+    try {
```
Pscheidl, line 34: This should better not be done in parallel - non-atomic.

extensions/src/main/scala/water/parser/CategoricalPreviewParseWriter.java
```diff
@@ -0,0 +1,43 @@
...
+import java.lang.reflect.Field;
+import water.util.IcedHashMap;
+
+public class CategoricalPreviewParseWriter extends PreviewParseWriter {
+
+  public CategoricalPreviewParseWriter(String[] domain, int totalCount) {
+    super(1);
+    this._nstrings[0] = totalCount;
+    IcedHashMap<String, String>[] domains = new IcedHashMap[1];
+    domains[0] = new IcedHashMapWrapper(domain);
```
Pscheidl, line 29: The proxy/wrapper is fine for now. There is a clear reason for that - memory savings. The very fact that the original `PreviewParseWriter` uses HashMap as a set effectively doubles the already doubled memory (extreme case). 

My suggestion:

- Keep this as-is, as the internal logic of the `guessTypes` method seems to be compliant with this proxy implementation of IcedHashMap.
- In the very next H2O fix release, make changes to `PreviewParseWriter` to enabled better embedding in SW. This could also include migration to a much later introducerd IcedHashSet.
- Overridable method `getDomain()` should be introduced in the parent class as wel..

extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
Pscheidl: After looking at the code mentioned by @mn-mikke, there is no reason why the frame should not be in DKV - at least to my understanding. H2O and presumably SW is a single-user environment. Therefore any accidental reads are not possible this way.

The original ParseDataset phase is finished once the frame is finalized. And there is no ParseDataset active during the `convertColumnsWithTooManyCategoricalLevelsToStringColumns` - or is there ? 

If not, then just removing the old vec (the `replace` method of a `Frame` is called) and replacing it with a new one seems to be safe operation. Chunk layout is adapted.

This is done sequentially.

extensions/src/main/scala/water/parser/CategoricalPreviewParseWriter.java
mn-mikke: Luckily, this is used by a single thread. https://github.com/h2oai/sparkling-water/pull/2341/files#diff-c8604610ecb616a35bcf8a8b81477e5fedd946767ca51d38316559057f34037bR95

Pscheidl (approved):
The business logic makes complete sense to me. 

LGTM. @mn-mikke Let's improve the `PreviewParseWriter` on the H2O-3 side in the next iteration. We made a deal with @mn-mikke I'll help with the improvement, won't take long.

PR #2351, opened 2020-10-13
[SW-2462] org.apache.spark.h2o.H2OConf Shouldn't Override Settings from Command Line

mn-mikke:


jakubhava:
Just want to point the original motivation, feel free to use it as you wish:
R & Python clients no longer need H2O clients at all since the client-less variant is 100% compatible with the original one -> that is the reason for hardcoding the property. 

The only place in a new release where user might benefit from H2O client is in scala since the API is not fully 100% copatible. 
I wouldn't give the user the ability to switch back to client in R/Python (seems like a step back and does not brign any additional benefit) + more cases to test.

If the option is confusing, then maybe renaming to `spark.ext.h2o.scala.rest.api.based.client` or anything similar maybe makes sense.

mn-mikke:
@jakubhava

>I wouldn't give the user the ability to switch back to client in R/Python (seems like a step back and does not brign any additional benefit) + more cases to test.

This PR doesn't do any change on Python/R API. it just makes `org.apache.spark.h2o.H2OConf` behave the same way as `ai.h2o.sparkling.H2OConf`

jakubhava:
I totally misread the context, forget I said anything :D Thanks for explanation :)

mn-mikke:
As per discussion with @michalkurka, I will keep the current behavior, just will make sure that doesn't override the parameter value from command line.

michalkurka (approved):

PR #2361, opened 2020-10-20
[SW-2466] Extend Target Encoder for Regression Problems

mn-mikke:
The support for Multinomial Problems will come in a different PR. It may require changes on H2O-3 side.


sebhrusen (approved):
looks good

PR #2362, opened 2020-10-21
[SW-2467] Fix Publishing of SW Booklet

mn-mikke:
The PR https://github.com/h2oai/sparkling-water/pull/2277 moved compiled pdf from `booklet/build` to `booklet/build/src` which broke publishing of the pdf.


jakubhava (approved):
Perfect, thanks!

PR #2375, opened 2020-10-27
[SW-2470] Fails to Convert Categorical Columns on Big Dataset and Identity Column

mn-mikke:
[`UpdateCategoricalIndicesTask`](https://github.com/h2oai/sparkling-water/blob/68a8a90d4e3c9242fb5109d58be3585da2da89ea/extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala#L62) used to work with the original local domains even if the column was converted to a string column due to too exceeding the maximum of categorical levels.

### DIFF
--- a/extensions/src/main/scala/ai/h2o/sparkling/extensions/internals/UpdateCategoricalIndicesTask.java
+++ b/extensions/src/main/scala/ai/h2o/sparkling/extensions/internals/UpdateCategoricalIndicesTask.java
@@ -72,7 +72,8 @@ public class UpdateCategoricalIndicesTask extends MRTask<UpdateCategoricalIndice
   }
 
   @Override
-  public void postGlobal() {
+  public void closeLocal() {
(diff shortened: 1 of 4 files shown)


extensions/src/main/scala/ai/h2o/sparkling/extensions/internals/UpdateCategoricalIndicesTask.java
```diff
@@ -72,7 +72,8 @@ public void map(Chunk[] chunks) {
   }
 
   @Override
-  public void postGlobal() {
```
mn-mikke, line 75: This looks like a memory leak, from H2O-3 code description:
```
 * <code>postGlobal</code> is called once only after all maps, reduces and
 * closeLocals, and only on the home node.</p>
```

honzasterba (approved):

PR #2378, opened 2020-10-29
[SW-2475] Enable to Pass Extra Java options to Automatically-started External Backend

mn-mikke:
I'm going utilize this feature in tests, but I think it might be useful for users as well.

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
@@ -142,7 +142,15 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Shell
       .add("-refreshTokens", conf.isKerberizedHiveEnabled)
       .add(conf.extraProperties)
       .add(ExternalH2OBackend.getExtraHttpHeaderArgs(conf).flatMap(arg => Seq("-J", arg)))
+      .add(getExternalExtraJavaProperties(conf))
       .buildArgs()
+  }
+
+  private def getExternalExtraJavaProperties(conf: H2OConf): Seq[String] = {
+    conf.externalExtraJavaProperties match {
+      case Some(x) => x.split("\s+").flatMap(Seq("-JJ", _))
+      case None => Seq.empty[String]
+    }
   }
 
   private def getExtensionsAssemblyJar(): File = {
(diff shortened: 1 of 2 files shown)


michalkurka (approved):

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
```diff
@@ -142,9 +142,17 @@ class ExternalH2OBackend(val hc: H2OContext) extends SparklingBackend with Shell
       .add("-refreshTokens", conf.isKerberizedHiveEnabled)
       .add(conf.extraProperties)
       .add(ExternalH2OBackend.getExtraHttpHeaderArgs(conf).flatMap(arg => Seq("-J", arg)))
+      .add(getExternalExtraJavaProperties(conf))
       .buildArgs()
   }
 
+  private def getExternalExtraJavaProperties(conf: H2OConf): Seq[String] = {
+    conf.externalExtraJavaProperties match {
+      case Some(x) => x.split("\\s+").flatMap(Seq("-JJ", _))
```
michalkurka, line 151: How does a user pass a property value that has a space in it?

honzasterba (approved):

core/src/main/scala/ai/h2o/sparkling/backend/external/ExternalH2OBackend.scala
mn-mikke: That's a good question. In this implementation it's not possible. 

Can you give me example of such a Java property?

michalkurka: No property key, possibly a property value. I cannot think of any in the context of Hadoop.

On Windows, you might have paths with white-spaces and such.

You wouldn't have the issue if the user had to themselves provide the -JJ.

mn-mikke: > You wouldn't have the issue if the user had to themselves provide the -JJ

This can be already achieved via another SW option `spark.ext.h2o.extra.properties`. This option puts the whole value to command line. I don't know why i didn't considered for my use case in tests.

My mind was driven Spark options `spark.executor.extraJavaOptions` and `spark.driver.extraJavaOptions` that we can use for the internal backend and the fact that we don't have exact alternative for the automatically started external backend. 

@michalkurka if you think that the existing option `spark.ext.h2o.extra.properties` is sufficient, I will close this PR.

michalkurka: I think maybe `spark.ext.h2o.extra.properties` could be sufficient then. It is not a common use case to try to modify the System properties of the Hadoop job (needs a lot of knowledge about the internals). So I think requiring users to explicitly use the -JJ is reasonable.

PR #2379, opened 2020-11-02
[SW-2476] asH2OFrame Could Fail on ArrayIndexOutOfBoundsException

mn-mikke:
https://h2oai.atlassian.net/browse/SW-2476


extensions/src/main/scala/ai/h2o/sparkling/extensions/rest/api/ImportFrameHandler.scala
```diff
@@ -83,7 +83,8 @@ class ImportFrameHandler extends Handler {
       val vectorsToBeConvertedToString = indicesOfChanges.map(frame.vec(_))
       val domainIndices = for ((domain, idx) <- stringDomains.zipWithIndex if domain == null) yield idx
       val convertCategoricalToStringColumnsTask = new ConvertCategoricalToStringColumnsTask(frame._key, domainIndices)
-      convertCategoricalToStringColumnsTask.doAll(Vec.T_STR, vectorsToBeConvertedToString: _*)
```
mn-mikke, line 86: Here is the problem.

Pscheidl (approved):

honzasterba (approved):

PR #2380, opened 2020-11-02
[SW-2477] Replace IcedHashMapWrapper with New guessType Method On PreviewParseWriter

mn-mikke:
This PR depends on https://github.com/h2oai/h2o-3/pull/5086


extensions/src/main/scala/water/parser/CategoricalPreviewParseWriter.java
```diff
@@ -17,29 +17,30 @@
...
-  }
+    PreviewParseWriter.IDomain domainWrapper =
+        new PreviewParseWriter.IDomain() {
+          public int size() {
+            return domain.length;
+          }
+
+          public boolean contains(String value) {
+            for (String domainValue : domain) {
+              if (value.equals(domainValue)) return true;
```
honzasterba, line 37: since this array can be pretty big what about doing sort and binary search after?

mn-mikke: @honzasterba I understand your motivation, but this method is called only when [the array size is 1](https://github.com/h2oai/h2o-3/pull/5086/files#diff-1f9b998f1c157b03f6fb652a49fa1460a79ec7a30e569563579f0d7326d027b4R181). I'm afraid if we sorted the array at the begging of `guessType` method, it would slow down most of the cases.

Pscheidl: For small amount of items, the slowdown introduced by sorting is absolutely negligible. For larger arrays, the search will affect performance heavily.

michalkurka: agree with @mn-mikke - this is fine as-is

If we wanted to be uber-correct then we could replace 'contains' with 'isSingleValue'

Pscheidl (approved):

michalkurka (approved):

PR #2394, opened 2020-11-17
[SW-2485] Deprecate Apache Spark 2.1

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
```diff
@@ -488,6 +493,11 @@ object H2OContext extends Logging {
           s" but your Spark is of version ${sc.version}. Please make sure to use correct Sparkling Water for your" +
           s" Spark and re-run the application.")
     }
+    if (isSparkVersionDeprecated) {
+      logWarning(
+        "Apache Spark 2.1.x is deprecated and " +
```
honzasterba, line 498: I would make this parametrized by the current version `SparkSessionUtils.active.version` as well

core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
```diff
@@ -467,9 +467,14 @@ object H2OContext extends Logging {
     getOrCreate(Option(instantiatedContext.get()).map(_.getConf).getOrElse(new H2OConf()))
   }
 
+  private def isSparkVersionDeprecated(): Boolean = {
+    SparkSessionUtils.active.version.startsWith("2.1")
```
satai, line 471: What about spark 2.10? ;-)
(I know, it doesn't exist but I just warn about this pattern.)

mn-mikke: ok, I will improve that

core/src/main/scala/ai/h2o/sparkling/H2OContext.scala
mn-mikke: will do

honzasterba (approved):

PR #2396, opened 2020-11-18
[SW-2486] Upgrade MOJO runtime to 2.5.3

mn-mikke:
Upgrade is needed because of running DAI big data recipe on Sparkling Water.


honzasterba (approved):

mmalohlava:
Thanks a lot!

PR #2405, opened 2020-11-26
[SW-2489] Downgrade Sparklyr to 1.3.1 due to Problems with K8s Deployment

mn-mikke:
This PR should fix R tests on K8s.

More details in https://github.com/sparklyr/sparklyr/issues/2832

### DIFF
--- a/kubernetes/build.gradle
+++ b/kubernetes/build.gradle
@@ -68,21 +68,8 @@ task createRDockerfile(type: Dockerfile) {
                 R -e 'install.packages("RCurl", repos = "http://cran.us.r-project.org")' && \\
                 R -e 'install.packages("jsonlite", repos = "http://cran.us.r-project.org")' && \\
                 R -e 'install.packages("testthat", repos = "http://cran.us.r-project.org")' && \\
-                R -e 'install.packages("base64enc’", repos = "http://cran.us.r-project.org")' && \\
-                R -e 'install.packages("config", repos = "http://cran.us.r-project.org")' && \\
-                R -e 'install.packages("DBI", repos = "http://cran.us.r-project.org")' && \\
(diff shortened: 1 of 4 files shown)


kubernetes/build.gradle
```diff
@@ -68,7 +68,21 @@ task createRDockerfile(type: Dockerfile) {
...
+                R -e 'install.packages("forge", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("generics", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("globals", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("purrr", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("r2d3", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("rappdirs", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("rjson", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("tidyr", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("uuid", repos = "http://cran.us.r-project.org")' && \\
+                R -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/sparklyr/sparklyr_1.3.1.tar.gz", repos=NULL, type="source")'
```
honzasterba, line 85: rather than installing all deps manually why not use `library(devtools); install_version(...)'?

Pscheidl (approved):
Installing the package once automatically with all dependencies would be nice. Is there any reason why this can't be done ? 

If this is about to take long,I vote for merging it. Some tests are better than no tests.

kubernetes/build.gradle
mn-mikke: Great suggestion, thanks!

honzasterba (approved):

PR #2408, opened 2020-11-30
[SW-2492] Disable Kubernetes R Tests for External Backend

mn-mikke:
The R tests will  be enabled again after Sparklyr 1.5.1. due to https://github.com/sparklyr/sparklyr/issues/2832. In the external beckend a multiple --conf options needs to be passed.

### DIFF
--- a/ci/Jenkinsfile-kubernetes
+++ b/ci/Jenkinsfile-kubernetes
@@ -640,8 +640,9 @@ node("docker") {
                 withTestEnv(commons, image) {
                     // R sets deployment mode automatically
                     testRInternalBackend(registryId, master, version, sparkVersion)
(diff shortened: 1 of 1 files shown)


honzasterba (commented):
is it such a problem to keep the tests failing? if it does not prevent other tests from running I would keep this as is. this is easy to forget about

mn-mikke:
> is it such a problem to keep the tests failing? if it does not prevent other tests from running I would keep this as is. this is easy to forget about

@honzasterba It happens time to time that EKS infrastructure keeps running because of a failed test.

mn-mikke:
This PR is not needed anymore. Sparklyr 1.5.1 has been released. I've manually tested the effect of the changes.

PR #2415, opened 2020-12-05
[SW-2498] Try to Lock Cloud Multiple Times

mn-mikke:
In automatic external backend of K8s deployment, cloud locking may fail since H2O headless service may route a call to a non leader node during its initialization phase.

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
@@ -114,21 +114,31 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
     }
   }
 
+  private def tryToLockCloud(conf: H2OConf, attemptId: Int, maximumNumberOfAttempts: Int): Boolean = {
+      val h2oCluster = conf.h2oCluster.get + conf.contextPath.getOrElse("")
+      val h2oClusterName = conf.cloudName.get
+      try {
+        logInfo(s"Trying to lock H2O cluster $h2oCluster - $h2oClusterName.")
+        lockCloud(conf)
+        true
+      } catch {
+        case cause: RestApiException if attemptId < maximumNumberOfAttempts - 1 =>
+          logWarning(s"Locking of the H2O cluster $h2oCluster - $h2oClusterName failed.", cause)
+          false
+      }
+  }
+
   protected def getAndVerifyWorkerNodes(conf: H2OConf): Array[NodeDesc] = {
-    val h2oCluster = conf.h2oCluster.get + conf.contextPath.getOrElse("")
-    val h2oClusterName = conf.cloudName.get
     try {
-      var numberOfTries = 6
-      while (numberOfTries > 1) {
-        try {
-          logInfo(s"Trying to lock H2O cluster $h2oCluster - $h2oClusterName.")
-          lockCloud(conf)
-          numberOfTries = 0
-        } catch {
-          case cause: RestApiException if numberOfTries > 1 =>
-            numberOfTries = numberOfTries - 1
-            logWarning(s"Locking of the H2O cluster $h2oCluster - $h2oClusterName failed.", cause)
-            Thread.sleep(10000) // Sleep for 10 seconds
+      val maximumNumberOfAttempts = 6
+      var attemptId = 0
+      while (attemptId < maximumNumberOfAttempts) {
+        val locked = tryToLockCloud(conf, attemptId, maximumNumberOfAttempts)
+        if (locked) {
+          attemptId = maximumNumberOfAttempts
+        } else {
+          Thread.sleep(10000) // Wait for 10 seconds
+          attemptId = attemptId + 1
         }
       }
       verifyWebOpen(conf)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -115,21 +115,34 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
   }
 
   protected def getAndVerifyWorkerNodes(conf: H2OConf): Array[NodeDesc] = {
+    val h2oCluster = conf.h2oCluster.get + conf.contextPath.getOrElse("")
+    val h2oClusterName = conf.cloudName.get
     try {
-      lockCloud(conf)
+      var numberOfTries = 6
+      while (numberOfTries > 1) {
```
honzasterba, line 122: this is confusing, this will perform the re-try only 5 times, while `numberOfTries` is 6
why not do the normal `var numberOfTries = 0; while (numberOfTries < 6)` and `numberOfTries += 1`?

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -115,21 +115,34 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
...
+        try {
+          logInfo(s"Trying to lock H2O cluster $h2oCluster - $h2oClusterName.")
+          lockCloud(conf)
+          numberOfTries = 0
+        } catch {
```
honzasterba, line 127: try-catch within try-catch is difficult to read, would move this into a separate method

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -115,21 +115,34 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
...
+          case cause: RestApiException if numberOfTries > 1 =>
```
honzasterba, line 128: interesting construct, so this means this catch only works when the `if` condition is sattisfied?

mn-mikke: Yes, you are correct.

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
mn-mikke: Np, will do.

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
mn-mikke: ok, will change to the proposed logic. Thanks!

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -114,21 +114,31 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
...
-          numberOfTries = 0
-        } catch {
-          case cause: RestApiException if numberOfTries > 1 =>
-            numberOfTries = numberOfTries - 1
-            logWarning(s"Locking of the H2O cluster $h2oCluster - $h2oClusterName failed.", cause)
-            Thread.sleep(10000) // Sleep for 10 seconds
+      val maximumNumberOfAttempts = 6
+      var attemptId = 0
+      while (attemptId < maximumNumberOfAttempts) {
+        val locked = tryToLockCloud(conf, attemptId, maximumNumberOfAttempts)
```
honzasterba, line 136: passing a boolean value of `attemptId < maximumNumberOfAttempts - 1` would reduce the number of args and make the code cleaner

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
```diff
@@ -114,21 +114,31 @@ trait H2OContextExtensions extends RestCommunication with RestApiUtils with Shel
...
+        if (locked) {
+          attemptId = maximumNumberOfAttempts
```
honzasterba, line 138: could be just `break`

mn-mikke: Break and return statements in Scala are quite tricky. It's possible to define them, but they are implemented via thrown exceptions.  https://alvinalexander.com/scala/scala-break-continue-examples-breakable/ If you don't mind, I will leave it as it's.

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
mn-mikke: ok, will change.

core/src/main/scala/ai/h2o/sparkling/backend/utils/H2OContextExtensions.scala
honzasterba: I was suspecting that was the case :)

honzasterba (approved):

PR #2416, opened 2020-12-07
[SW-2489] Revert Sparklyr  Downgrade to 1.3.1

mn-mikke:
Sparklyr 1.5.1 with the fix for SW has been released. I've tested the effect of changes manually.

### DIFF
--- a/ci/Jenkinsfile-kubernetes
+++ b/ci/Jenkinsfile-kubernetes
@@ -33,9 +33,14 @@ def runCommandWithK8sDebugInfo(GString script) {
             kubectl logs sparkling-water-app
             kubectl get events
         """
-        if (!suppressFailure) {
-            error("The script finished with the exit code ${result}.")
-        }
+    }
+    return result
+}
+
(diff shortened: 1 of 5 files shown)


ci/Jenkinsfile-kubernetes
```diff
@@ -23,15 +23,19 @@ def readPropertiesFile(file) {
     return properties
 }
 
-def runCommandWithK8sDebugInfo(GString script) {
+def runCommandWithK8sDebugInfo(GString script, boolean suppressFailure = true) {
```
honzasterba, line 26: `false` would seem like a more reasonable default

mn-mikke: ok, will change it, thanks!

honzasterba (commented):
there seems to be a lot of stuff for which we ignore failiures
at least I would put this into a variable like `IGNORE_FAILIURES_FOR_SPARKLYR=true`

honzasterba (approved):

PR #2417, opened 2020-12-11
[SW-2496] Fix the Flow link for Databricks Azure

mn-mikke:
There are two problems.
- Databricks chanced base address which is not easy to resolve since contains a random number which is not stored anywhere on the driver or in the environment properties. The problem is solved via resolving base address in the web browser.

More details on: ```https://docs.microsoft.com/en-us/azure/databricks/workspace/per-workspace-urls```

- The `UIUtils.headerSparkPage` method in   Databricks Spark has more parameters than Apache Spark. I solved the problem via reflection call.

Databricks: ```
request: javax.servlet.http.HttpServletRequest,
title: String,content: => Seq[scala.xml.Node],
activeTab: org.apache.spark.ui.SparkUITab,
helpText: Option[String],
showVisualization: Boolean,
useDataTables: Boolean,
supportScreenshot: Boolean```

Apache: ```
request: javax.servlet.http.HttpServletRequest,
title: String,content: => Seq[scala.xml.Node],
activeTab: org.apache.spark.ui.SparkUITab,
helpText: Option[String],
showVisualization: Boolean,
useDataTables: Boolean```

### DIFF
--- a/core/src/main/scala_spark_others/org/apache/spark/h2o/SparkSpecificUtils.scala
+++ b/core/src/main/scala_spark_others/org/apache/spark/h2o/SparkSpecificUtils.scala
@@ -22,6 +22,7 @@ import org.apache.spark.SparkContext
 import org.apache.spark.h2o.ui.{AppStatusListener, AppStatusStore, CrossSparkUtils, SparklingWaterUITab}
 import org.apache.spark.status.ElementTrackingStore
 import org.apache.spark.ui.{SparkUITab, UIUtils}
+import java.lang.Boolean._
 
 import scala.xml.Node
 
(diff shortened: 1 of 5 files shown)


honzasterba (approved):

core/src/main/scala_spark_others/org/apache/spark/h2o/SparkSpecificUtils.scala
```diff
@@ -32,7 +33,24 @@ object SparkSpecificUtils extends CrossSparkUtils {
       content: => Seq[Node],
       activeTab: SparkUITab,
       helpText: String): Seq[Node] = {
-    UIUtils.headerSparkPage(request, "Sparkling Water", content, activeTab, helpText = Some(helpText))
+    val method = UIUtils.getClass.getMethods.find(m => m.getName == "headerSparkPage").get
```
satai, line 36: This one is obviously fragile and can break if an overloaded method is added to spark.ui

But I guess that you can live with this?

satai: Otherwise LGTM

honzasterba (approved):

core/src/main/scala_spark_others/org/apache/spark/h2o/SparkSpecificUtils.scala
mn-mikke: Yes, if the method doesn't exist the Spark UI shows an exception if Sparkling Water tab is clicked. For Apache Spark we know that there is such a method. This could if a user deploys SW to a obscure platform with a custom build of Spark which we haven't tested. IMHO, we can live with that.

PR #2419, opened 2020-12-13
[SW-2461] Add H2O RuleFit to Algo API

mn-mikke:


doc/src/site/sphinx/ml/sw_drf.rst
```diff
@@ -1,10 +1,17 @@
-.. _drf:
```
michalkurka, line 1: Why do we see changes in sw_drf documentation? Is that intended?

doc/src/site/sphinx/ml/sw_rule_fit.rst
```diff
@@ -0,0 +1,114 @@
+.. _drf:
```
michalkurka, line 1: I think the files got mixed up

mn-mikke: Spot on! Thanks for catching that

koniecsveta (approved):
LGTM, thanks!

PR #2420, opened 2020-12-14
[SW-2500] Temporarily Disable Beta Constraints Tests

mn-mikke:
https://h2oai.atlassian.net/browse/SW-2500

### DIFF
--- a/py/tests/unit/with_runtime_sparkling/test_glm.py
+++ b/py/tests/unit/with_runtime_sparkling/test_glm.py
@@ -177,28 +177,28 @@ def testRandomColsMustBeWithinTrainingDataset(semiconductorDataset):
         glm.fit(semiconductorDataset)
 
 
-def testBetaConstraintsAffectResult(spark, prostateDataset):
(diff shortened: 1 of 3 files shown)


py/tests/unit/with_runtime_sparkling/test_glm.py
```diff
@@ -177,28 +177,28 @@ def testRandomColsMustBeWithinTrainingDataset(semiconductorDataset):
         glm.fit(semiconductorDataset)
 
 
-def testBetaConstraintsAffectResult(spark, prostateDataset):
```
satai, line 180: Wouldn't 
@unittest.skip("SW-2500")
work here?

satai (approved):

honzasterba (changes requested):
there has to be a better way than this:
are failing tests such a problem influencing other test? then we should fix the test-framework
is using a @ignore annotation not possible? we should fix the test framework

mn-mikke:
@honzasterba @satai Thanks guys for your comments, much appreciated! I was just too lazy to read pytest documentation.

honzasterba (approved):

Issue #2414, opened 2020-12-03
Tried using port 54321 for Flow proxy, but port was already occupied


dzlab:
Providing us with the observed and expected behavior definitely helps. Giving us with the following information definitively helps:

-  Spark version `2.4.7`
- Sparkling Water `3.32.0.1-2-2.4`
- Hadoop Version & Distribution `3.3.0`
- Execution mode Kubernetes deployment (spark jobs submitted in client mode)
- Are you using Windows/Linux/MAC? Kubernetes on Docker Desktop on Mac

I've a Spark application with dependencies on H2O Sparkling Water jars

```
// H2O Sparkling Water version (spark 2.4)
val h2oVersion = "3.32.0.1-2-2.4"
lazy val h2oLibs = Seq(
  "org.apache.spark" %% "spark-repl" % sparkVersion,
  "ai.h2o" % "sparkling-water-package_2.11" % h2oVersion //exclude("ai.h2o", "sparkling-water-api-generation")
)
```
When I try to start it on k8s, the Spark executor starts properly (see logs from executor container below) and I can open a port-forward connection to it and access flow UI.

```
12-03 17:04:55.129 10.1.0.22:54321       33     ent-loop-0  INFO water.default: ----- H2O started  -----
12-03 17:04:55.129 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Build git branch: rel-zermelo
12-03 17:04:55.129 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Build git hash: 8a289cc0c15718842afc54dca7c8ed104aec9bd4
12-03 17:04:55.129 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Build git describe: jenkins-master-5215-21-g8a289cc
12-03 17:04:55.130 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Build project version: 3.32.0.1
12-03 17:04:55.130 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Build age: 1 month and 24 days
12-03 17:04:55.130 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Built by: 'jenkins'
12-03 17:04:55.130 10.1.0.22:54321       33     ent-loop-0  INFO water.default: Built on: '2020-10-08 18:16:09'
[...]

PR #2425, opened 2020-12-16
[SW-2503] Tried using port 54321 for Flow proxy, but port was already occupied

dzlab:
Updated `ProxyStarter.scala`. so that when tries to bind on a port it uses the address `0.0.0.0`

The problem description: https://github.com/h2oai/sparkling-water/issues/2414


mn-mikke (approved):
LGTM

Pscheidl (approved):
This is already used for the Jetty to start anyway, so checking all local interfaces is more appropriate anyway.

See `SparklingWaterJettyHelper.scala` class.

```scala
  def startServer(port: Int): Server = {
    val jettyServer = createJettyServer("0.0.0.0", port)
    val context = createServletContextHandler()
    if (h2oHttpView.getConfig.loginType != LoginType.NONE) {
      context.setSecurityHandler(authWrapper(jettyServer).asInstanceOf[SecurityHandler])
    }
    jettyServer.setHandler(context)
    jettyServer.start()
    jettyServer
  }
  ```

mn-mikke:
@dzlab Thanks for you contribution! Merging...

PR #2428, opened 2020-12-17
[SW-2505] Scala Scoring Package

mn-mikke:
The Scala scoring package will be tested via integration tests of Python Scoring package, which is going to be created.


README.rst
```diff
@@ -43,8 +43,9 @@ Each Sparkling Water release is published into Maven central with following coor
 - ``ai.h2o:sparkling-water-examples_{{scala_version}}:{{version}}`` - Includes example applications
 - ``ai.h2o:sparkling-water-repl_{{scala_version}}:{{version}}`` - Spark REPL integration into H2O Flow UI
 - ``ai.h2o:sparkling-water-ml_{{scala_version}}:{{version}}`` - Extends Spark ML package by H2O-based transformations
-- ``ai.h2o:sparkling-water-scoring_{{scala_version}}:{{version}}`` - Lightweight package used for Scoring with H2O MOJOs. This package does not have run-time dependency on H2O cluster.
-- ``ai.h2o:sparkling-water-package_{{scala_version}}:{{version}}`` - Uber Sparkling Water package containing all dependencies. This is designed to use as Spark package via ``--packages`` option
+- ``ai.h2o:sparkling-water-scoring_{{scala_version}}:{{version}}`` - A library containing scoring logic and definition of Sparkling Water MOJO models.
+- ``ai.h2o:sparkling-water-package-scoring_{{scala_version}}:{{version}}`` - Lightweight Sparkling Water package including all dependencies required just for scoring with H2O-3 and DAI MOJO models.
```
honzasterba, line 47: I dont like the naming very much, is there some hidden meaning in
- scoring
- scoring_package
- package
?
I think scoring/scoring_package is confusing, what about `scoring_package` and `mojo_package`?

mn-mikke: `package` is a fat jar including all dependencies required for traning and model scoring. This naming comes from spark terminology since you can do ``spark-submit --packages ai.h2o:sparkling-water-scoring_2.11:3.32.0.2-1-2.4`` and spark downloads the dependency from maven repository and distributes across the cluster.

So `scoring-package` sounds good to me for the fat jar with all dependencies required just for scoring. (7MB vs. 150MB of `package)

`scoring` is the same as `scoring-package`, but it's a regular jar without dependecies. I think we could rename it to `models` if you think that the existence of `scoring` and `scoring-package` is confusing. WDYT?

honzasterba: thanks for the explanation, now the names make sense to me, I would leave it as is

honzasterba (approved):

README.rst
michalkurka: @mn-mikke you can also use "standalone" classifier to denote the difference between fat jar and regular artifact, this is sometimes used - hive uses it eg. for the jdbc driver

mn-mikke: Ok, I didn't know about the classifiers, thanks for mentioning it! Personally, I don't have any preferences. So @michalkurka, @honzasterba, @Pscheidl, @satai  what name of the new artefact do you prefer?
`sparkling-water-scoring-package-{version}.jar` or `sparkling-water-scoring-{version}-standalone.jar`?

Note: we already have `sparkling-water-package-{version}.jar` containing all dependencies for both (training and scoring).

michalkurka: I think it is fine as you have it, just wanted to mention another option

honzasterba: Standalone is something regulary used in java world, but as @mn-mikke said if `package` is the term used for standalone libs in spark I would stick with `package`

mmalohlava:
👏 nice nice!

PR #2432, opened 2020-12-26
[SW-2509] Increase Limit of K8s Tests for Automatic External Backend

mn-mikke:
After everything is finished. Spark driver pod remains in a running state for several minutes. This PR helps tests to be always green. The issue will be definitely fixed after rewriting automatic external backend to H2O K8s operator.


honzasterba (approved):

honzasterba:
420 baby!

PR #2435, opened 2021-01-04
[SW-2499] Document Properties for running SW on EMR 5.32

mn-mikke:
JIRA ticket: https://h2oai.atlassian.net/jira/software/c/projects/SW/issues/SW-2499
Answer from AWS team: https://forums.aws.amazon.com/thread.jspa?threadID=332806

### DIFF
--- a/doc/src/site/sphinx/configuration/internal_backend_tuning.rst
+++ b/doc/src/site/sphinx/configuration/internal_backend_tuning.rst
@@ -57,6 +57,12 @@ Furthermore, we recommend that you configure the following Spark properties to s
 | ``spark.dynamicAllocation.enabled``             | ``false``                | Disable Spark support for  |
 |                                                 |                          | dynamic allocation.        |
 +-------------------------------------------------+--------------------------+----------------------------+
+| ``spark.yarn.heterogeneousExecutors.enabled``   | ``false``                | Disable heterogeneous      |
+|                                                 |                          | executors. This option is  |
+|                                                 |                          | enabled by default on some |
+|                                                 |                          | versions of AWS EMR        |
+|                                                 |                          | (e.g. 5.32.0)              |
++-------------------------------------------------+--------------------------+----------------------------+
 | ``spark.yarn.am.extraJavaOptions``              | ``-XX:MaxPermSize=384m`` | Increase PermGem if you    |
 |                                                 |                          | are running in Java7 on    |
 |                                                 |                          | the Yarn application       |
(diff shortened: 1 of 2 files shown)


neemah2o:
@mn-mikke , should we add some guidelines for users in the docs explaining this issue for EMR 5.32? I assume the aws_emr.rst addition patches the issue, but maybe we can also put a note at the bottom of this [page](http://docs.h2o.ai/sparkling-water/2.4/latest-stable/doc/deployment/aws_emr.html)

```
Note: It is important to add the following Spark configuration to any of your EMR jobs:

--conf "spark.dynamicAllocation.enabled=false"
--conf spark.yarn.heterogeneousExecutors.enabled=false --conf spark.scheduler.minRegisteredResourcesRatio=1
```

honzasterba (approved):

mn-mikke:
@neemah2o Should I explicitly mention that `spark.yarn.heterogeneousExecutors.enabled=false` is needed for EMR 5.32.0. IMHO, it's not a big deal if user specifies this parameter also on other versions of EMR. https://github.com/h2oai/sparkling-water/pull/2435/files#diff-296fd6ffe209a7d8c45d5fb274fe650cbfa9aa6ba6d49a86d856c39078c89efcR65

neemah2o:
@mn-mikke , it would be great if AWS was more transparent and told us which EMR versions need `spark.yarn.heterogeneousExecutors.enabled=false`, but unfortunately it's not documented. I think it would be good. to mention it is needed for EMR 5.32.0. If things change in EMR 5.33+ then we can adjust for that.
Do we have automated tests for EMR like we have for DBC? It may be good to have tests to catch these issues for future versions.

mn-mikke:
@neemah2o ok, I will update the note.

> Do we have automated tests for EMR like we have for DBC?
Yes and no. There are no smoke tests to test compatibility with specific EMR versions. More than a year ago, I've created a performance tests which run on EMR. So we can leverage the current logic for deploying EMR cluster and make modification to run smoke tests. This should take much time, but we need to discuss what EMR versions to test? (just the latest version for EMR 5 and the latest for EMR 6?)

neemah2o:
> This should take much time, but we need to discuss what EMR versions to test? (just the latest version for EMR 5 and the latest for EMR 6?)

Good point, @mn-mikke , definitely the latest should be tested but I don't think that would cover enough. Maybe the last few for each type (5.x, 6.x). @g-eoj , do you have any thoughts on what would be a good range?

PR #2438, opened 2021-01-13
[SW-2514] Delete Train and Validation H2O Frame after Training a Model

mn-mikke:
The H2O Frames are created during the call of SW wrapper's `fit` function. The frames are accessible just via Flow UI and are not exposed to the user. We can consider this as a memory leak.


honzasterba (approved):

honzasterba:
what about some tests making sure that generated objects are being cleaned up?

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
```diff
@@ -90,6 +90,8 @@ class H2OAutoML(override val uid: String)
 
     val algoName = getLeaderboard().select("model_id").head().getString(0)
     deleteRegisteredH2OFrames()
+    train.delete()
```
michalkurka, line 93: should you instead call registerFrameForDeletion? and let the frames be cleaned up by deleteRegisteredH2OFrames?

michalkurka:
unrelated idea: I checked the method "prepareDatasetForFitting" and got the idea of converting training and validation frame at the same time in parallel

question: didn't notice blending frame - do we support blending in SW?

mn-mikke:
> question: didn't notice blending frame - do we support blending in SW?
@michalkurka  `blending_frame` and `leaderboard_frame" haven't exposed on AutoML yet. If these frames are not too big (can be serialized to one node), we can expose them. We already have parameters of the `DataFrame` type in SW.

michalkurka:
@mn-mikke Leaderboard is tiny

blending frame is just like validation frame (same size), it is used on AutoML input when people don't want to do cross validation

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
mn-mikke: Like this idea!

honzasterba (approved):

Pscheidl (approved):

PR #2457, opened 2021-02-11
[SW-2527] Remove Deprecated Parameter `withDetailedPredictionCol` from MOJOSettings

mn-mikke:


michalkurka:
@mn-mikke how long was this option marked as deprecated?

mn-mikke:
@michalkurka This was deprecated in 3.30.1.1-1 (Released: 12/Aug/20) by [SW-2406](https://h2oai.atlassian.net/browse/SW-2406)

michalkurka:
This is only about half year ago. Given the recent issues with deprecating parameters - I would keep it around for longer. WDYT?

mn-mikke:
@michalkurka Considering that this option is frequently used and not all users upgraded from 3.28 for example, your proposal makes perfect sense. Thanks for questioning it! WDYT about removing it in 3.36?

PR #2462, opened 2021-02-19
[SW-2532] Extend Target Encoder to Multinomial Problems

mn-mikke:
The pr depends on h2oai/h2o-3#5231 and https://github.com/h2oai/sparkling-water/pull/2462.


sebhrusen (commented):
Looks good overall, I just have a small question regarding the TE model output when you use it as input for a model: do you add the non-numerical columns ex post in your TEModel `transform` method?
I made changes on the backend so that it shouldn't be needed anymore there, but when you use the Mojo, I'm afraid it still necessary for you to do it.

doc/src/site/sphinx/ml/sw_target_encoder.rst
```diff
@@ -153,7 +154,8 @@ Target Encoder in Sparkling Water is implemented as a regular estimator and thus
             from pysparkling.ml import H2OTargetEncoder
             targetEncoder = H2OTargetEncoder()\
               .setInputCols(["RACE", "DPROS", "DCAPS"])\
-              .setLabelCol("CAPSULE")
+              .setLabelCol("CAPSULE")\
+              .setProblemType("Binomial")
```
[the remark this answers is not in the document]
mn-mikke: Good to know, thanks!

ml/src/main/scala/ai/h2o/sparkling/ml/params/H2OTargetEncoderProblemType.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package ai.h2o.sparkling.ml.params;
+
+public enum H2OTargetEncoderProblemType {
+    Binomial,
+    Multinomial,
+    Regression
+}
```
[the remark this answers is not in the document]
mn-mikke: Good point, will improve it.

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -96,6 +97,11 @@ class H2OTargetEncoder(override val uid: String)
   }
 
   def setNoiseSeed(value: Long): this.type = set(noiseSeed, value)
+
+  def setProblemType(value: String): this.type = {
```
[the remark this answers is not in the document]
mn-mikke: This is the way how enum values are passed in Spark or SW in general. Personally, I see two benefits. it gives consistent API across different languages. And the usage is more straightforward  in interactive shells. In sparkling-shell, an enum would have to be imported first before passing its value.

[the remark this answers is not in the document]
sebhrusen: definitely makes sense if this is used directly in shell.

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderMultinomialTestSuite.scala
```diff
@@ -0,0 +1,490 @@
...
+
+  test("A pipeline with a target encoder transform training and testing dataset without an exception") {
+    val targetEncoder = new H2OTargetEncoder()
+      .setInputCols(Array("RACE", "DPROS", "DCAPS"))
+      .setLabelCol("AGE")
+      .setProblemType("Multinomial")
+
+    val gbm = new H2OGBM().setLabelCol("AGE")
+
+    val pipeline = new Pipeline().setStages(Array(targetEncoder, gbm))
```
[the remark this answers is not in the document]
mn-mikke: As a dataframe flows through the pipeline, each stage adds own columns from the dataset. It's up to each stage to decide which columns are imported for its job. SW algorithms use all columns if `intputCols` are not specified, so "yes" is a short answer for your question. :-)

core/src/test/scala/ai/h2o/sparkling/TestUtils.scala
```diff
@@ -74,6 +75,29 @@ object TestUtils extends Matchers {
...
+      produced: DataFrame): Unit = {
+    val tmpDir = Files.createTempDirectory("dfComparison").toFile
+    val loadedProduced = try {
+      val path = tmpDir.getAbsolutePath + "/file"
+      produced.write.parquet(path)
+      spark.read.parquet(path)
+    } finally {
+      deleteOnExit(tmpDir)
+    }
+    assertDataFramesAreIdentical(expected, loadedProduced)
```
michalkurka, line 90: I also check the implementation of this method and it is not actually testing the datasets are identical.

    scala> val left = Seq((1, 0.0), (1, 0.0), (1, 1.0))
    left: Seq[(Int, Double)] = List((1,0.0), (1,0.0), (1,1.0))

    scala> val right = Seq((1, 1.0), (1, 0.0), (1, 1.0))
    right: Seq[(Int, Double)] = List((1,1.0), (1,0.0), (1,1.0))

    scala> spark.createDataFrame(left).except(spark.createDataFrame(right)).count()
    res17: Long = 0                                                                 

    scala> spark.createDataFrame(right).except(spark.createDataFrame(left)).count()
    res18: Long = 0

    scala> spark.createDataFrame(right).distinct().count()
    res20: Long = 2

    scala> spark.createDataFrame(left).distinct().count()
    res21: Long = 2

The compared datasets are not identical because left has 2 zeros in second column, while the right has only 1.

doc/src/site/sphinx/ml/sw_target_encoder.rst
```diff
@@ -120,6 +120,7 @@ Target Encoder in Sparkling Water is implemented as a regular estimator and thus
             val gbm = new H2OGBM()
                 .setFeaturesCols(targetEncoder.getOutputCols())
                 .setLabelCol("CAPSULE")
+                .setProblemType("Binomial")
```
michalkurka, line 123: do we use "problem type" term other places in SW?

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
```diff
@@ -43,7 +43,8 @@ class H2OTargetEncoder(override val uid: String)
       "H2OContext needs to be created in order to use target encoding. Please create one as H2OContext.getOrCreate().")
     val input = h2oContext.asH2OFrame(dataset.toDF())
     val distinctInputCols = getInputCols().flatten.distinct
-    input.convertColumnsToCategorical(distinctInputCols)
+    val toCategorical = if (getProblemType() == "Regression") distinctInputCols else distinctInputCols ++ Seq(getLabelCol())
```
michalkurka, line 46: shouldn't this be case insensitive, spark ml accepts eg. "family" = "multinomial"

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBinomialTestSuite.scala
```diff
@@ -118,6 +126,8 @@ class H2OTargetEncoderBinomialTestSuite extends FunSuite with Matchers with Shar
       .setHoldoutStrategy("None")
       .setBlendedAvgEnabled(true)
       .setNoise(0.0)
+      .setProblemType("Binomial")
```
michalkurka, line 129: is setting problem type mandatory? if it is - can we auto-detect it and allow old code to run?

doc/src/site/sphinx/ml/sw_target_encoder.rst
mn-mikke: No, just here. Other places use H2O-3 terms like `family` and `distribution`. Can you think of a better name for this field?

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoder.scala
mn-mikke: the [setter method](https://github.com/h2oai/sparkling-water/pull/2462/files#diff-cfe2d864665ec1d4b957eab0d39ad7117e6af6a2728d509270caee83b8449475R102) and python -> scala parameter serializer should convert any string value into this capitalized form, but I will use the lowercase form in test to be 100% sure.

michalkurka: I didn't know that, don't worry then! thank you for explaining

doc/src/site/sphinx/ml/sw_target_encoder.rst
michalkurka: I think it is strange to set both problem type and distribution/family on the H2OGBM object,
in other places you have

```

new H2OGBM()
   .setSeed(42)
   .setDistribution("bernoulli")
```

adding `setProblem` there with no connection to the distribution parameters is in my opinion not ideal.

We could 
 - drop the problem type parameter completely and let them set the distribution parameter
 - keep the problem type but make it optional and derive it automatically from the distribution parameter if it is passed
 
 @sebhrusen thoughts?

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OTargetEncoderBinomialTestSuite.scala
mn-mikke: It's not mandatory. Problem typy is [binomial by default](https://github.com/h2oai/sparkling-water/pull/2462/files/aa72f6e665bd80005d79d9836e293a16ba68e607#diff-de07ee0fa2b229cdb67aee97213ff7d1e27a9089c556e9ec92b02c84df6c29a3R91). For automatic detection, I would need to change the output type for binomial and regression to a vector since [here](https://github.com/h2oai/sparkling-water/pull/2462/files#diff-5c5fcf2c0792550b3e9863f7c39fbcf432ef350bef92b70f4793830cc1fe7e4aR26-R33) I don't have an access to a dataset. This might be a breaking change to some level, but people using TE in pipeline with a H2O algorithm or [VectorAssembler](https://spark.apache.org/docs/latest/ml-features#vectorassembler) (Spark algorithms take vector as an input) wouldn't see any difference IMHO.

core/src/test/scala/ai/h2o/sparkling/TestUtils.scala
mn-mikke: Nice catch, thanks a lot!

doc/src/site/sphinx/ml/sw_target_encoder.rst
sebhrusen: from what I understand, `problemType` is useful only to decide if the response column should be converted to categorical, right? SW's approach being different from H2O-3 on this aspect, being more declarative.

Also by the way, `binomial` and `multinomial` usually refer to distributions, so using them as `problemType` values may confuse some. I would suggest to use `binary` and `multiclass` instead for those classification types.

This being said, I'd vote for your 2nd suggestion @michalkurka :
> keep the problem type but make it optional and derive it automatically from the distribution parameter if it is passed

and I don't see any better alternative.

mn-mikke:
@sebhrusen @michalkurka I tried to address your comments. May I ask you to take a look one more time?

py-scoring/src/ai/h2o/sparkling/ml/params/H2OTargetEncoderMOJOParams.py
```diff
@@ -92,6 +92,16 @@ class H2OTargetEncoderMOJOParams(Params):
         "A seed of the generator producing the random noise",
         H2OTypeConverters.toInt())
 
+    problemType = Param(
+        Params._dummy(),
+        "problemType",
+        """A type of ML problem type for which the target encoder will be used for:
+              Auto           - If this option is chosen (default), the problem type will be automatically resolved based
```
michalkurka, line 99: Maybe it should specify what is the logic behind it.

Eg. some people might expect 0/1 column to be categorical.

michalkurka (commented):
Looks Great to me! Thank you @mn-mikke for making the changes.

sebhrusen (approved):

py-scoring/src/ai/h2o/sparkling/ml/params/H2OTargetEncoderMOJOParams.py
mn-mikke: Good point, i will also make the logic simpler. ( boolean or string => classification, otherwise regression)

michalkurka (approved):
Thank you @mn-mikke

PR #2474, opened 2021-03-16
[SW-2539] Disable SSL Certificate Verification in Python Client and Spark Instances Separately

mn-mikke:
This PR introduces new configuration option `spark.ext.h2o.internal.rest.verify_ssl_certificates` and renames the unreleased property `spark.ext.h2o.verify_ssl_hostnames` to `spark.ext.h2o.internal.rest.verify_ssl_hostnames`.


michalkurka (approved):
LGTM!

PR #2499, opened 2021-04-14
[SW-2549] Fix Memory Leak of Frames in H2OAutoML

mn-mikke:


michalkurka:
Thank you @mn-mikke. Leaking the training frames is definitely a worse problem than not having the "extra" columns calculated. Good fix.

michalkurka (approved):

PR #2500, opened 2021-04-16
[SW-2551] Introduce Configuration Property for Setting CA Certificate(s)

mn-mikke:
(for PySparkling and RSparkling client)

### DIFF
--- a/py/src/ai/h2o/sparkling/H2OContext.py
+++ b/py/src/ai/h2o/sparkling/H2OContext.py
@@ -48,6 +48,7 @@ class H2OContext(object):
         kwargs = {}
         kwargs["https"] = schema == "https"
         kwargs["verify_ssl_certificates"] = conf.verifySslCertificates()
+        kwargs["cacert"] = conf.sslCACert()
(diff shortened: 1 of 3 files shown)


satai (approved):

py/src/ai/h2o/sparkling/H2OContext.py
```diff
@@ -48,6 +48,7 @@ def __h2o_connect(h2o_context):
         kwargs = {}
```
michalkurka, line 48: It is definitely a good idea to have the options that make sense to be exposed curated, however, wouldn't it also make sense to provide a fallback mechanism so that users can always override/define an option even if is not exposed by SW? This could help with experimenting and problem-solving. WDYT?

michalkurka (approved):

py/src/ai/h2o/sparkling/H2OContext.py
mn-mikke: Yeah, good idea! I will work on it in a different PR. Thanks!

PR #2505, opened 2021-04-19
[SW-2545] Expose Parameter for Setting Leaderboard DataFrame on H2OAutoML

mn-mikke:


ml/src/main/scala/ai/h2o/sparkling/ml/params/NullableBigDataFrameParam.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream, ObjectInputStream, ObjectOutputStream}
+import java.util.Base64
+
+import ai.h2o.sparkling.utils.ScalaUtils.withResource
+import org.apache.spark.ml.param.{Param, Params}
+import org.apache.spark.sql.{DataFrame, Row}
+import org.json4s.JsonAST.{JNull, JString}
+import org.json4s.jackson.JsonMethods.{compact, parse, render}
+
+// This type of Dataframe parameter just serialize execution plans, not data.
```
michalkurka, line 29: What is the motivation behind that?

I was thinking - once the pipeline is fitted, the reference to the leaderboard frame can be released. And serializing an unfitted pipeline with a defined leaderboard frame doesn't really make sense to me.

Please take this as me thinking out loud.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AutoMLConfiguration.scala
```diff
@@ -16,22 +16,25 @@ trait AutoMLConfiguration extends AlgorithmConfigurations {
       ("H2OAutoMLStoppingCriteriaParams", classOf[AutoMLStoppingCriteriaV99], classOf[AutoMLStoppingCriteria], Getter),
       ("H2OAutoMLBuildModelsParams", classOf[AutoMLBuildModelsV99], classOf[AutoMLBuildModels], Field))
 
+    val leaderboardFrame =
```
michalkurka, line 19: please see bellow - could we just make it transient?

ml/src/main/scala/ai/h2o/sparkling/ml/params/NullableBigDataFrameParam.scala
mn-mikke: > once the pipeline is fitted, the reference to the leaderboard frame can be released.

Yes, that's correct. If  the pipeline has been trained, the `PipelineModel` contains just `H2OMOJOModel`, which doesn't hold `leaderboard_frame`.


This class is needed in two cases:
- In Spark you can save and load unfitted pipeline. ([This test](https://github.com/h2oai/sparkling-water/pull/2505/files#diff-cbc1bc71745ce2fce921cacdcd297c07c410ff66f7a6dff978c22c6f10b13205R139) demonstrates the scenario)
- A parameter is also serialized during the communication between pyspark and scala backend

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AutoMLConfiguration.scala
mn-mikke: It's not exposed on the MOJO model and and pipeline.load() resets the value to `null`.

ml/src/main/scala/ai/h2o/sparkling/ml/params/NullableBigDataFrameParam.scala
mn-mikke: > A parameter is also serialized during the communication between pyspark and scala backend
This is not true, it's serialized only during the serialization of unfitted pipeline.

I suggest to remove this class in the bellow PR:
https://github.com/h2oai/sparkling-water/pull/2516

PR #2513, opened 2021-04-23
[SW-2554] Expose Blending Frame on H2OAutoML

mn-mikke:


.gitignore
```diff
@@ -87,6 +87,7 @@ jenkins/docker/regular-tests/Dockerfile
 ml/src-gen
 scoring/src-gen
 py/src-gen
+py-scoring/src-gen
 r/src-gen
```
sebhrusen, line 91: FYI, in `git-ignore` , the following pattern will ignore any subfolder named `src-gen`
```
src-gen/
```

sebhrusen: for confirmation (https://git-scm.com/docs/gitignore)
```
If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.
```

py/tests/unit/with_runtime_sparkling/test_automl.py
```diff
@@ -142,7 +143,31 @@ def testDeserializationOfUnfittedPipelineWithAutoML(classificationDataset):
...
+
+    automl = setParametersForTesting(H2OAutoML())
+    automl.fit(trainingDateset)
+    defaultLeaderboard = truncateModelId(automl.getLeaderboard())
+
+    automl = setParametersForTesting(H2OAutoML()).setBlendingDataFrame(blendingDataset)
+    automl.fit(trainingDateset)
+    leaderboardWithBlendingDataFrameSet = truncateModelId(automl.getLeaderboard())
+
+    unit_test_utils.assert_data_frames_have_different_values(defaultLeaderboard, leaderboardWithBlendingDataFrameSet)
```
sebhrusen, line 162: maybe it would be better to compare only the SEs, and verify that the other models are the same (otherwise you can't prove that the difference in SEs is due to the blending frame).

.gitignore
mn-mikke: `py-scoring/src-gen` worked for me locally, but will change according to your suggestion. Thank you!

py/tests/unit/with_runtime_sparkling/test_automl.py
mn-mikke: Will improve that, thank you!

.gitignore
sebhrusen: it's just that you were listing all the `src-gen` modules, and usually you want to excluded those by default.

sebhrusen (approved):
looks good!

michalkurka:
PR as such looks good, I don't have yet full understanding why NullableBigDataFrameParam is necessary but that is discussed in another PR.

michalkurka (approved):

PR #2515, opened 2021-04-26
[SW-2551] Mention the ``spark.ext.h2o.ssl.ca.cert`` Property in Secured Flow Documentation

mn-mikke:


neemah2o (changes requested):
Please correct small typos, but otherwise, looks good.

mn-mikke:
Thanks @neemah2o for your review! I've implemented your comments.

PR #2521, opened 2021-05-04
[SW-2558] Upgrade dbplyr in SW Testing Docker Image

mn-mikke:
Since dbplyr is fixed 1.4.2 and it's not compatible with Sparklyr 1.6.2, R tests fail.

```
13:06:39  Failed to connect to localhost port 54323: Connection refused
13:06:39  Backtrace:
13:06:39    1. dplyr::collect(mojo$transform(sdf)) testMojo.R:54:2
13:06:39   25. h2o:::.h2o.__remoteSend(...)
13:06:39   26. h2o:::.h2o.__checkConnectionHealth()
```

### DIFF
--- a/ci/build.gradle
+++ b/ci/build.gradle
@@ -36,7 +36,7 @@ task createDockerfile(type: Dockerfile, dependsOn: copyFiles) {
 
   runCommand """\\
                 R -e 'install.packages("testthat", repos = "http://cran.us.r-project.org")' && \\
-                R -e 'install.packages("dbplyr", repos = "http://cran.us.r-project.org")' && \\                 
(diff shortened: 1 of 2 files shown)


mn-mikke:
@michalkurka Merging without your review to trigger the release. IMHO, versions of R dependencies shouldn't be fixed in the docker image, but upgraded before each run to catch incompatibility problems ASAP. Let's discuss it and eventually improve it in future.

michalkurka:
yes, let's discuss that - usually you want to test:
- minimal supported version
- current mainstream version
- latest version

doing all that might not be always practical - but imho you should never skip the minimal supported version because the chances are higher that you would use a feature that is not available in an older version than that the feature you are using would get deprecated

PR #2525, opened 2021-05-06
[SW-2559] Expose Scoring History and Variable Importances on H2OMOJOModel

mn-mikke:
Jira ticket: https://h2oai.atlassian.net/browse/SW-2559

### DIFF
--- a/py-scoring/src/ai/h2o/sparkling/ml/params/H2OMOJOModelParams.py
+++ b/py-scoring/src/ai/h2o/sparkling/ml/params/H2OMOJOModelParams.py
@@ -49,8 +49,8 @@ class H2OMOJOModelParams(H2OMOJOModelBase):
     def getScoringHistory(self):
         return H2OTypeConverters.scalaToPythonDataFrame(self._java_obj.getScoringHistory())
 
-    def getVariableImportances(self):
-        return H2OTypeConverters.scalaToPythonDataFrame(self._java_obj.getVariableImportances())
+    def getFeatureImportances(self):
+        return H2OTypeConverters.scalaToPythonDataFrame(self._java_obj.getFeatureImportances())
 
 
 class HasOffsetCol:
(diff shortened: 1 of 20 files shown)


py-scoring/src/ai/h2o/sparkling/ml/params/H2OMOJOModelParams.py
```diff
@@ -46,6 +46,12 @@ def getTrainingParams(self):
     def getModelCategory(self):
         return self._java_obj.getModelCategory()
 
+    def getScoringHistory(self):
```
michalkurka, line 49: is it documented anywhere what should users expect on the output?

mn-mikke: Currently, It's not documented. I will improve that.

py-scoring/src/ai/h2o/sparkling/ml/params/H2OMOJOModelParams.py
```diff
@@ -46,6 +46,12 @@ def getTrainingParams(self):
...
+        return H2OTypeConverters.scalaToPythonDataFrame(self._java_obj.getScoringHistory())
+
+    def getVariableImportances(self):
+        return H2OTypeConverters.scalaToPythonDataFrame(self._java_obj.getVariableImportances())
+
```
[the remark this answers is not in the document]
mn-mikke: Great suggestion, regarding naming conventions. We try to follow Spark terminology if possible. variables -> features, target -> label. 

Regarding the type, I would keep using DataFrames. Spark algorithms get all features as one Vector columns and you can identify an individual feature only by its position. The same analogy is used for `featureImportances`

SW algorithms accept features in multiple columns and IMHO it's great benefit to associate features with their names. H2O-3 `variableImportances` also offer more values for a particular feature (`Relative Importance`, `Scaled Importance`, `Percentage`)

mn-mikke:
> When people are training CV models, how do they have access to scoring history of the CV models?

As discussed, it's not currently possible

> Also how do they access CV metrics, training metrics and validation metrics? 
```
scala> model.getTrainingMetrics()
res2: Map[String,Double] = Map(Logloss -> 0.5407563235669484, Gini -> 0.5738086864779322, RMSE -> 0.42418099556613403, MSE -> 0.17992951699947662, R2 -> 0.24477535552017637, AUC -> 0.7869043432389661)

scala> model.getValidationMetrics()
res3: Map[String,Double] = Map()

scala> model.getCrossValidationMetrics()
res4: Map[String,Double] = Map()
```

michalkurka (approved):

PR #2526, opened 2021-05-11
[SW-2560] Expose all H2OMOJOModels from AutoML Leaderboard

mn-mikke:


michalkurka (approved):
I think the PR looks good.

It made me realize that the AutoML models are not deleted from DKV after the training is finished - can this create a memory leak?

sebhrusen (approved):
This looks good!

sebhrusen:
> It made me realize that the AutoML models are not deleted from DKV after the training is finished - can this create a memory leak?

@michalkurka if SW is only using the Mojos after training, then I agree with you that it could, should even, delete the AutoML instance from DKV, which will remove the attached models alltogether.

mn-mikke:
Binary models are not deleted even after fitting of a standalone SW algorithm. Let's improve this in another PR https://h2oai.atlassian.net/browse/SW-2566.

PR #2533, opened 2021-06-03
[SW-2562] Expose AutoEncoder as SW Estimator

mn-mikke:


michalkurka (approved):
Thank you @mn-mikke - this is a great addition and great starting point for other unsupervised algos like PCA

PR #2536, opened 2021-06-18
[SW-2568] Tutorial for H2OAutoEncoder

mn-mikke:


doc/src/site/sphinx/ml/sw_auto_encoder.rst
```diff
@@ -0,0 +1,263 @@
+Auto Encoder in Sparkling Water
```
michalkurka, line 1: usually spelled Autoencoder

michalkurka (approved):
thank you, I think this is great to illustrate the usage

PR #2537, opened 2021-06-21
[SW-2030] Expose PCA as SW Feature Estimator

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/FeatureEstimatorConfigurations.scala
```diff
@@ -79,7 +86,8 @@ trait FeatureEstimatorConfigurations extends ConfigurationsBase {
   override def algorithmConfiguration: Seq[AlgorithmSubstitutionContext] = super.algorithmConfiguration ++ {
 
     val algorithms = Seq[(String, Class[_], String, Seq[String])](
-      ("H2OAutoEncoder", classOf[DeepLearningParameters], "H2OAutoEncoderBase", Seq.empty))
+      ("H2OAutoEncoder", classOf[DeepLearningParameters], "H2OAutoEncoderBase", Seq.empty),
+      ("H2OPCA", classOf[PCAParameters], "H2ODimReductionEstimator", Seq.empty))
```
michalkurka, line 90: Why is this not called H2OPCABase eg.? Just wonder what makes these 2 different?

mn-mikke: Here I assume that ```H2ODimReductionEstimator``` will be a base class also for  GLRM in future and the prediction logic for MOJO models will be shared in https://github.com/h2oai/sparkling-water/pull/2537/files#diff-a194bca6a411b0b4777dd860d539c5a6b1784ab2d5d13b893519a94e329fa5b0R28

`AutoEncoder` is very specific and we need to define extra parameters like `mseCol` and `originalCol` somewhere (`AutoEncoderBase`).

I don't have a strong opinion about naming conventions, I can change it to `H2OPCABase` if you have prefer that. The base class can be changed eventually once we add GLRM.

koniecsveta (approved):
LGTM

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/FeatureEstimatorConfigurations.scala
michalkurka: Ok, thanks for the explanation. Makes sense

michalkurka (approved):
LGTM - didn't find any issue - great work!

PR #2541, opened 2021-06-24
[SW-2571] Get AutoML Python Tests Aligned with PUBDEV-8175

mn-mikke:


py/tests/unit/with_runtime_sparkling/test_automl.py
```diff
@@ -115,7 +115,7 @@ def testH2OAutoMLClassifierBehavesDiffenrentlyThanH2OAutoMLRegressor(prostateDat
 
 
 def prepareLeaderboardForComparison(df):
-    return df.withColumn("model_id", substring(col("model_id"), 1, 20)).drop("")
+    return df.withColumn("model_id", regexp_extract(col("model_id"), "(.*)_AutoML_[\d_]+((?:_.*)?)$", 1)).drop("")
```
sebhrusen, line 118: picking the 1st group is enough for your current tests if you're training only 5 models (and as soon as we don't change the automl steps) and in this case you could simplify the regexp to `"(.*)_AutoML_"`.
However if you want sth more robust that would work also when the grids are triggered, then maybe you should use
```suggestion
    return df.withColumn("model_id", regexp_replace(col("model_id"), "(.*)_AutoML_[\d_]+((?:_.*)?)$", r"\1\2")).drop("")
```
(if `regexp_replace` allows back references)

sebhrusen (approved):

py/tests/unit/with_runtime_sparkling/test_automl.py
mn-mikke: ok, I will simplify the regexp, thanks!

Issue #2544, opened 2021-07-02
Need a better zip check


NivekNey:
https://github.com/h2oai/sparkling-water/blob/05831cbbca38ec1e5b54874d431d97f134850d0c/py-scoring/src/ai/h2o/sparkling/Initializer.py#L206-L207

Having substring `.zip` in the path doesn't guarantee it to be a zip archive.

NivekNey:
Maybe `zipfile.is_zipfile` https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile is a better mechanism.

mn-mikke:
Hi @NivekNey,
your suggestion will be addressed by https://github.com/h2oai/sparkling-water/pull/2650.

PR #2545, opened 2021-07-02
[SW-XXX] Add `zipfile.is_zipfile` to determine zip or not

NivekNey:
Attempt to mitigate #2544


h2o-ops:
Hi @NivekNey, your PR title "Add `zipfile.is_zipfile` to determine zip or not" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #2553, opened 2021-07-22
[SW-2572] Fix getMonotoneConstraints() on H2OGBM and H2OXGBoost MOJO model

mn-mikke:
The method was returning `None`.


michalkurka (approved):
awesome, thank you

koniecsveta (approved):

PR #2554, opened 2021-07-22
[SW-2573] Fix interactionConstraints on H2OXGBoostMOJOModel in Python

mn-mikke:
The inner arrays were represented as java objects of py4j.


michalkurka (approved):
one less fixme in the code, nice work @mn-mikke

koniecsveta (approved):

PR #2557, opened 2021-07-27
[SW-2582] Expose H2O-3 Mojo Model on H2OMOJOModel in Scala

mn-mikke:


koniecsveta (approved):

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -116,6 +116,11 @@ abstract class H2OMOJOModel
 
   def getFeatureImportances(): DataFrame = $(featureImportances)
 
+  def getHexMojoModel(): _root_.hex.genmodel.MojoModel = {
```
michalkurka, line 119: maybe avoid using "hex"? we could call it "unwrapMojoModel" instead - since H2OMojoModel is a wrapper of MOJO

michalkurka (approved):

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mn-mikke: Updated, thank you!

PR #2559, opened 2021-07-29
[SW-2583] Expose "cross_validation_metrics_summary" on H2OMOJOModel

mn-mikke:


koniecsveta (approved):

py/tests/unit/with_runtime_sparkling/test_mojo.py
```diff
@@ -195,3 +195,12 @@ def testMojoModelCouldBeSavedAndLoaded(gbmModel, prostateDataset):
     result = loadedModel.transform(prostateDataset).drop("detailed_prediction")
 
     unit_test_utils.assert_data_frames_are_identical(expected, result)
+
+
+def testGetCrossValidationSummaru():
+    mojo = H2OMOJOModel.createFromMojo(
```
koniecsveta, line 201: nitpick: there is a typo in Summaru

mn-mikke: Good catch! Thank you!

michalkurka (approved):
LGTM - but in further PRs I think we should clarify dichotomy of SW metric vs H2O metric

mn-mikke:
>  I think we should clarify dichotomy of SW metric vs H2O metric

@michalkurka I've changed the functionality to show only SW metric names.

PR #2560, opened 2021-07-30
[SW-2585] Migrate SW Automated Tests to CDH 6.3

mn-mikke:


ci/build.gradle
```diff
@@ -19,12 +19,8 @@ task copyFiles(type: Copy) {
 
 task createDockerfile(type: Dockerfile, dependsOn: copyFiles) {
   destFile = outputFile
-  from 'harbor.h2o.ai/opsh2oai/h2o-3-hadoop-hdp-2.2:84'
+  from testingBaseImage
 
-  environmentVariable("LANG", "'C.UTF-8'")
```
koniecsveta, line 24: with cdh-6.3 these removed commands are not needed anymore?

mn-mikke: - When I run the new base image, the `LANG` seems to be already set to `C.UTF-8`.:
```
docker run --privileged  -it --rm harbor.h2o.ai/opsh2oai/h2o-3-hadoop-cdh-6.3:84
root@8d663c9d305a:/# echo $LANG
C.UTF-8
root@8d663c9d305a:/# su jenkins 
jenkins@8d663c9d305a:/$ echo $LANG
C.UTF-8
```

- According to [the linux man page](https://man7.org/linux/man-pages/man1/locale.1.html), `locale` just displays current locale information.

- `yarn-site.xml` override was breaking hive tests in `/etc/startup/53_hive_load`. The default `yarn-site.xml` doesn't cause any problems + all sw tests are passing.

koniecsveta (approved):

PR #2562, opened 2021-08-05
[SW-2595] Add 'mean_per_class_error' to model trainings map

mn-mikke:
I've tested the feature manually:
image


scoring/src/main/scala/ai/h2o/sparkling/ml/internals/H2OMetric.java
```diff
@@ -43,7 +43,8 @@
   Withinss(false),
   Betweenss(true),
   TotWithinss(false),
-  Totss(false);
+  Totss(false),
+  MeanPerClassError(false);
```
mn-mikke, line 47: The boolean flag false in this case means: the higher value the metric has, the worse model we get.

michalkurka (approved):

koniecsveta (approved):

PR #2567, opened 2021-08-12
[SW-2591] Expose Cross Validation MOJO Models on H2OMOJOModel in Python

mn-mikke:


py/tests/unit/with_runtime_sparkling/test_mojo.py
```diff
@@ -195,3 +195,25 @@ def testMojoModelCouldBeSavedAndLoaded(gbmModel, prostateDataset):
     result = loadedModel.transform(prostateDataset).drop("detailed_prediction")
 
     unit_test_utils.assert_data_frames_are_identical(expected, result)
+
+
+def testCrossValidationModelsAreAvialableAfterSavingAndLoading(prostateDataset):
```
koniecsveta, line 200: typo in Avialable

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/MOJOModelFactoryTemplate.scala
```diff
@@ -74,31 +75,42 @@ object MOJOModelFactoryTemplate extends ((Seq[AlgorithmSubstitutionContext]) =>
     }
 
     val models = s"""
-       |class H2OMOJOModel(H2OMOJOModelParams, H2OMOJOModelBase, H2OMOJOModelFactory):
+       |class WithCVModels(H2OMOJOModelFactory):
```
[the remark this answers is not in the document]
mn-mikke: @michalkurka Yes, CV models are saved always when cross-validation is applied. Np, I can add a flag parameter saying whether the CV models will be appended to the main model or not. Could you share more details why it should be optional? impact on performance? The size of the final model?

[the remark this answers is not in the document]
michalkurka: You are typically primarily interested in metrics of the CV fold models, you might also want to preserve the holdout  predictions (in case you want to stack the models), the actual model might be useful in specific cases but I don't believe this is common enough to warrant an n-times increase of the size of the saved data.

It would make sense to save the metrics always but not the actual models.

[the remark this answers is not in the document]
mn-mikke: Thanks for  the clarification! I've add a parameter to both PRs as requested.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
```diff
@@ -95,18 +95,30 @@ trait AlgorithmConfigurations extends ConfigurationsBase {
...
+      ("H2OXGBoostParams", classOf[XGBParamsV3], classOf[XGBoostParameters], xgboostFields, noDeprecation, true),
+      ("H2OGBMParams", classOf[GBMV3.GBMParametersV3], classOf[GBMParameters], gbmFields, noDeprecation, true),
+      ("H2ODRFParams", classOf[DRFV3.DRFParametersV3], classOf[DRFParameters], drfFields, noDeprecation, true),
+      ("H2OGLMParams", classOf[GLMV3.GLMParametersV3], classOf[GLMParameters], glmFields, noDeprecation, true),
+      ("H2OGAMParams", classOf[GAMV3.GAMParametersV3], classOf[GAMParameters], gamFields, noDeprecation, true),
+      ("H2ODeepLearningParams", classOf[DLParamsV3], classOf[DeepLearningParameters], dlFields, dlDeprecations, true),
+      ("H2OKMeansParams", classOf[KMeansParamsV3], classOf[KMeansParameters], kmeansFields, noDeprecation, true),
+      ("H2OCoxPHParams", classOf[CoxPHParametersV3], classOf[CoxPHParameters], coxPHFields, noDeprecation, false),
+      (
+        "H2OIsolationForestParams",
```
koniecsveta, line 108: why is this on multiple lines when previous algo params are just one liners?

koniecsveta (approved):
LGTM

py/src/ai/h2o/sparkling/ml/algos/SupportsCrossValidation.py
```diff
@@ -0,0 +1,34 @@
...
+#
+
+from ai.h2o.sparkling.ml.params.H2OTypeConverters import H2OTypeConverters
+from pyspark.ml.param import *
+
+
+class SupportsCrossValidation(Params):
+    generateCrossValidationModels = Param(
+        Params._dummy(),
+        "generateCrossValidationModels",
```
michalkurka, line 25: whether CV models will be generated/created is controlled by nfold parameter, whether they will be saved/exported from H2O - is controlled by this parameters

this makes me thing we need to change the naming, it could be "keepCrossValidationModels" - with similar semantics as "keep_crossvalidation_models" in h2o

michalkurka: essentially it does the same thing - it preserves the models

mn-mikke: no problem, fill fix that. Thank you!

michalkurka (approved):
Thank you @mn-mikke

PR #2568, opened 2021-08-13
[SW-2597] Algorithms Supporting Cross-validation Must Remove Fold Column from the List of Features

mn-mikke:


api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
```diff
@@ -135,9 +135,13 @@ trait AlgorithmConfigurations extends ConfigurationsBase {
       ("H2OXGBoost", classOf[XGBoostParameters], "H2OTreeBasedSupervisedAlgorithm", Seq(withDistribution)),
       ("H2OGBM", classOf[GBMParameters], "H2OTreeBasedSupervisedAlgorithm", Seq(withDistribution)),
       ("H2ODRF", classOf[DRFParameters], "H2OTreeBasedSupervisedAlgorithm", Seq(withDistribution)),
-      ("H2OGLM", classOf[GLMParameters], "H2OSupervisedAlgorithm", Seq(withFamily)),
-      ("H2OGAM", classOf[GAMParameters], "H2OSupervisedAlgorithm", Seq(withFamily)),
-      ("H2ODeepLearning", classOf[DeepLearningParameters], "H2OSupervisedAlgorithm", Seq(withDistribution)),
+      ("H2OGLM", classOf[GLMParameters], "H2OSupervisedAlgorithmWithFoldColumn", Seq(withFamily)),
+      ("H2OGAM", classOf[GAMParameters], "H2OSupervisedAlgorithmWithFoldColumn", Seq(withFamily)),
+      (
+        "H2ODeepLearning",
```
koniecsveta, line 141: this also seems to be on multiple lines while previous items are one liners

koniecsveta (approved):
LGTM, thanks!

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
mn-mikke: The SW uses spotless plugging during the gradle build. There is a rule for the maximum line size.  This line is above the limit and when I run `./gradlew spotlessApply` to apply the rules, the code was re-formatted to multiple lines.

py/src/ai/h2o/sparkling/ml/algos/H2OSupervisedAlgorithmWithFoldColumn.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from ai.h2o.sparkling.ml.algos.H2OSupervisedAlgorithm import H2OSupervisedAlgorithm
+
+
+class H2OSupervisedAlgorithmWithFoldColumn(H2OSupervisedAlgorithm):
```
michalkurka, line 21: would it make more sense to make it a trait? the hierarchy is getting quite complicated

what if you need to add another characteristics/behavior - fold column, stop column... if you have 3+ traits you will not be able to get away with inheritance

michalkurka (approved):
Looks good functionality wise - I just had a concern about complex class hierarchy

py/src/ai/h2o/sparkling/ml/algos/H2OSupervisedAlgorithmWithFoldColumn.py
mn-mikke: Yes, it makes sense to use trait on Scala side at first look, but I believe that @satai did it this way because of the logic in [this method](https://github.com/h2oai/sparkling-water/blob/bf7f34571b86bfc45f48a197b87de1947b086c85/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OSupervisedAlgorithmWithFoldColumn.scala#L35). It expects other fields to be present on the algorithm.

michalkurka: thanks for explaining

PR #2575, opened 2021-08-23
[SW-2603] Remove Cross-validation-related Parameters from H2OAutoEncoder

mn-mikke:
From [H2O documentation](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/deep-learning.html?highlight=autoencoder):

>Note: Cross-validation is not supported when autoencoder is enabled.


michalkurka (approved):
Thank you @mn-mikke

PR #2582, opened 2021-09-06
[SW-2607] Change Error Caused by Setting MOJO Model Parameters to Warning

mn-mikke:
When `H2OMOJOModel` is loaded, it parsers training parameter values from mojo internals. When parsing of an individual parameter goes wrong, an error could be displayed in `sparkling-shell` and that's confusing for users.:
image

Since training parameter values are not the key functionality and  the model can be still used for scoring, I'm suggesting to change such errors to warnings.


koniecsveta (approved):

mn-mikke:
Thanks @koniecsveta for your review!

PR #2584, opened 2021-09-06
[SW-2608] Remove Support for Spark 2.1

mn-mikke:
Apache Spark 2.1 was deprecated in SW `3.32.0.2-1` (2020-11-19) by [SW-2485](https://0xdata.atlassian.net/browse/SW-2485).


koniecsveta (approved):

michalkurka:
@mn-mikke do we have a technical reason to deprecate?

I would suggest graceful deprecation (we follow this pattern in H2O-core):

Phases:

1. Warning (you already did that)
2. Turn the warning into an error (=it won't start), remove support from the documentation. Introduce a property that would allow the user to defeat the error (=start even though it is not officially supported). In this phase, everything is still technically supported and the tests are running. Do **not** document the switch that lets you defeat the error.
3. If phase 2 shows no one actually needed the support for Spark 2.1 - remove the support from the codebase completely

michalkurka:
@mn-mikke good plan - lets do it this way

PR #2585, opened 2021-09-07
[SW-2609] Update AutoML Tests to Consider 3 StackEnsamble Models in Leaderboard

mn-mikke:
The other PR fails because of the change in AutoML. The number of StackEnsemble models was increased from 2 to 3.

### DIFF
--- a/py/tests/unit/with_runtime_sparkling/test_automl.py
+++ b/py/tests/unit/with_runtime_sparkling/test_automl.py
@@ -159,12 +159,13 @@ def testBlendingDataFrameHasImpactOnAutoMLStackedEnsambleModels(classificationDa
     automl = setParametersForTesting(H2OAutoML())
     automl.fit(trainingDateset)
     defaultLeaderboard = separateEnsembleModels(prepareLeaderboardForComparison(automl.getLeaderboard()))
+    automl.getLeaderboard().show(truncate=False)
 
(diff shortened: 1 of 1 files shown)


sebhrusen (approved):
thanks @mn-mikke

sebhrusen:
@mn-mikke for posterity, adding my DM here explaining the changes causing this test failure on your side
```
We didn’t change the API for this release, but the “engine” has many modifications.
Basically, we split models and grids into groups, and when running automl with a limited time budget, we allocate the entire remaining time to the next group. Also, at the end of each group, we train 1 or 2 intermediary SEs (avoiding duplicate ones).
That’s why your test is failing, expecting 3 SEs here instead of previously 2.

for example, for a 1h run:
- before 3.30 (iirc): automl would allocate all budget to first model, if it converges, allocate all remaining budget to next model, … then all budget to next grid… then 2 SEs.
- before 3.34: before, automl would try to assign a bit of the time budget to every step, having a few minutes for each, and then build 2 SEs.
- from 3.34: automl assign all budget to 1st group (contains 3 models + 1 SE, each having their part of the budget). if models completed before budget, then assign all the remaining budget to group 2 (more models + SEs) and so on until we’re done with all the groups and if we still have time, there are additional groups with SEs trying different metalearners, and we resume the 2 most promising grids that converged before, this time with no convergence constraint + 2 finals SEs.
```

mn-mikke:
@sebhrusen Thanks for sharing your insight!

PR #2645, opened 2021-09-20
[SW-XXX] Shade google packages

dzlab:
Shade google packages like it is done for amazonaws


h2o-ops:
Hi @dzlab, your PR title "Shade google packages" is missing JIRA issue number! Please specify it in the following form [SW-]

PR #2646, opened 2021-09-21
[SW-2616] Enable Publishing of api-generation Project

mn-mikke:
The [`sparkling-water-ml` project references](https://github.com/h2oai/sparkling-water/blob/master/ml/build.gradle#L29) `sparkling-water-api-generation` which is not currently published.

### DIFF
--- a/api-generation/build.gradle
+++ b/api-generation/build.gradle
@@ -17,8 +17,6 @@ shadowJar {
   mergeServiceFiles()
 }
 
-artifacts {
-  api shadowJar
-}
-
 build.dependsOn shadowJar
+
+defineStandardPublication().call()
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,7 @@ if (project.hasProperty("doRelease")) {
 ext {
   // Published projects
   publishedProjects = [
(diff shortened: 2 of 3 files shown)


api-generation/build.gradle
```diff
@@ -17,8 +17,6 @@ shadowJar {
   mergeServiceFiles()
 }
 
-artifacts {
-  api shadowJar
-}
-
 build.dependsOn shadowJar
+
+defineStandardPublication().call()
```
michalkurka, line 22: interesting, never seen this one before outside of SW project

michalkurka (approved):

api-generation/build.gradle
mn-mikke: in SW, the publishing section is not defined globally. I'm not sure whether it's good or bad, but we can decide in each project whether a [regular jar](https://github.com/h2oai/sparkling-water/blob/master/gradle/publish.gradle#L68) or a [shadowed jar](https://github.com/h2oai/sparkling-water/blob/master/gradle/publish.gradle#L47) will be published.

PR #2647, opened 2021-09-24
[SW-2617] Add Official Support for Python 3.7, 3.8

mn-mikke:


ci/build.gradle
```diff
@@ -54,6 +54,32 @@ task createDockerfile(type: Dockerfile, dependsOn: copyFiles) {
...
+
+  runCommand """\\
+                cd /envs && \\
+                pip install -U virtualenv && \\
+                for pythonVersion in ${pythonEnvironments}; \\
+                do \\
+                  virtualenv --python python\$pythonVersion sw_env_python\$pythonVersion && \\
+                  . /envs/sw_env_python\$pythonVersion/bin/activate && \\
+                  pip install -U setuptools && \\
+                  pip install twine; \\
```
koniecsveta, line 79: this library doesn't need to be installed also with -U?

mn-mikke: Actually, the `-U` parameter is not needed for `setuptools` since we are installing packages to newly created virtual environments. Will update, thank you!

PR #2648, opened 2021-09-27
[SW-2618] Deprecate autoencoder Parameter on H2ODeepLearning

mn-mikke:


michalkurka (approved):
Thank you for removing the broken parameter

PR #2649, opened 2021-09-30
[SW-2619] Make Maximum Size of Requests and Responses on Flow UI Proxy Configurable

mn-mikke:
This PR makes the limits configurable and also changes the default maximum request size from 4kB to 32 kB and the default maximum request size from 16kB to 32 kB. This change fixes Flow UI accessibility  on Databricks Azure cluster when `spark.databricks.passthrough.enabled` is set to `true`.

### DIFF
--- a/core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
+++ b/core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
@@ -277,9 +277,9 @@ trait SharedBackendConf extends SharedBackendConfExtensions {
     setFlowExtraHttpHeaders(stringRepresentation)
   }
 
-  def setFlowProxyRequestMaxSize(value : Int): H2OConf = set(PROP_FLOW_PROXY_REQUEST_MAX_SIZE._1, value.toString)
+  def setFlowProxyRequestMaxSize(value: Int): H2OConf = set(PROP_FLOW_PROXY_REQUEST_MAX_SIZE._1, value.toString)
 
-  def setFlowProxyResponseMaxSize(value : Int): H2OConf = set(PROP_FLOW_PROXY_RESPONSE_MAX_SIZE._1, value.toString)
+  def setFlowProxyResponseMaxSize(value: Int): H2OConf = set(PROP_FLOW_PROXY_RESPONSE_MAX_SIZE._1, value.toString)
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/water/webserver/jetty9/SparklingWaterJettyHelper.scala
```diff
@@ -60,8 +60,10 @@ class SparklingWaterJettyHelper(hc: H2OContext, conf: H2OConf, h2oHttpView: H2OH
     handler.addServletMapping(m)
     val ipPort = conf.h2oCluster.get
     holder.setInitParameter("proxyTo", s"${conf.getScheme()}://$ipPort${conf.contextPath.getOrElse("")}")
-    holder.setInitParameter("idleTimeout", s"${conf.restApiTimeout}")
-    holder.setInitParameter("timeout", s"${conf.restApiTimeout}")
+    holder.setInitParameter("idleTimeout", conf.restApiTimeout.toString)
+    holder.setInitParameter("timeout", conf.restApiTimeout.toString)
+    holder.setInitParameter("requestBufferSize", conf.flowProxyRequestMaxSize().toString)
+    holder.setInitParameter("responseBufferSize", conf.flowProxyRequestMaxSize().toString)
```
michalkurka, line 66: looks like a typo? should be responseMaxSize?

michalkurka:
What about the head max size? In H2O we do this


    connector.setRequestHeaderSize(getSysPropInt(proto+".requestHeaderSize", 32*1024));
    connector.setRequestBufferSize(getSysPropInt(proto+".requestBufferSize", 32*1024));
    connector.setResponseHeaderSize(getSysPropInt(proto+".responseHeaderSize", connector.getResponseHeaderSize()));
    connector.setResponseBufferSize(getSysPropInt(proto+".responseBufferSize", connector.getResponseBufferSize()));

core/src/main/scala/water/webserver/jetty9/SparklingWaterJettyHelper.scala
mn-mikke: Good catch, thanks!

mn-mikke:
@michalkurka Changed the default values to 32kB, thanks!

koniecsveta (approved):

PR #2650, opened 2021-10-01
[SW-2620] Improve Zip Archive Check in Pysparkling Initializer

mn-mikke:
The pysparkling package is published as a zip archive on S3 which can be deployed to a spark cluster via `--py-files`. The pysparkling package can be also installed as pypi or conda package, which leads to extracted package on the file system.

The `Initializer` has to identify the case when source files are in the zip archive and when not. The original version did that via presence of `.zip` string in the absolute of a source file. This might lead to wrong decisions when a conda environment path contains a folder with `.zip` in its name.

This change was tested manually.:
- Run pysparkling shell that utilizes the zip archive and started the h2o context.
- Built an egg file, create new conda env, installed the egg file, start spark cluster and started h2o context


py-scoring/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -203,12 +201,13 @@ def isRunningViaDBCConnect():
 
     @staticmethod
     def getVersion():
-        here = path.abspath(path.dirname(__file__))
-        if '.zip' in here:
-            with zipfile.ZipFile(here[:-len("ai/h2o/sparkling/")], 'r') as archive:
-                version = archive.read('ai/h2o/sparkling/version.txt').decode('utf-8').strip()
+        packagePath = Initializer.__get_pysparkling_package_path()
```
michalkurka, line 204: just curious: 

the way it is done now, ai.h2o.sparkling is a python module

could you instead do 

```
import ai.h2o.sparkling
versionFile = path.join(ai.h2o.sparkling.__path__, "version.txt")
...
```

?

mn-mikke: `ai.h2o.sparkling.__path__[0]` returns full path of the package:
```
/tmp/spark-60b8d4b0-6fdd-4d94-a6aa-e157d8220f8f/userFiles-9162dd06-0f93-4b7e-9b49-f376b3f2dbfe/h2o_pysparkling_2.2-3.36.0.1-1-2.2.zip/ai/h2o/sparkling
```

This could help me to replace `path.join(packagePath, versionFile)` in [the else branch](https://github.com/h2oai/sparkling-water/pull/2650/files#diff-11c71d1b42b3764d81a184c81a61175e2db363726e94067a9de9f28c4f069baeR210) with just `versionFile`, but i need to work with a relative path in [the if branch](https://github.com/h2oai/sparkling-water/pull/2650/files#diff-11c71d1b42b3764d81a184c81a61175e2db363726e94067a9de9f28c4f069baeR210).

koniecsveta (approved):

PR #2659, opened 2021-10-15
[SW-2628] Java serialization of H2OMOJOModel

mn-mikke:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -39,6 +39,7 @@ import org.apache.spark.expose.Logging
 import org.apache.spark.ml.Model
 import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
 import org.apache.spark.sql.types._
+import ai.h2o.sparkling.utils.DataFrameSerializationWrappers._
```
kanech, line 42: Is the additional import needed here?

kanech (approved):
LGTM

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
mn-mikke: yes, [here](https://github.com/h2oai/sparkling-water/pull/2659/files/c4b15fc10c7d22afbbb0d465a0d6a8dc8d0da71d#diff-5b64d0b9c48d4b1b735251846a288e0c159484ca06de444229c98853b3ba2723L200) for example. `$(crossValidationMetricsSummary)` returns `DataFrameSerializationWrapper` and imported implicit method `toDataFrame` will look after the conversion to `DataFrame`.

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -34,6 +39,80 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
...
+    val expectedCVScoringHistory = model.getCrossValidationModelsScoringHistory()
+    val expectedCrossValidationSummary = model.getCrossValidationMetricsSummary()
+
+    val crossValidationSummary = deserializedModel.getCrossValidationMetricsSummary()
+    val deserializedCVModels = deserializedModel.getCrossValidationModels()
+    val deserializedCVScoringHistory = deserializedModel.getCrossValidationModelsScoringHistory()
+    TestUtils.assertDataFramesAreIdentical(expectedCrossValidationSummary, crossValidationSummary)
+    assertDataframesOnMOJOModels(model, deserializedModel)
+    var i = 0
+    while (i < expectedCVModels.length) {
```
krasinski, line 108: `for (i <- expectedCVModels.indices)`
or
`for (i <- 0 until expectedCVModels.length)`

would be more "scala-ish"

krasinski (approved):
LGTM 👍 

btw. I'm wondering if it could be possible to use some custom data structure instead of dataframes in the future for params

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
mn-mikke: Yeah, here it makes sense. The number of cv models is small. On a bigger collection I would prefer `while` loop because of the significant performance benefits.

mn-mikke:
>  I'm wondering if it could be possible to use some custom data structure instead of dataframes in the future for params.

DataFrames might be too big gun for such a small task. I went for data frames since binding for Python and R is for free and it's API is well defined and well-known.  I'm open to ideas for improvements, but we must be careful with breaking changes.

PR #2661, opened 2021-10-19
[SW-2632] Make H2OMOJOModel.load Independent on Scala Version

mn-mikke:
Scala 2.11 and 2.12 generates different `serializationVersionUID`. This could cause problems when moving H2OMOJOModel between different versions of Spark. This PR introduces  a new ObjectInputStream that ignores `serializationVersionUID`.


michalkurka:
Wouldn't it be better to serialize the DataFrames using a different format? Serializing using java serialization is security risk.

michalkurka:
Looks, good - just out of curiosity - is there no other way to convert a dataframe to json? something like https://stackoverflow.com/a/42274161/7301183

mn-mikke:
I didn't know about this DataFrame = Dataset<Row> conversion to Dataset<String> that you mentioned. There some sql functions that I forget about to do the conversion on column level ([from_json](https://spark.apache.org/docs/latest/api/sql/index.html#from_json), [to_json](https://spark.apache.org/docs/latest/api/sql/index.html#to_json))

I've removed the custom serialization of Spark Rows to JSON ASTs.

michalkurka:
thank you @mn-mikke

michalkurka (approved):

PR #2668, opened 2021-11-02
[SW-2640] InternalBackend Should Set IP Address Explicitly to H2O Node

mn-mikke:
It seems that Databricks HC cluster doesn't allow ICMP echo requests to the address of eth0 worker interfaces. H2O nodes uses ICMP echo requests to check if the address and its interface is in the healthy state. The explicit setting of IP address should bypass this check which is in the context of SW unnecessary.

### DIFF
--- a/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendUtils.scala
+++ b/core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendUtils.scala
@@ -41,14 +41,9 @@ private[backends] trait InternalBackendUtils extends SharedBackendUtils {
       })
   }
 
-  private def resolveWorkerIPAddress(): String = {
-    val hostname = getHostname(SparkEnv.get)
-    translateHostnameToIp(hostname)
-  }
-
-  private[sparkling] def setSelfAddressToH2ONode(conf: H2OConf): Unit = {
-    if (conf.nodeNetworkMask.isEmpty) {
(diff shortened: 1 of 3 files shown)


krasinski (approved):
LGTM 👍

core/src/main/scala/org/apache/spark/h2o/backends/internal/InternalBackendUtils.scala
```diff
@@ -39,6 +41,17 @@ private[backends] trait InternalBackendUtils extends SharedBackendUtils {
...
   }
 
+  private def resolveWorkerIPAddress(): String = {
+    val hostname = getHostname(SparkEnv.get)
+    translateHostnameToIp(hostname)
+  }
+
+  private[sparkling] def setSelfAddressToH2ONode(conf: H2OConf): Unit = {
+    if (conf.nodeNetworkMask.isEmpty) {
+      water.H2O.SELF_ADDRESS = InetAddress.getByName(resolveWorkerIPAddress())
```
michalkurka, line 51: at this point the ip parameter should be in `args` https://github.com/h2oai/sparkling-water/pull/2668/files#diff-fe97e9f0f98522f054e3201152e784827ef9332e678937fdd6b139aae956db50R151

would it be more consistent if we took it from there? if the ip definition changes in getH2OWorkerArgs it would never be applied here

mn-mikke: np, will update, thanks!

mn-mikke:
I will add a parameter for the original behavior.

PR #2669, opened 2021-11-02
[SW-2641] Upgrade MOJO runtime to 2.7.5

mn-mikke:


michalkurka (approved):

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModelTestSuite.scala
```diff
@@ -277,19 +277,19 @@ class H2OMOJOPipelineModelTestSuite extends FunSuite with SparkTestContext with
   }
 
   private def assertPredictedValues(preds: Array[Row]): Unit = {
-    assert(preds(0).getSeq[Double](0).head == 65.36320409515132)
-    assert(preds(1).getSeq[Double](0).head == 64.96902128114817)
-    assert(preds(2).getSeq[Double](0).head == 64.96721023747583)
-    assert(preds(3).getSeq[Double](0).head == 65.78772654671035)
-    assert(preds(4).getSeq[Double](0).head == 66.11327967814829)
+    preds(0).getSeq[Double](0).head shouldEqual 65.36320409515132 +- 0.001
```
pkozelka, line 280: just curious, is introducing the tolerance related (=required because of) mojo version change?

pkozelka (approved):

krasinski (approved):

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModelTestSuite.scala
mn-mikke: yes, you are correct. The new version of mojo runtime library gives a bit different results on old mojos. it seems I still need to update the tolerances:. 
Is difference ```64.98931238070139 did not equal 64.96902128114817``` expected?

mn-mikke: I get the same values till mojo runtime library `2.5.12`. `2.5.13` is not in maven and `2.5.14` gives a bit different results.

michalkurka:
@mn-mikke I think it looks okay, but the differences are significant


With this kind of test, we don't know if there is no issue on SW side. Instead of comparing to exact values, we could compare to reference results generated by that particular mojo runtime.

g-eoj:
https://h2oai.slack.com/archives/CB6C4LK7Y/p1637280398142400

mn-mikke:
Thanks @g-eoj, merging...

PR #2670, opened 2021-11-03
[SW-2642] Add Missing Scala Setters for 'spark.ext.h2o.extra.cluster.nodes' Property

mn-mikke:


core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
```diff
@@ -56,6 +56,10 @@ trait InternalBackendConf extends SharedBackendConf with InternalBackendConfExte
   /** Setters */
   def setNumH2OWorkers(numWorkers: Int): H2OConf = set(PROP_CLUSTER_SIZE._1, numWorkers.toString)
 
+  def setExtraClusterNodesEnabled(): H2OConf = set(PROP_EXTRA_CLUSTER_NODES._1, true)
```
michalkurka, line 59: is this something we can automatically test? == list all the options from InternalBackendConf and check that gthere is a getter/setter for each of them?

krasinski (approved):

core/src/main/scala/ai/h2o/sparkling/backend/internal/InternalBackendConf.scala
mn-mikke: This is good idea, but I don't know how to make association between property names and getters and setters in the current state. IMHO it would require to rename properties to a constant names.

michalkurka: Oh, I see - I didn't notice that. Maybe just a simple check that number of properties is the same as the number of setters/getters?

mn-mikke: I've added some tests based on the number of properties and number of getters/setters

michalkurka: thank you Marek, great to see that no other properties had missing setters/getters

michalkurka (approved):

PR #2671, opened 2021-11-03
[SW-2629] Rewrite Word2Vec to use H2OFeatureEstimator api

krasinski:
that one took me waaay to much time but made me understand a lot more


ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OWord2VecBase.scala
```diff
@@ -0,0 +1,50 @@
...
+  with H2OWord2VecExtraParams {
+
+  private val appendEmptyStringUdf: UserDefinedFunction = udf[Seq[String], Seq[String]](_ :+ "")
+
+  override def fit(dataset: Dataset[_]): H2OWord2VecMOJOModel = {
+    val inputCol: String = getInputCol()
+    val ds = dataset
+      .filter(col(inputCol).isNotNull)
+      .filter(size(col(inputCol)) =!= 0)
+      .withColumn(inputCol, appendEmptyStringUdf(col(inputCol)))
```
krasinski, line 24: the old impl did that and till this point I'm not 100% sure what's the reason behind that

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OWord2VecMOJOBase.scala
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.spark.sql.expressions.UserDefinedFunction
+import org.apache.spark.sql.types.StructType
+
+import scala.collection.mutable
+
+trait H2OWord2VecMOJOBase extends H2OFeatureMOJOModel with H2OWord2VecExtraParams {
+
+  override protected def inputColumnNames: Array[String] = Array(getInputCol())
+
+  protected override def mojoUDF: UserDefinedFunction = {
```
krasinski, line 33: that udf is a shameless copy-paste from H2OMOJOPredictionWordEmbedding, it will require some bigger refactoring to reuse that trait because it requires H2OAlgorithmMOJOModel, and that's H2OFeatureMOJOModel, I even tried refactoring but it turned out too complex for that PR

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OWord2VecTestSuite.scala
```diff
@@ -75,8 +79,14 @@ class H2OWord2VecTestSuite extends FunSuite with Matchers with SharedH2OTestCont
   test("Basic Word2Vec") {
```
krasinski, line 79: I did not change much in the test besides that one which didn't check anything anyway mostly to prove the new impl passes the same tests as the old one

ml/src/main/scala/ai/h2o/sparkling/ml/features/H2OWord2VecBase.scala
mn-mikke: it seems that empty string is used as a delimiter of sentences. explode will crate a string column from an array column. See https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-munging/tokenize.html#tokenize

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OWord2VecMOJOBase.scala
mn-mikke: `H2OMOJOPredictionWordEmbedding` needs to be removed sooner or latter. After this PR it will become a death code. Word2Vec should be the only algorithm with word embedding predictions.

scoring/src/main/scala/ai/h2o/sparkling/ml/params/H2OWord2VecExtraParams.scala
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package ai.h2o.sparkling.ml.params
+
+import ai.h2o.sparkling.ml.models.H2OFeatureEstimatorBase
+import org.apache.spark.sql.types.{ArrayType, FloatType, StructField, StructType}
+
+trait H2OWord2VecExtraParams extends H2OFeatureEstimatorBase with HasOutputCol with HasInputColOnMOJO {
+
+  override protected def validate(schema: StructType): Unit = {
```
mn-mikke, line 25: Don't we need to add additional type checks for inputCol? E.g array column?

scoring/src/main/scala/ai/h2o/sparkling/ml/params/HasInputColOnMOJO.scala
```diff
@@ -0,0 +1,14 @@
+package ai.h2o.sparkling.ml.params
+
+import org.apache.spark.ml.param.{Param, Params}
+
+trait HasInputColOnMOJO extends Params {
+  protected val inputCol: Param[String] = new Param[String](this, "inputCol", "Input column name")
+
+  def getInputCol(): String = $(inputCol)
+
+  setDefault(inputCol -> "")
```
mn-mikke, line 10: do we need to set default? or could we set it to a regular value like `inputCol`

mn-mikke (commented):
Don't we need to change code also in Python? Can you add tests to `test_mojo_parameters.py` and `MOJOParameterTestSuite`?

Otherwise, it looks good.

scoring/src/main/scala/ai/h2o/sparkling/ml/params/H2OWord2VecExtraParams.scala
krasinski: I added the "array of strings" check, wdyt?

ml/src/test/scala/ai/h2o/sparkling/ml/algos/TransformSchemaTestSuite.scala
```diff
@@ -42,12 +40,12 @@ trait TransformSchemaTestSuite extends FunSuite with Matchers {
     mojo
   }
 
-  test("transformSchema with detailed prediction col") {
+  test("transformSchema with prediction col") {
     val model = loadMojo(
       H2OMOJOSettings(withLeafNodeAssignments = getWithLeafNodeAssignments, withStageResults = getWithStageResults))
 
     val datasetFields = dataset.schema.fields
-    val expectedSchema = StructType(datasetFields ++ (expectedDetailedPredictionCol :: expectedPredictionCol :: Nil))
```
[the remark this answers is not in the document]
krasinski: it's because the only test which is using this test base is WordEmbeddingTestSuite.scala

scoring/src/main/scala/ai/h2o/sparkling/ml/params/H2OWord2VecExtraParams.scala
krasinski: nice catch!

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPreTrained.scala
```diff
@@ -0,0 +1,42 @@
...
+package ai.h2o.sparkling.ml.params
+
+import ai.h2o.sparkling.H2OFrame
+import org.apache.spark.sql.DataFrame
+
+trait HasPreTrained extends H2OAlgoParamsBase with HasDataFrameSerializer {
+  private val preTrained = new NullableDataFrameParam(
+    this,
+    "preTrained",
+    "Id of a data frame that contains a pre-trained (external) word2vec model.")
```
mn-mikke, line 27: Is the comment correct? :` Id of a data frame`

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OWord2VecTestSuite.scala
```diff
@@ -33,53 +37,75 @@ class H2OWord2VecTestSuite extends FunSuite with Matchers with SharedH2OTestCont
...
+    new Pipeline().setStages(Array(tokenizer, stopWordsRemover, w2v, kMeans))
   }
 
-  test("H2OWord2Vec Pipeline serialization and deserialization") {
-    getPipeline().write.overwrite().save("ml/build/w2v_pipeline")
-    val loadedPipeline = Pipeline.load("ml/build/w2v_pipeline")
-    val model = loadedPipeline.fit(trainingDataset)
-    val expected = model.transform(testingDataset)
-
-    model.write.overwrite().save("ml/build/w2v_pipeline_model")
```
mn-mikke, line 69: IMHO, it's good to check that persistence of the model doesn't affect the model behavior.

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPreTrained.scala
krasinski: that's copypasted from the [h2o-3 docs](https://docs.h2o.ai/h2o/latest-stable/h2o-docs/data-science/word2vec.html?highlight=pre_trained):
`pre_trained: Specify the ID of a data frame that contains a pre-trained (external) Word2vec model.`
that's a bit confusing but I guess every h2o frame has an ID and that could be passed here, not sure what to do about it

ml/src/test/scala/ai/h2o/sparkling/ml/algos/WordEmbeddingTestSuite.scala
```diff
@@ -32,25 +32,17 @@ class WordEmbeddingTestSuite extends FunSuite with Matchers with SharedH2OTestCo
 
   override protected lazy val dataset: DataFrame = {
     Seq(
-      "Lorem ipsum dolor sit amet, consectetuer adipiscing elit",
-      "Curabitur vitae diam non enim vestibulum interdum",
-      "Fusce suscipit libero eget elit",
-      "Nunc auctor").toDF("text")
+      Seq(
```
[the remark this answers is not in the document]
krasinski: deleted, moved to the word2vectestsuite

ml/src/test/scala/ai/h2o/sparkling/ml/algos/TransformSchemaTestSuite.scala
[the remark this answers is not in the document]
krasinski: deleted, moved to the word2vectestsuite

ml/src/test/scala/ai/h2o/sparkling/ml/features/H2OWord2VecTestSuite.scala
krasinski: not sure why I removed it, probably by mistake, it's back now, thanks

ml/src/main/scala/ai/h2o/sparkling/ml/params/HasPreTrained.scala
mn-mikke: Can we remove at least: `Id of `? Each H2OFrame has its own key(id), which serves as a reference the H2O cluster DKV. SW API uses Spark dataframes instead of H2O frames. After SW transforms dataframe to h2o frame, it passes [its key to h2o backend](https://github.com/h2oai/h2o-3/blob/879e58d3e5f98064b738189febecd845bf4f87c4/h2o-algos/src/main/java/hex/schemas/Word2VecV3.java#L53), but this is hidden to users.

This description will be utilized as source for [generated documentation](https://docs.h2o.ai/sparkling-water/3.0/latest-stable/doc/parameters/parameters_H2OWord2Vec.html).

mn-mikke (approved):
LGTM - Great work, thank you!

PR #2678, opened 2021-11-18
[SW-2634] Display Warnings Coming from ModelBuilders

krasinski:
I was trying to replicate the h2o python client behavior but I'm not sure if that's enough

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/utils/EstimatorCommonUtils.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/utils/EstimatorCommonUtils.scala
@@ -85,9 +85,10 @@ trait EstimatorCommonUtils extends RestCommunication {
   }
 
   private def printWarnings(messages: Array[ValidationMessageV3]): Unit = {
+    val warn = "WARN"
     messages
-      .filter(_.message_type == "WARN")
-      .map(warn => s"${warn.message} (field name: ${warn.field_name})")
+      .filter(_.message_type == warn)
+      .map(msg => s"$warn: ${msg.message} (field name: ${msg.field_name})")
       .foreach(System.err.println)
   }
 }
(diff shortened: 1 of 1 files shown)


ml/src/main/scala/ai/h2o/sparkling/ml/utils/EstimatorCommonUtils.scala
```diff
@@ -82,4 +83,11 @@ trait EstimatorCommonUtils extends RestCommunication {
     } while (H2OModel.modelExists(replacement))
     replacement
   }
+
+  private def printWarnings(messages: Array[ValidationMessageV3]): Unit = {
+    messages
+      .filter(_.message_type == "WARN")
+      .map(warn => s"${warn.message} (field name: ${warn.field_name})")
+      .foreach(logWarning(_))
```
mn-mikke, line 91: Shouldn't the warnings go to stdout to be visible in jupyter or databricks notebooks?

mn-mikke (commented):
What about pysparkling

ml/src/main/scala/ai/h2o/sparkling/ml/utils/EstimatorCommonUtils.scala
krasinski: done, I noticed h2o-3 python client uses "warnings.warn" which prints to stderr so I used that, please let me know if that's ok, I'm fine with changing that to stdout

I think pysparkling calls that code and the messages sent to stdout/stderr are printed in jupyter notebook

mn-mikke: If H2O python client uses `stderr`, your solution is correct!

> I think pysparkling calls that code and the messages sent to stdout/stderr are printed in jupyter notebook
yes, pysparkling calls the scala code via py4j, but it's better to check that it works since, python code and scala code run in different processes.

krasinski: I decided to add "WARN:" before the message

a little proof that it works:
Screenshot 2021-12-09 at 17 09 55

mn-mikke (approved):
LGTM - Great job, thank you!

PR #2682, opened 2021-11-29
[SW-2637] Add progress bar

krasinski:


core/src/main/scala/ai/h2o/sparkling/backend/utils/ProgressBar.scala
```diff
@@ -0,0 +1,24 @@
...
+object ProgressBar {
+  private final val BarLength = 50
+  private final val CarriageReturn = "\r"
+  private final val Empty = "_"
+  private final val Filled = "█"
+  private final val Side = "|"
+
+  private[sparkling] def printProgressBar(progress: Float, leaveTheProgressBarVisible: Boolean = false): Unit = {
+    val crOrNewline = if (leaveTheProgressBarVisible) System.lineSeparator else CarriageReturn
+    System.err.print(CarriageReturn + renderProgressBar(progress) + crOrNewline)
```
mn-mikke, line 12: Wouldn't it be better to print progressbar to stdout insteadoff stderr?

core/src/main/scala/ai/h2o/sparkling/backend/utils/ProgressBar.scala
```diff
@@ -0,0 +1,24 @@
+package ai.h2o.sparkling.backend.utils
```
mn-mikke, line 1: Missing license header

core/src/test/scala/ai/h2o/sparkling/backend/utils/ProgressBarTestSuite.scala
```diff
@@ -0,0 +1,34 @@
+package ai.h2o.sparkling.backend.utils
```
mn-mikke, line 1: missing license header

mn-mikke (commented):
The scala part looks good! What about the scenario when training algorithm in pysparkling API?

core/src/main/scala/ai/h2o/sparkling/backend/utils/ProgressBar.scala
krasinski: There are two reasons for why I did that:
1. I noticed that Spark does that with it's progress bar and I think it works well so I did the same thing
2. `java.lang.System.err` javadoc says:
> By convention, this output stream is used to display error messages or other information that should come to the immediate attention of a user even if the principal output stream, the value of the variable out, has been redirected to a file or other destination that is typically not continuously monitored.

even so I think changing that to stdout should be ok if you think it would be better

mn-mikke: My motivation to move the progress to stdout is based on the below points:
- H2O python and R client prints progressbar to stdout
- Spark prints out to stderr also other logging messages that were captured via `org.apache.spark.internal.Logging`. So the stderr can contain lots of messages coming from Spark depending the logging level and the information about the progress bar might get easily lost.
- Is stderr visible in Jupyter/Databricks notebook?

Please, correct me if any of the points is not true.

krasinski: now it's clear, thank you, I'm checking right now how it works in the jupyter notebook and will implement accordingly

krasinski: image
so on stderr it looked like that, changing to stdout as requested

krasinski: image
and that's the current version I'm pushing right now

mn-mikke: This looks great!!!

core/src/main/scala/ai/h2o/sparkling/backend/SharedBackendConf.scala
```diff
@@ -366,6 +371,13 @@ object SharedBackendConf {
       |value Runtime.getRuntime()
       |.availableProcessors()""".stripMargin)
 
+  val PROP_PROGRESS_BAR_ENABLED: BooleanOption = (
+    "spark.ext.h2o.progressbar.enabled",
+    true,
+    """setProgressBarEnabled()
+      |setProgressBarDisabled()""".stripMargin,
+    "Decides whether to display progress bar related to H2O jobs on stderr or not.")
```
mn-mikke, line 379: on stderr: Haven't you changed it to stdout?

mn-mikke (approved):
LGTM - Great job, thank you!

PR #2687, opened 2021-12-07
[SW-2657] Add example of structured streaming application

kanech:


examples/src/main/scala/ai/h2o/sparkling/examples/CraigslistJobTitlesStructuredStreamingApp.scala
```diff
@@ -0,0 +1,57 @@
+
```
mn-mikke, line 1: Missing license header

mn-mikke (approved):
Missing header, otherwise LGTM

PR #2688, opened 2021-12-09
[SW-2645] Treat Spark Boolean type as Categorical type within H20 frame.

kanech:


mn-mikke:
@kanech May I ask you to add a note about this change to [migration guide](https://github.com/h2oai/sparkling-water/blob/master/doc/src/site/sphinx/migration_guide.rst#from-334-to-336) ?

core/src/main/scala/ai/h2o/sparkling/backend/Writer.scala
```diff
@@ -134,7 +134,9 @@ private[backend] object Writer {
           con.putNA(idxField)
         } else {
           entry.dataType match {
-            case BooleanType => con.put(row.getBoolean(idxField))
+            case BooleanType =>
+              val valueIndex = domainBuilder.addStringToDomain(row.getBoolean(idxField).toString, idxField)
```
michalkurka, line 138: It seems more natural to encode the boolean value as "0"/"1" instead of "true"/"false".

That would also be beneficial for backwards compatibility because if previously someone had a conversion code asfactor in their script they would end up with a domain represented as "0"/"1".

mn-mikke:
@kanech I've been thinking about this change. WDYT about updating also the reverse method `asSparkFrame`, which converts h2o frame to spark dataframe. [Currently categorical columns are converted to StringType](https://github.com/h2oai/sparkling-water/blob/bf7f34571b86bfc45f48a197b87de1947b086c85/core/src/main/scala/ai/h2o/sparkling/backend/utils/ReflectionUtils.scala#L162). WDYT about checking if a domain of a categorical column consists of `"0"`, `"1"` labels and then converting the column to BooleanType?

core/src/test/scala/ai/h2o/sparkling/backend/converters/DataFrameConverterTestSuite.scala
```diff
@@ -262,14 +262,17 @@ class DataFrameConverterTestSuite extends FunSuite with SharedH2OTestContext {
...
     val h2oFrame = hc.asH2OFrame(df)
 
     assertH2OFrameInvariants(df, h2oFrame)
-    assert(h2oFrame.columns(0).isNumeric())
-    assertVectorIntValues(h2oFrame.collectInts(0), Seq(1, 0, 1, 0))
+    assert(h2oFrame.columns(0).isCategorical())
+    assert(h2oFrame.columns(0).domain != null)
+    assert(h2oFrame.columns(0).domain.length == 2)
+    assert(h2oFrame.columns(0).domain.contains("0"))
+    assert(h2oFrame.columns(0).domain.contains("1"))
```
krasinski, line 275: I think `h2oFrame.columns(0)` should be extracted to a variable
also some more expressive assertions like `... should have size 2` and `... should contain("0")` could be used

krasinski (approved):
looks good to me 👍

doc/src/site/sphinx/design/spark_h2o_mapping.rst
```diff
@@ -67,3 +67,7 @@ As type ``T``, we support the following types:
 +--------------------------------------------------+
 
 As is specified in the table, Sparkling Water provides support for transforming arbitrary scala class extending ``Product``, which are, for example, all case classes.
+
+.. rubric:: Footnotes
+.. [1] The H20 type is String if cardinality is greater than 10 000 0000 or ratio of unique values to all values is 95% at least.
```
mn-mikke, line 72: nit: `at least` -> `or higher`?

doc/src/site/sphinx/design/spark_h2o_mapping.rst
```diff
@@ -67,3 +67,7 @@ As type ``T``, we support the following types:
...
+.. [2] The possible values are "1" and "0" for true and false respectively.
```
mn-mikke, line 73: nit: Could we mention in the sentence that "1" and "0" belong to the categorical column of H2OFrame?

mn-mikke (approved):
Minor comments, otherwise LGTM. Good job, thank you!

PR #2700, opened 2021-12-21
[SW-2670] Fix Databricks Smoke Tests

mn-mikke:
The databricks tests fails with the following error:
```
+ python setup.py bdist_wheel

/home/jenkins/miniconda/lib/python3.9/site-packages/setuptools/dist.py:501: UserWarning: The version specified ('3.36.0.1_1') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.

warnings.warn(

usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]

or: setup.py --help [cmd1 cmd2 ...]

or: setup.py --help-commands

or: setup.py cmd --help

error: invalid command 'bdist_wheel'

script returned exit code 1
```

### DIFF
--- a/ci/Jenkinsfile-databricks
+++ b/ci/Jenkinsfile-databricks
@@ -248,7 +248,7 @@ node("docker") {
                     def dbcVersion = sparkSpecificProps["databricksVersion"]
                     def scalaVersion = sparkSpecificProps["scalaBaseVersion"]
                     sh """
-                        . /envs/sw_env_python3.6/bin/activate"
+                        . /envs/sw_env_python3.6/bin/activate
                         pip install databricks-cli
                         """
                     buildSparklingWater(props, sparkMajorVersion)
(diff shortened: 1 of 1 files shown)


ci/Jenkinsfile-databricks
```diff
@@ -32,6 +32,7 @@ static String getSparklingVersion(props) {
 
 def startCluster(dbcVersion, sparkMajorVersion) {
     def jsonOutput = sh(script: """
+    . /envs/sw_env_python3.6/bin/activate
```
kanech, line 35: Not sure, but would it make sense to factor out the "activation command" and use a string interpolation then? Having the same command path on so many places seems a little bit clumsy.

krasinski: I think a method could be introduced, for example `runShWithPythonEnv` or something like that (not sure about the naming), that would take a string command and then just prefix it with the activation and run with `sh`

mn-mikke: Thanks for the feedback, I will try @krasinski's suggestion.

krasinski (changes requested):
LGTM

krasinski (approved):

kanech (approved):

PR #2701, opened 2021-12-21
[SW-2663] Expose "start_time", "end_time", "run_time" and "default_threshold" on H2OMOJOModel

kanech:


ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -463,6 +463,70 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
...
+      this.getClass.getClassLoader.getResourceAsStream("multi_model_iris.mojo"),
+      "multi_model_iris.mojo")
+
+    mojo.getStartTime() shouldBe 1631392711317L
+  }
+
+  test("getStartType is exposed for freshly trained model") {
+    val (_, mojo) = multinomialModelFixture()
+
+    mojo.getStartTime() should not be 0L
```
krasinski, line 477: I would go with `mojo.getStartTime() should be > 0L`

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -463,6 +463,70 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
...
+      this.getClass.getClassLoader.getResourceAsStream("multi_model_iris.mojo"),
+      "multi_model_iris.mojo")
+
+    mojo.getEndTime() shouldBe 1631392711360L
+  }
+
+  test("getEndTime is exposed for trained model") {
+    val (_, mojo) = multinomialModelFixture()
+
+    mojo.getEndTime() should not be 0L
```
krasinski, line 491: the same as with startTime

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -463,6 +463,70 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
...
+      this.getClass.getClassLoader.getResourceAsStream("multi_model_iris.mojo"),
+      "multi_model_iris.mojo")
+
+    mojo.getRunTime() shouldBe 43L
+  }
+
+  test("getRunTime is exposed for trained model") {
+    val (_, mojo) = multinomialModelFixture()
+
+    mojo.getRunTime() should not be 0L
```
krasinski, line 505: as above

krasinski (approved):
LGTM 👍 only commented minor details

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
kanech: Slightly academic point of view, but this would increase chance of accidental test failure - e.g. in case the system clock is set before 1.1.1970 for whatever reason.

michalkurka: We can also just save the timestamp before we start made the mojo and compare to this timestamp.

-1 sometimes (not in this specific case) indicates null/NA in h2o, it is a good suggestion by @krasinski

mn-mikke (approved):
LGTM

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -88,6 +88,15 @@ abstract class H2OMOJOModel
...
   protected var crossValidationModels: Array[H2OMOJOModel] = null
 
+  protected final val startTime: LongParam =
+    new LongParam(this, "startTime", "Start time in milliseconds")
+  protected final val endTime: LongParam =
+    new LongParam(this, "endTime", "End time in milliseconds")
+  protected final val runTime: LongParam =
+    new LongParam(this, "runTime", "Runtime in milliseconds")
+  protected final val defaultThreshold: DoubleParam =
+    new DoubleParam(this, "defaultThreshold", "Default threshold used for predictions")
```
mn-mikke, line 98: Default threshold used for predictions of classification models?

kanech: Sounds good, I'm gonna update the documentation string.

PR #2704, opened 2021-12-23
[SW-2671] change GCS import test file

krasinski:
the old file is no longer available and the test fails
that file is coming from some official public dataset and is not ideal as that's not a CSV but some other format and the resulting dataframe doesn't make much sense, but I think that's not that bad given that test just checks if some file coming from GCS can be used


kanech (approved):
Looks OK to me.

PR #2705, opened 2021-12-28
[SW-2664] Expose "cv_scoring_history" on H2OMOJOModel

kanech:


scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -190,6 +197,10 @@ abstract class H2OMOJOModel
 
   def getScoringHistory(): DataFrame = $(scoringHistory)
 
+  def getCrossValidationScoringHistory(): Array[DataFrame] = $(cvScoringHistory)
+
+  def getReproducibilityInformationTable(): Array[DataFrame] = $(reproducibilityInformationTable)
```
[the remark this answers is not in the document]
michalkurka: @mn-mikke I think this perhaps should not be exposed in Sparkling Water

Is it really applicable to sparkling water? Even if the user starts the same cluster the data might be shuffled in a way that will make the algo irreproducible.

This is in a Scala context - meaning the user is very unlikely calling directly "import_file" and is rather transferring data frames instead.

My suggestion here is to file a separate ticket for this feature and remove it from this PR and we can discuss if it is useful to have.

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -527,6 +527,70 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
     mojo.getDefaultThreshold() shouldBe 0.5
   }
 
+  test("getCrossValidationScoringHistory returns histories when they are available") {
```
michalkurka, line 530: I think "cross-validation scoring history" is a misleading name, I wouldn't know how to intuitively interpret this.

The output of this method seems to be the scoring history of "cross-validation models" (each of k-fold models).

The main issue here is that it could be interpreted of the scoring history of cross-validation itself (eg. training all cv models in concert and calculating cv metrics after each scoring iteration) - we don't have this.

cc: @tomasfryda

tomasfryda: @michalkurka This was introduced with the learning curve PR so we can make learning curves even for the CV models (thus `cross_validation_scoring_history` stores the scoring history of individual CV models) which can be helpful when a user doesn't provide `validation_frame`.

I don't mind changing it. Do you think `cross_validation_models_scoring_history` or `cv_models_scoring_history` would be clearer?

michalkurka: @tomasfryda I think this is fine on h2o side because it is internal afaik to (auto-)explain

however, in the context of SW I find this name confusing - there are other ways to make it clearer - eg. make it a map of model key to scoring history

kanech: As concluded in (other) discussion, CrossValidationModelsScoringHistory will be utilized.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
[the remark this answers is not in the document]
kanech: Due to questionable usability I decided to remove "reproducibility information table" from this PR for now.

scoring/src/main/scala/ai/h2o/sparkling/ml/params/NullableDataFrameArrayParam.scala
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  def jsonEncodeDF(dataFrame: DataFrame): JValue = {
+    if (dataFrame == null) {
+      JNull
+    } else {
+      val serializerClassName = parent.getDataFrameSerializer()
+      val serializer = Class.forName(serializerClassName).newInstance().asInstanceOf[DataFrameSerializer]
+      val serializedValue = serializer.serialize(dataFrame)
+      JObject(JField("serializer", JString(serializerClassName)), JField("value", serializedValue))
```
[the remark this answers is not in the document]
kanech: Ok, let's instantiate the serializer just once (when needed).

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
```diff
@@ -81,6 +81,9 @@ abstract class H2OMOJOModel
     new NullableStringParam(this, "modelCategory", "H2O's model category")
   protected final val scoringHistory: NullableDataFrameParam =
     new NullableDataFrameParam(this, "scoringHistory", "Scoring history acquired during the model training.")
+  protected var crossValidationModelsScoringHistory =
+    new NullableDataFrameArrayParam(this, "crossValidationModelsScoringHistory", "Cross validation scoring history.")
```
mn-mikke, line 85: Would it possible to update description to have the same meaning as the parameter name?

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -528,6 +528,44 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
...
+      TestUtils.assertDataFramesAreIdentical(
+        mojo.getCrossValidationModelsScoringHistory()(i),
+        reloadedModel.getCrossValidationModelsScoringHistory()(i))
+    }
+  }
+
+  test("getCrossValidationModelsScoringHistory returns empty array when model doesn't contain it") {
+    val mojo = H2OMOJOModel.createFromMojo(
+      this.getClass.getClassLoader.getResourceAsStream("airlines_boolean.mojo"),
+      "airlines_boolean.mojo")
```
mn-mikke, line 564: Would it be possible  to use different mojo model. This one was removed from the repository/

py/tests/unit/with_runtime_sparkling/test_mojo.py
```diff
@@ -261,23 +261,41 @@ def compareMetricValues(metricsObject, metricsMap):
...
+    mojo = H2OMOJOModel.createFromMojo("file://" + os.path.abspath("../ml/src/test/resources/gbm_cv.mojo"))
+    history = mojo.getCrossValidationModelsScoringHistory()
+    assert len(history) == 3
+    for history_df in history:
+        assert len(history_df.columns) == 16
+        assert history_df.count() == 3
+
+
+def testGetCrossValidationModelsScoringHistoryWhenDataIsMissing():
+    mojo = H2OMOJOModel.createFromMojo("file://" + os.path.abspath("../ml/src/test/resources/airlines_boolean.mojo"))
```
mn-mikke, line 299: ditto: different mojo model

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
kanech: Switched to another model.

scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOModel.scala
kanech: Documentation string fixed.

scoring/src/main/scala/ai/h2o/sparkling/ml/params/NullableDataFrameArrayParam.scala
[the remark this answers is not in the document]
kanech: NullableDataFrameArrayParam json de/serialization was updated.

PR #2707, opened 2022-01-02
[SW-2673] Fix ChicagoCrimeApp example not working

krasinski:


krasinski:
...and I thought that would be a quick fix :) 

converting to a draft as `BooleanColumn for mojo predictions` in `H2OMOJOModelTestSuite` test fails

so if I understand correctly what happens is:
1. an old MOJO has "true" and "false" values in the domain for a boolean column
2. that fails when trying to score after my change in that PR, because there's no "1" and "0" in that domain
3. but without that change scoring current on "fresh" model fails

so if I'm not mistaken we might need to go back to that "1" and "0" vs "true" and "false" discussion, or possibly h2o-3 could handle that special case?
@mn-mikke @kanech @michalkurka

mn-mikke:
Thanks @krasinski for noticing the problem. IMHO, this change is an a good path. The PR https://github.com/h2oai/sparkling-water/pull/2688 changed mapping of `BooleanType` from numeric to categorical. 

> I'm not mistaken we might need to go back to that "1" and "0" vs "true" and "false" discussion because of compatibility with old mojos

I don't think we changed labels for BooleanType. If a user wanted to convert a Boolean column to categorical, he/she could leverage `setColumnsToCategorical` method on an algorithm. This approach converted boolean column to numeric h2o column with 0/1 values and then the column was subsequently casted to categorical with "0", "1". 

I guess the problem you identified is there for a long time.

kanech (commented):
As far as I remember, airlines_boolean.mojo was used in other tests, so I wouldn't remove it. Otherwise OK.

mn-mikke:
@kanech I have replaced `airlines_boolean.mojo` with `deep_learning_prostate.mojo` which is missing `default_threshold` field as well. Please let me know if it's ok. It should be the only place where airlines_boolean.mojo is used.

kanech (approved):
LGTM

mn-mikke (approved):
LGTM

mn-mikke:
Merging without finished tests. The changes will be tested on pre-release tests.

PR #2709, opened 2022-01-05
[SW-2665] Updated Scala/Python example of RuleFit usage.

kanech:


mn-mikke (approved):
LGTM

koniecsveta:
LGTM, there is just one thing, the new rulefit demo is here: https://github.com/h2oai/h2o-3/pull/5982/files, I am fixing py2.7 issue but otherwise it is almost ready. So maybe it might be better to use this example, also for the future improvements

koniecsveta (approved):

PR #2715, opened 2022-01-12
[SW-2635] Model summary after training - basic version without html

krasinski:
as I mentioned some time ago I'm giving up with HTML in notebooks, at least for now - uploading a basic stdout version instead, also reimplemented from a map to a Dataframe as discussed

the text is displayed in jupyter both when printing the mojo and after the training:
Screenshot 2022-01-12 at 01 56 57


kanech (approved):
LGTM

ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OGridSearch.scala
```diff
@@ -176,11 +176,15 @@ class H2OGridSearch(override val uid: String)
...
       }
     } else {
-      sortedGridModels.map { case (modelId, _) => H2OModel(modelId).tryDelete() }
+      sortedGridModels.foreach { case (modelId, _) => H2OModel(modelId).tryDelete() }
     }
     algo.deleteRegisteredH2OFrames()
     deleteRegisteredH2OFrames()
-    gridModels.head
+    val result: H2OMOJOModel = gridModels.head
+    if (H2OContext.get().forall(_.getConf.isModelPrintAfterTrainingEnabled)) {
```
mn-mikke, line 184: Proposal: The method [`getGridModels()`](https://github.com/h2oai/sparkling-water/pull/2715/files#diff-2813143693098236fae6163182f97f766f262fa5d827764e1ca73afb7565b8d5R163) accesses h2oContext. I think you can get the h2oContext in the fit method and use it on two places (here, pass h2oContext to `getGridModels()`

ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOModelTestSuite.scala
```diff
@@ -623,4 +623,219 @@ class H2OMOJOModelTestSuite extends FunSuite with SharedH2OTestContext with Matc
       cvModels should be(null)
     }
   }
+
+  test("should return model summary") {
```
mn-mikke, line 627: nit: Can we add subjects to test names? E.g.:`H2OGBMModel`/`H2ODeepLearnigMOJOModel` should return model summary?

mn-mikke (approved):
LGTM, great job!

PR #2719, opened 2022-01-20
[SW-2681] Add Comment to Documentation about Contributions Support only in Binomial and Regression Models

mn-mikke:
See [this piece of code](https://github.com/h2oai/sparkling-water/blob/2a8edb271b4f7478e7bd2adee7cfe59b6a4c77ae/scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OAlgorithmMOJOModel.scala#L85) in `H2OAlgorithmMOJOModel`.


dzlab:
@mn-mikke is there any plan to also support contributions calculation for multi-label classification?

mn-mikke:
> is there any plan to also support contributions calculation for multi-label classification (like on iris dataset)?

I don't think there is any plan for near future to add that support. cc @michalkurka

PR #2730, opened 2022-02-16
[SW-2682] do constant check & row count in one iteration

krasinski:
model training from 1% up to 20% faster depending on the dataset size and how long is the training on h2o-3 side


core/src/test/scala/ai/h2o/sparkling/backend/PartitionStatsGeneratorTestSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+
+    resultOnConstantColumn.areFeatureColumnsConstant.value shouldBe true
+    resultOnConstantColumns.areFeatureColumnsConstant.value shouldBe true
+    resultOnNotConstantColumn.areFeatureColumnsConstant.value shouldBe false
+    resultOnNotConstantColumns.areFeatureColumnsConstant.value shouldBe false
+    resultWhereOnlyOneColumnIsConstant.areFeatureColumnsConstant.value shouldBe false
+  }
+
+  test("should correctly count values") {
+    val inputWithOnePartition = dataset.toDF(datasetCols: _*).rdd.coalesce(numPartitions = 2)
```
kanech, line 55: Variable name (inputWithOnePartition) seems to be misleading here.

michalkurka:
@krasinski would remove the constant check on Sparkling Water side make the code easier?

For most types, constant values don't really matter, there will be represented in an efficient way on H2O side. We could drop them after the frame is moved and sending them over shouldn't be expensive either.

The only issue is String columns - which are not compressed in any way in H2O.

Question: what is your definition of constant - do you consider NAs? Eg. a column that has one regular value and the other is NA could still be valuable - do we keep those?

core/src/main/scala/ai/h2o/sparkling/backend/converters/SparkDataFrameConverter.scala
```diff
@@ -40,23 +40,37 @@ object SparkDataFrameConverter extends Logging {
...
+      hc: H2OContext,
+      dataFrame: DataFrame,
+      frameKeyName: Option[String] = None,
+      featureColsForConstCheck: Option[Seq[String]] = None): H2OFrame = {
     val df = dataFrame.toDF() // Because of PySparkling, we can receive Dataset[Primitive] in this method, ensure that
     // we are dealing with Dataset[Row]
     val flatDataFrame = flattenDataFrame(df)
     val schema = flatDataFrame.schema
-    val rdd = flatDataFrame.rdd // materialized the data frame
+    val rdd = flatDataFrame.rdd.cache()
```
mn-mikke, line 52: Shouldn't we cache the rdd rather to a disk in case of internal backend , where h2o and spark executors share memory space?

krasinski:
@michalkurka 
- removing the check would make it a lot easier, the whole PR would be just removal of that check (5 lines) and unit test related to it - that's actually what I would prefer to do but I thought it's necessary
- when it comes to NA's - in my solution it doesn't matter - null is also some kind of value in this case so if we had 100000 nulls and one "real" value it wouldn't be considered as a constant column

core/src/main/scala/ai/h2o/sparkling/backend/PartitionStatsGenerator.scala
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.spark.sql.Row
+
+/**
+  * Goes over RDD partitions counting records and checking if given set of columns has constant values
+  */
+private[backend] object PartitionStatsGenerator {
+
+  def getPartitionStats(
+      rdd: RDD[Row],
+      maybeFeatureColumnsForConstantCheck: Option[Seq[String]] = None): PartitionStats = {
```
mn-mikke, line 29: nit: I would maybe use more generic expression for the variable. feature columns are more related to ml, conversions are more generic.

core/src/main/scala/ai/h2o/sparkling/backend/converters/SparkDataFrameConverter.scala
krasinski: My thought was "at the stage before training we should have plenty of RAM in most cases, why not use that?" - what do you think?

mn-mikke: H2OFrame as such in DKV takes usually majority of the memory. There is a rule of thumb for the memory of H2O cluster. The memory size of H2O cluster should be at least 4-5x more than the size of the dataset represented as CSV. 

if you cache the source Spark DF then there will be less space for the final H2OFrame since you keep them in memory at the same time.

michalkurka: > "at the stage before training we should have plenty of RAM in most cases

data import and transfer into H2O has much higher memory requirements than the training itself typically

michalkurka: It is better to prefer stability over performance - I think caching on disk is better in this case

krasinski: I added caching to disk when h2o backend mode is internal and default cache spark when external

mn-mikke (approved):
LGTM, thank you!

PR #2733, opened 2022-02-18
[SW-2690] Remove Sparkling Water P4J Gateway

mn-mikke:
The gateway has been introduced  in #2173 as a PoC. The logic has never been utilized and the related integration test is failing   because of a migration to conda environments.


krasinski (approved):

kanech (approved):
LGTM

PR #2736, opened 2022-02-18
[SW-2633][FOLLOWUP] Fix reference to scala version in gradle.properties

mn-mikke:
#2662 changed property name from `scalaBaseVersion` to `scalaVersion`. This change caused a problem in building k8s images.


kanech (approved):
LGTM

PR #2744, opened 2022-03-11
[SW-2693] Check Json field for emptiness before accessing it 

kanech:
For a metrics object containing a confusion matrix field, it was found that receiving a JSON that contains null confusion matrix leads to an exception in testing setup. However, as a null value is still legitimate JSON value a small fix was needed to avoid the exception.


michalkurka (approved):
This looks good to me - any concerns @mn-mikke / @krasinski?

mn-mikke (approved):
LGTM

PR #2754, opened 2022-03-31
[SW-2697] Fix Binary Model Cleaning in H2OAutoML

mn-mikke:
After the change https://github.com/h2oai/h2o-3/pull/6113, The stacked ensemble models needs to be delete first. Otherwise, during a try to delete a stacked ensemble model, [this line](https://github.com/h2oai/h2o-3/pull/6113/files#diff-ad93828b9151a2ed49334d31e84b42bb51f128704c09cde92ced8a6da673db3cR64) will throw NPE since a base model has already been deleted.

### DIFF
--- a/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
+++ b/ml/src/main/scala/ai/h2o/sparkling/ml/algos/H2OAutoML.scala
@@ -201,10 +201,13 @@ class H2OAutoML(override val uid: String)
   }
 
   private def deleteBinaryModels(): Unit = {
-    getLeaderboard().select("model_id").collect().foreach { row =>
-      val modelId = row.getString(0)
-      val model = H2OModel(modelId)
-      model.tryDelete()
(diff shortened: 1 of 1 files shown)


krasinski (approved):

mn-mikke:
@sebhrusen we delete binary models via `DELETE /3/Models/{modelId}`. Model ids are obtained via `GET /99/Leaderboards/$automlId`. We also delete all frames that were used for model training. 

> If so, calling this with the automl instance id or the job key (depending how you run AutoML) and with param cascade=true 

Will try this option, thanks!

sebhrusen:
@mn-mikke please note that deleting AutoML object with `cascade=true` will delete ALL artifacts created by this instance (models with submodels and metrics, split frames, leaderboard…) but also ONLY those. Especially, the training frame being created before being passed to AutoML remains untouched. Same for validation or leaderboard frame if they are passed explicitely to the AutoML instance, but if those are created by the AutoML instance by necessity as a split from the training frame, then those are deleted together with the instance.

PR #2760, opened 2022-04-11
[SW-2700] Add spline_orders to Tests Covering Parameter Propagation to  H2OGAMMOJOModel

mn-mikke:
`spline_orders` is a new parameter introduced to H2O-3. The tests covering parameter propagation to  `H2OGAMMOJOModel` are currently failing, since the parameter on the algorithm is not set, H2OGAMMOJOModel contians a value was set by H2O-3 backend.

### DIFF
--- a/ml/src/test/scala/ai/h2o/sparkling/ml/models/MOJOParameterTestSuite.scala
+++ b/ml/src/test/scala/ai/h2o/sparkling/ml/models/MOJOParameterTestSuite.scala
@@ -102,6 +102,7 @@ class MOJOParameterTestSuite extends FunSuite with SharedH2OTestContext with Mat
       .setNumKnots(Array(5, 5))
       .setBs(Array(1, 1))
       .setScale(Array(.5, .5))
+      .setSplineOrders(Array(-1, -1))
     val mojo = algorithm.fit(dataset)
 
     compareParameterValues(algorithm, mojo, Set("getFeaturesCols"))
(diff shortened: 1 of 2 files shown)


ml/src/test/scala/ai/h2o/sparkling/ml/models/MOJOParameterTestSuite.scala
```diff
@@ -102,6 +102,7 @@ class MOJOParameterTestSuite extends FunSuite with SharedH2OTestContext with Mat
       .setNumKnots(Array(5, 5))
       .setBs(Array(1, 1))
       .setScale(Array(.5, .5))
+      .setSplineOrders(Array(-1, -1))
```
kanech, line 105: Just a question, are those values (-1, -1) meaningful ones?

mn-mikke: These values haven't been selected to improve performance algorithm, but are valid according to the tests on H2O-3 side:
https://github.com/h2oai/h2o-3/pull/6090/files#diff-4340a0b6fcce0ae9216a6d99f650d45af3f198370c43a422a9f7056707c104e1R233

kanech: Well, I can't judge if combination of Bs, NumKnots and SplineOrders is a meaningful or not, but regarding the interface extension, this is OK (straightforward).

kanech (approved):
LGTM

PR #2768, opened 2022-04-27
[SW-2708] Fix Py4J classloader jar inject

krasinski:


mn-mikke (approved):
LGTM, thanks!

py-scoring/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -173,10 +173,11 @@ def __add_url_to_classloader(gateway, url):
         logger = Initializer.__get_logger(jvm)
         while loader:
             try:
+                urlClassLoaderClass = jvm.py4j.reflection.ReflectionUtil.classForName("java.net.URLClassLoader")
```
michalkurka, line 176: Why do we have this in the while-loop?

the instance of urlClassLoaderClass will not change in the loop

michalkurka, line 176: is it possible this will actually load the class using a different classloader resulting in a different instance of the class object?

py-scoring/src/ai/h2o/sparkling/Initializer.py
```diff
@@ -173,10 +173,11 @@ def __add_url_to_classloader(gateway, url):
...
                 classClass = gateway.jvm.Class
                 classArray = gateway.new_array(classClass, 1)
                 classArray[0] = url.getClass()
-                method = loader.getClass().getDeclaredMethod("addURL", classArray)
+                method = urlClassLoaderClass.getDeclaredMethod("addURL", classArray)
                 method.setAccessible(True)
```
michalkurka, line 181: if addURL was public we could just use `loader.getMethod`, correct? is this why we need to look up the method on URLClassLoader directly?

PR #2769, opened 2022-04-29
[SW-2707] Deprecated method causes slow down because of excessive logging.

kanech:
Don't evaluate deprecated method within a cycle as it leads to excessive logging.


mn-mikke (commented):
The change looks good, but R tests fail for some reason. Can you investigate why?

mn-mikke (approved):
LGTM

krasinski (approved):

PR #2770, opened 2022-05-06
[SW-2706] Generated Configuration Classes Should Handle Overloaded Methods

kanech:
Both Python and R (configuration classes) generation was modified to handle overloaded methods. Py4j and sparklyr/java/spark-1.5.2/invoke.scala handle parameter types resolution but adjustments were needed to pass variable number of parameters to those methods. Unit tests covering various corner cases in Python and R were added. Furthermore, an option to test just specified file was added to sparkling-water-r:test task. 

E.g. `./gradlew :sparkling-water-r:test -Pfile="tests/testthat/testConfigurationParameters.R"`

### DIFF
--- a/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/ConfigurationTemplate.scala
+++ b/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/ConfigurationTemplate.scala
@@ -80,8 +80,8 @@ object ConfigurationTemplate extends ((Array[Method], Array[Method], Class[_]) =
          |        return self
          |""".stripMargin
     } else {
-      s"""    def ${m.getName}(self, value):
-         |        self._jconf.${m.getName}(value)
+      s"""    def ${m.getName}(self, *args):
+         |        self._jconf.${m.getName}(*args)
          |        return self
          |""".stripMargin
     }
--- a/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/r/ConfigurationTemplate.scala
+++ b/api-generation/src/main/scala/ai/h2o/sparkling/api/generation/r/ConfigurationTemplate.scala
@@ -77,10 +77,6 @@ object ConfigurationTemplate extends ((Array[Method], Array[Method], Class[_]) =
   }
 
   private def generateSetter(m: Method): String = {
-    if (m.getParameterCount == 0) {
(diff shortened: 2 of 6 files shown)


krasinski (approved):
awesome 👍

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/ConfigurationTemplate.scala
```diff
@@ -80,8 +80,8 @@ object ConfigurationTemplate extends ((Array[Method], Array[Method], Class[_]) =
          |        return self
          |""".stripMargin
     } else {
-      s"""    def ${m.getName}(self, value):
-         |        self._jconf.${m.getName}(value)
+      s"""    def ${m.getName}(self, *args):
```
mn-mikke, line 83: Shouldn't the changes affect only the overloaded setters? Wouldn't variable arguments break auto-completion features in interactive python environments?

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/r/ConfigurationTemplate.scala
```diff
@@ -77,10 +77,6 @@ object ConfigurationTemplate extends ((Array[Method], Array[Method], Class[_]) =
   }
 
   private def generateSetter(m: Method): String = {
-    if (m.getParameterCount == 0) {
-      s"""    ${m.getName} = function() { invoke(jconf, "${m.getName}"); .self }"""
-    } else {
-      s"""    ${m.getName} = function(value) { invoke(jconf, "${m.getName}", value); .self }"""
-    }
+    s"""    ${m.getName} = function(...) { invoke(jconf, "${m.getName}", ...); .self }"""
```
mn-mikke, line 80: ditto

mn-mikke (commented):
One question, otherwise looks good. Great job!

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/python/ConfigurationTemplate.scala
kanech: Implementation was changed to use variable number of arguments just for overloaded setters. In other cases, parameter names of (original) Java methods were used for setter parameters. This is a slight difference to previous implementation where a generic parameter name `value` was utilized.

PR #2778, opened 2022-05-18
[SW-2713] Fix Failing R Tests

mn-mikke:
It seems that PR https://github.com/h2oai/h2o-3/pull/5107 introduced type attributes on data of H2OFrame.  `expect_equal` method looks also at attributes during value comparison by default. That's the reason why we are getting failures since the expected constants don't have any attributes. This PR proposes to use`expect_equal`  with  the parameter `ignore_attr` set to `TRUE` in order to  compare just values.

### DIFF
--- a/r/src/tests/testthat/testConversions.R
+++ b/r/src/tests/testthat/testConversions.R
@@ -33,9 +33,9 @@ test_that("Test transformation from h2o frame to data frame", {
   hf <- hc$asH2OFrame(sdf)
   sdf2 <- hc$asSparkFrame(hf)
 
-  expect_equivalent(sdf_nrow(sdf2), nrow(hf))
-  expect_equivalent(sdf_ncol(sdf2), ncol(hf))
-  expect_equivalent(colnames(sdf2), colnames(hf))
+  expect_equal(sdf_nrow(sdf2), nrow(hf))
+  expect_equal(sdf_ncol(sdf2), ncol(hf))
+  expect_equal(colnames(sdf2), colnames(hf))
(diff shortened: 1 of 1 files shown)


tomasfryda:
@mn-mikke just a small clarification, that PR (https://github.com/h2oai/h2o-3/pull/5107) didn't introduce type attributes, it just made sure that the type attributes are propagated when subsetting a frame.

Before the change, there was this kind of issue:
```r
df <- H2OFrame(a=1:5, b=.., c=datetime)
df$c # => datetime
df[, c("a","c")]$c # => numerical
```

r/src/tests/testthat/testConversions.R
```diff
@@ -33,9 +33,9 @@ test_that("Test transformation from h2o frame to data frame", {
   hf <- hc$asH2OFrame(sdf)
   sdf2 <- hc$asSparkFrame(hf)
 
-  expect_equal(sdf_nrow(sdf2), nrow(hf))
-  expect_equal(sdf_ncol(sdf2), ncol(hf))
-  expect_equal(colnames(sdf2), colnames(hf))
+  expect_equivalent(sdf_nrow(sdf2), nrow(hf))
```
kanech, line 36: Isn't expect_equivalent() deprecated? See https://testthat.r-lib.org/reference/expect_equivalent.html

mn-mikke: Good catch!

kanech (approved):
LGTM

PR #2783, opened 2022-05-23
[SW-2715] Remove PySpark Integration with Hive Test

mn-mikke:
The test is failing on Spark 3.2. I propose to remove the test since any other SW test relies on integration on Spark integration with Hive.


kanech (approved):
Yeah, it's probably not worth keeping the test.

krasinski (approved):

PR #2802, opened 2022-07-14
[SW-2726] Migration Jenkins under OSS account

mn-mikke:


kanech (approved):
I don't have any terraform experience, but I don't see any questionable change so hopefully everything is OK.  :-)

PR #2804, opened 2022-07-26
[SW-2726][FOLLOWUP] Fix Builds Against H2O branch

mn-mikke:
The new CI infrastructure sets `CI` environment variable in the testing docker image which causes H2O build [loading artifacts from local nexus](https://github.com/h2oai/h2o-3/blob/master/build.gradle#L6-L22).


krasinski (approved):

kanech (approved):
LGTM

PR #2807, opened 2022-08-01
[SW-2734] Fix Time Conversion Tests in Python API

mn-mikke:
The textual layout of Python `H2OFrame` on H2O-3 master branch has  changed from:
```
time
-------------------
2019-04-04 00:00:00
2020-01-01 00:00:00
2020-02-02 00:00:00
2020-03-03 00:00:00

[4 rows x 1 column]
```
to:
```
time
2019-04-04 00:00:00
2020-01-01 00:00:00
2020-02-02 00:00:00
2020-03-03 00:00:00
[4 rows x 1 column]
```

### DIFF
--- a/py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
+++ b/py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
@@ -276,8 +276,8 @@ def testConvertTimeValueFromSparkToH2OAndBack(spark, hc, timeZone, sparkType):
     df = spark.createDataFrame(data, ['strings']).select(col('strings').cast(sparkType).alias('time'))
 
     hf = hc.asH2OFrame(df)
-    hfResultString = hf.__unicode__()
-    hfParsedItems = hfResultString.split('\n')[1:5]
+    hfResultString = hf.to_str()
+    hfParsedItems = hfResultString.split('\n')[1:-2]
     hfParsedItems.sort()
 
     assert hfParsedItems == expected
--- a/py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
+++ b/py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
@@ -277,7 +277,7 @@ def testConvertTimeValueFromSparkToH2OAndBack(spark, hc, timeZone, sparkType):
 
     hf = hc.asH2OFrame(df)
     hfResultString = hf.__unicode__()
-    hfParsedItems = hfResultString.split('\n')[2:6]
+    hfParsedItems = hfResultString.split('\n')[1:5]
     hfParsedItems.sort()
 
     assert hfParsedItems == expected


py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
```diff
@@ -277,7 +277,7 @@ def testConvertTimeValueFromSparkToH2OAndBack(spark, hc, timeZone, sparkType):
 
     hf = hc.asH2OFrame(df)
     hfResultString = hf.__unicode__()
-    hfParsedItems = hfResultString.split('\n')[2:6]
+    hfParsedItems = hfResultString.split('\n')[1:5]
```
kanech, line 280: I'm not sure the code is expressive enough - the specified upper bound seems weird to me. I would probably wrote something like:
`# skip column name`
`hfParsedItems = hfResultString.split('\n')[1:]`

kanech (approved):

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
mn-mikke: We need to skip also the footer `[4 rows x 1 column]`. 

> I'm not sure the code is expressive enough

Imho, we don't need to be worried about performance if dataset has just six rows.

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
```diff
@@ -277,7 +277,7 @@ def testConvertTimeValueFromSparkToH2OAndBack(spark, hc, timeZone, sparkType):
 
     hf = hc.asH2OFrame(df)
     hfResultString = hf.__unicode__()
```
sebhrusen, line 279: why do you use `__unicode__()` here? This is a remnant from Py2.7 and is obsolete. Is SW also supporting Py2.7?
FYI, `hf.to_str()` should now provide a correct unicode representation in all versions of Py

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
kanech: Well, I missed the footer but I still think the sequence
`# skip header and footer`
`hfParsedItems = hfResultString.split('\n')[1:-1]`
is easier to understand for an occasional reader than
`hfParsedItems = hfResultString.split('\n')[1:5]`

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
mn-mikke: > why do you use __unicode__() here?

I'm not the author of this test

> Is SW also supporting Py2.7?

Yes,  for Spark 2.2 - 3.0

kanech (approved):

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
mn-mikke: I've replaced `hf.__unicode__()` with `hf.to_str()`

py/tests/unit/with_runtime_sparkling/test_Spark_to_H2O_conversions.py
mn-mikke: Changed to `hfResultString.split('\n')[1:-2]`

PR #2825, opened 2022-09-20
[SW-2199] Remove code related to H2OClient from SW codebase

mn-mikke:
H2O client is a legacy and unstable way to communicate  with H2O cluster. H2O client was a virtual H2O node running on Spark driver and communicated with the cluster via the same protocol as H2O nodes between each other.


michalkurka (approved):
Thank you, Marek, do we need to mention this in the migration guide or elsewhere in the documentation?

PR #2832, opened 2022-10-19
[SW-2732] Fix startup scripts for Windows environment

kanech:
Fixed reference to Scala version (in gradle) and provided binaries.


mn-mikke (approved):
LGTM, thank you!

PR #2855, opened 2023-01-28
[SW-2694] Stabilise the PCA test

krasinski:


mn-mikke:
Thanks for looking into this problem. What was causing instability of the test? I see you removed the option from autoencoder [here](https://github.com/h2oai/sparkling-water/pull/2855/files#diff-b59c36f2f6952f2aebc856179ce70d799ccbfbfb43d9f09d64e72b634ea48e54R143)

mn-mikke (approved):
LGTM, thank you!

PR #2861, opened 2023-02-21
[SW-2773] Save CV mojos properly

### LINKED ISSUE #2860: Bug in writing CV mojos
I've been trying to train a MOJO using the cross validation functionality. When saving the overall mojo, it should also be writing each each individual CV model (for each fold) in a folder called `cross_validation_models`. However after parsing through the code, it appears as though it only writes the zeroth-fold mojo every time--so no matter how many folds there are, it will only write the zeroth-fold model to each folder.

The bug can be found here: https://github.com/h2oai/sparkling-water/blob/master/scoring/src/main/scala/ai/h2o/sparkling/ml/models/H2OMOJOWriter.scala#L46

I believe this line should be changed to:
`val writer = new H2OMOJOWriter(cvModels(i), cvModels(i).getMojo().getAbsolutePath)`

krasinski:
fixes #2860


mn-mikke (approved):
LGTM, thank you!

PR #2864, opened 2023-03-03
[SW-2776] mojo2 test improvements

krasinski:


ml/src/test/scala/ai/h2o/sparkling/ml/models/H2OMOJOPipelineModelTestSuite.scala
```diff
@@ -107,88 +89,88 @@ class H2OMOJOPipelineModelTestSuite extends FunSuite with SparkTestContext with
...
+
+  test("Testing dataset with column order mixed should give the same prediction") {
+    val rdd = sc.parallelize(Seq(Row("1", "0", "65", "1", "2", "1", "1.4", "0", "6")))
+    val record = spark.createDataFrame(rdd, prostateTestData.schema)
+    val recordWithDifferentColumnOrder = record.selectExpr(record.columns.reverse: _*).toDF()
+    record.columns shouldNot equal(recordWithDifferentColumnOrder.columns)
+    record.columns.sorted shouldEqual recordWithDifferentColumnOrder.columns.sorted
+
+    val prediction = prostateMojoPipeline.transform(record).select(PredictionCol).collect().head
+    val predictionWithDifferentColumnOrder =
```
mn-mikke, line 146: In the test like this, we need to start doing assertions on multiple rows and make sure the predicted values are not the same.

mn-mikke (approved):
LGTM

PR #5639, opened 2023-06-27
[SW-2796] Upgrade H2O to 3.42.0.1

krasinski:


gradle.properties
```diff
@@ -1,11 +1,11 @@
 # artifacts group
 group=ai.h2o
 # Major version of H2O release
-h2oMajorVersion=3.40.0
+h2oMajorVersion=3.42.0
 # Name of H2O major version
-h2oMajorName=zz_kurka
+h2oMajorName=rel-3.42
```
mn-mikke, line 6: it should be `3.42.0`

gradle.properties
```diff
@@ -29,11 +29,11 @@ pythonEnvironments=3.6 3.7 3.8 3.9
...
 spark=3.4
 # Sparkling Water Version
-version=3.40.1.1-1-SNAPSHOT
+version=3.42.0.1-1-SNAPSHOT
 # Spark version from which is Kubernetes Supported
 kubernetesSupportSinceSpark=2.4
 databricksTestSinceSpark=2.4
 spotlessModern=true
-testH2OBranch=master
+testH2OBranch=rel-3.42
```
mn-mikke, line 37: it should be` rel-3.42.0`

PR #5698, opened 2023-12-04
[GH-3018] Add UpliftDRF

krasinski:


py/tests/unit/with_runtime_sparkling/test_mojo_parameters.py
```diff
@@ -85,6 +85,11 @@ def testExtendedIsolationForestParameters(prostateDataset):
     model = algorithm.fit(prostateDataset)
     compareParameterValues(algorithm, model)
 
+def testExtendedIsolationForestParameters(prostateDataset):
```
maurever, line 88: This seems unrelated to UpliftDRF, so I suggest moving it to another task. :)

maurever (approved):
Just one suggestion otherwise, LGTM.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
```diff
@@ -97,6 +100,10 @@ class AlgorithmConfigurations extends MultipleAlgorithmsConfiguration {
     val gamFields = Seq(ignoredCols, betaConstraints, gamCols)
     val gbmFields = Seq(monotonicity, calibrationDataFrame, ignoredCols)
     val drfFields = Seq(calibrationDataFrame, ignoredCols)
+    val upliftDrfFields = Seq(
+      ExplicitField("treatment_column", "HasTreatmentCol", "treatment"),
+      ExplicitField("response_column", "HasLabelCol", "label"),
```
mn-mikke, line 105: Why do we need to specify label column explicitly? Isn't uplift DRF just another supervised algorithm?

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
```diff
@@ -97,6 +100,10 @@ class AlgorithmConfigurations extends MultipleAlgorithmsConfiguration {
     val gamFields = Seq(ignoredCols, betaConstraints, gamCols)
     val gbmFields = Seq(monotonicity, calibrationDataFrame, ignoredCols)
     val drfFields = Seq(calibrationDataFrame, ignoredCols)
+    val upliftDrfFields = Seq(
+      ExplicitField("treatment_column", "HasTreatmentCol", "treatment"),
```
mn-mikke, line 104: Why do we need an explicit field for treatment? Could we just add another rule to  `ParameterNameConverter`?

maurever: Yes, for the uplift algorithm, the new treatment column is crucial.  However I am not sure how this algorithm configuration works, so I am not sure if this is the correct way to add the treatment column.

api-generation/src/main/scala/ai/h2o/sparkling/api/generation/common/AlgorithmConfigurations.scala
maurever: The response column is used the same way as in the supervised algorithm.

valenad1 (approved):
LGTM, thank you!

PR #5700, opened 2023-12-04
[GH-5699] Scala 2.13 preparations - part 1

krasinski:


gradle/scala.gradle
```diff
@@ -27,7 +27,9 @@ configurations
         }
 
 dependencies {
-    scalaCompilerPlugin "org.scalamacros:paradise_${scalaVersion}:2.1.1"
+    if (scalaVersion < '2.13') {
```
mn-mikke, line 30: Don't we need it for scala 2.13 as well?

mn-mikke (approved):
One question, otherwise LGTM.

PR #5740, opened 2024-09-29
remove `setuptools` as runtime dependency 

### LINKED ISSUE #5738: setuptools as run-time dependency
setuptools is not declared as a run-time dependency here: https://github.com/h2oai/sparkling-water/blob/1a0a318dce01180a6b5fe208fa85f110341d67a1/py/src/pyproject.toml#L12

Same with the setup.py:

https://github.com/h2oai/sparkling-water/blob/1a0a318dce01180a6b5fe208fa85f110341d67a1/py/src/setup.py#L40

Currently, it is only listed as a build dependency in both. However, the following line is run when sparkling-water is imported in Python

https://github.com/h2oai/sparkling-water/blob/1a0a318dce01180a6b5fe208fa85f110341d67a1/py-scoring/src/ai/h2o/sparkling/Initializer.py#L195

Note: starting in Python 3.12, setuptools does not come with a venv anymore by default

hutch3232:
- Fixes #5738

Avoid requiring `setuptools` at runtime. Using `importlib` from the python standard library.

`importlib.resources.path` is actually deprecated:
https://docs.python.org/3.12/library/importlib.resources.html#importlib.resources.path
however, the replacement was new in python 3.9. Can switch to the new `files` function once support is dropped for python < 3.9.


hutch3232:
As of yesterday, python 3.8 is now end of life. If this project wants to be consistent with that, I could revise this PR to drop support for 3.8 and update this code to use the new function instead of the deprecated version. Would that be preferred?

wendycwong:
We currently support python 3.6 and above.

hutch3232:
Sounds good, I'll leave it as is unless you suggest otherwise

hutch3232:
Hi @mmalohlava,
Could we _please_ fix this? `setuptools` is going to remove their `pkg_resources` module in two months which means this package will start breaking at runtime. https://github.com/pypa/setuptools/blob/main/NEWS.rst#v8090

PR #5757, opened 2025-08-19
fix(build): quick fix of JMH leak that was propagated via H2O-3 dependencies.

mmalohlava:
In h2o-3 the leak is fixed in https://github.com/h2oai/h2o-3/pull/16654,
but this PR simply omits JMH libraries from putting into final assembly.


valenad1 (approved):
lgtm, thank you!

PR #5762, opened 2026-03-04
fix: pkg_resources no longer supported

wesbarnett:
**Description:**

[`setuptools` has removed `pkg_resources`](https://setuptools.pypa.io/en/stable/history.html). This raises errors when running model predictions. This change fixes that by using the modern `importlib` equivalents.

**Related Issues (optional):**

* Looks like this might be a duplicate of #5740


wesbarnett:
Any news here (or anything of the related PR)? I'm happy to make adjustments to this as well. Currently we have an internal patch that we apply for this. @mn-mikke @tomasfryda @valenad1 @maurever

